PR #82, opened 2013-09-25
Use additional auth field - Sip-Source-IP-Address

lemenkov:
In some cases knowing login and password isn't enough. Sometimes we have to check SIP IP address as well (to implement whitelisted/blacklisted subnets or IPs).


lemenkov:
Time to close this. It doesn't not have a lot of sense since the introduction of `$acc_extra` pseudo-variable [in acc module](https://opensips.org/html/docs/modules/3.1.x/acc.html#ACC-extra-id).

PR #127, opened 2013-11-14
Add listen interface (socket) checking to db based natping.

wdoekes:
Hi,

re: http://opensips.org/pipermail/users/2010-April/012320.html

I have multiple opensipsen sharing the same database. I want to limit the nat pings to only send pings from the machines where the entity has registered. Pings from other machines make no sense and only produce useless traffic.

See patch.

I'm aware it could use some tweaking, like:
- making this optional (this probably fails when people use "udp:0.0.0.0:5060")
- making this work for get_all_mem_ucontacts too

Feedback is welcome.

Cheers,
Walter


wdoekes:
I think we can close this now :P

PR #259, opened 2014-06-30
Add Call-ID mangling support. Dialog module not needed.

wdoekes:
Add the "spiral" module that mangles the call-id without requiring the dialog state to be kept.

The `topoh` module pull request does the same, but more. This is limited to only altering the call-id header.


bogdan-iancu:
Outdated, anyhow topo-hiding is now supported without dialog module also

PR #327, opened 2014-09-09
store registration problem

fabriziopicconi:
when a registration arrives:
 ERROR:usrloc:wb_timer: inserting contact into database failed
 ERROR:db_text:dbt_table_check_row: null value not allowed - field 2


liviuchircu:
@ovidiusas seems to have fixed the definition of the `domain` column in 764968f386f3c580, allowing `NULL` values.

PR #345, opened 2014-10-01
load_balancer module enhancements

shripchenko:
Apparently, in its current implementation load_balancer module could be used in the only one way:

load balance on static destinations group and static resources list.
first - because do_load_balance() ignores new group.
second - because do_load_balance() - clean up dialog profiles according to new resources string. not a previous one.

I tried to enhance this module to be able to use it in the following ways (this is just a sample to cover all possible cases):

route[LB1] {
      if( load_balancer("ID_1", "res1;res2") ) {
           t_on_failure("LB2_RETRY");
           route(RELAY);
       }
       route(LB2);
}
route[LB2] {
      if( load_balancer("ID_1", "res3") ) {
           t_on_failure("LB2_RETRY");
           route(RELAY);
       }
       route(LB3);
}
failure_route[LB2_RETRY] {
       route(LB2);
}
route[LB3] {
      if( load_balancer("ID_2", "res4") ) {
           t_on_failure("LB3_RETRY");
           route(RELAY);
       }
       route(FAILURE);
}
failure_route[LB3_RETRY] {
        route(LB3);
}

I also need to implement a scenario where I load-balance some amount of calls and reject excessive load, but some high-priority calls have to be handled at all cost. so I still want to load-balance them to select less busiest node, but ignore the fact that the amount of traffic on this node might exceed configured limits.

Also, if i register call as a 'load' using lb_count_call() function, I want to be able undo it.

### DIFF
--- a/modules/load_balancer/lb_data.c
+++ b/modules/load_balancer/lb_data.c
@@ -951,7 +951,7 @@ int lb_is_dst(struct lb_data *data, struct sip_msg *_m,
 
 
 int lb_count_call(struct lb_data *data, struct sip_msg *req,
-			struct ip_addr *ip, int port, int grp, struct lb_res_str_list *rl)
+			struct ip_addr *ip, int port, int grp, struct lb_res_str_list *rl, int mode)
 {
 	static struct lb_resource **call_res = NULL;
 	static unsigned int call_res_no = 0;
@@ -1014,9 +1014,16 @@ end_search:
 
 	/* add to the profiles */
 	for( i=0 ; i<rl->n ; i++) {
-		if (lb_dlg_binds.set_profile( req, &dst->profile_id,
-		call_res[i]->profile, 0)!=0)
-			LM_ERR("failed to add to profile\n");
+		if( !mode ) {
+			if (lb_dlg_binds.set_profile( req, &dst->profile_id,
+			call_res[i]->profile, 0)!=0)
+				LM_ERR("failed to add to profile\n");
+		}
+		else {
+			if (lb_dlg_binds.unset_profile( req, &dst->profile_id,
+			call_res[i]->profile)!=1)
+				LM_ERR("failed to remove from profile\n");
+		}
 	}
 
 	/* unlock the resources*/
--- a/modules/load_balancer/lb_data.c
+++ b/modules/load_balancer/lb_data.c
@@ -733,6 +733,78 @@ int do_load_balance(struct sip_msg *req, int grp, struct lb_res_str_list *rl,
 }
 
 
+int do_lb_reset(struct sip_msg *req, struct lb_data *data)
+{
+	struct usr_avp *id_avp;
+	struct usr_avp *prfs_avp, *del_prfs_avp;
+	int_str id_val;
+	int_str prfs_val;
+
+	struct lb_dst *it_d, *last_dst;
+	struct lb_resource *it_r;
+	struct dlg_profile_table *it_p;
+
+
+	/* get previous iteration destination, if any */
+	last_dst = NULL;
+	id_avp = search_first_avp(0, id_avp_name, &id_val, 0);
+	if( id_avp ) {
+		if( is_avp_str_val(id_avp) == 0 ) {
+			for( it_d=data->dsts ; it_d ; it_d=it_d->next ) {
+				if( it_d->id == id_val.n ) {
+					last_dst = it_d;
+					LM_DBG("sequential call of LB - use previous dst %d [%.*s]\n", last_dst->id, last_dst->profile_id.len, last_dst->profile_id.s);
+					break;
+				}
+			}
+		}
+		destroy_avp(id_avp);
+	}
+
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Hi @shripchenko - my apologizes for the delay - I read your suggestions and I do agree - it is a better approach on the matter. Thank you for the idea.
On the code, does the current push reflects those suggestions ?

shripchenko:
@bogdan-iancu, yes it does.
You could make a review and merge.

shripchenko:
@bogdan-iancu, pinging you again ^)

bogdan-iancu:
Hi @shripchenko, I uploaded your patch - thank you very much. I did some changes and I will be thankful to you if you could review and retest.

Also, please open a new request with the updates on the documentation files :)

Regards,
Bogdan

shripchenko:
@bogdan-iancu, I did a small review.

I have a doubts about this one block of code in `lb_route()` function:
```c++
        /* save state - res */
        /* iterate AVPs once and delete old resources */
        destroy_avps(0, res_avp_name, 1 /*all*/);
        /* add new resources */
        for( i=0 ; i<res_cur_n ; i++ ) {
                res_val.s = res_cur[i]->name;
                if( add_avp(AVP_VAL_STR, res_avp_name, res_val) != 0 )
                        LM_ERR("failed to add RES AVP\n");
        }
```
Previously we searched through AVPs and delete old/add new.
This way we could be sure that no unnecessary memory allocations are made.
Our usual scenario is a continuous call of `lb_next()` with cached/unchanged resources list.

So we usually expect no memory allocations when handling resources list AVPs during subsequent `lb_next()` calls.

This is not true in your version of code.

What do you think?

bogdan-iancu:
Yes, I understood your logic in the original code, but you spare couple of mem allocs (which btw, are pretty fast) with the downside of searching and matching the resources - imho the effort of all those memcmp is comparable with the effort of 3-4 extra allocs (usually the list of used resources is relatively small). And I opted for this approach because code is simpler and more clear.

shripchenko:
Ok, I got it. I just try to avoid alloc's at any cost.

There are also some problems i found: please look at my commit: https://github.com/shripchenko/opensips/commit/26de7268598c54c1581befe21204d592b5ce2f09

And one more commit to review:https://github.com/shripchenko/opensips/commit/41812d4512a5ab2d331fc390e93e700a2fb88b8d

In my work I face a situation when there was a big amount of calls with duration=0 (like rejected calls).
And since LB normally takes first matched destination with minimum load first/one node was overwhelmed with calls and other nodes are almost free.
(load was even for all nodes and equal to zero)
it is, of course, a corner situation and not widely reproducible, but still possible.
And I believe it adds a new level of 'fairness' to LB.
It is cheap also.

bogdan-iancu:
@shripchenko, I uploaded your first patch, the one with the fixes on mallocs - thank you.

On the second patch - the load are evaluated (and destination selected) under the lock of the resources. So if you have the first call, it will get the first destination; next call will take the second (as the first one already has +1 load); third call will go to the third destination and so on. Unless the first one is already freed (0 load) by short call. So the load distribution is still good, but the first will keep getting the 0 duration calls. I guess this is your case, right ?

shripchenko:
@bogdan-iancu, Yes. exactly!

bogdan-iancu:
What about doing Round Robin ? the implementation will be simpler (you do not need that array), just remember the last selected destination. Also, in this case we do not need even the flag - it can be the default behavior - when we do evaluate the destinations (as load) we start all the time from the last selected.....just 2 cents without too much thinking :)

shripchenko:
global level round robin?
it would need a write lock at LB *data level to store last used index. (or maybe atomic inc).

per-call round robin? this means that every new call will start from 1st destination and it is actually how things implemented now.

Or am I miss something?

bogdan-iancu:
Global level, of course :) - but this may require locking indeed. The atomic funcs will not help here as we do not have an atomic get & inc.

shripchenko:
i believe that one(or a few) malloc's is better than continuous locking.

in theory we also could soften (add a rounding) by a some (configurable) _factor_ to `it_l <=> load` equation to slightly move from load-based distribution to (pseudo-)equal-rate distribution (with`rand()`)

So what do you think about this commit?


P.S.
we also could use a trick to do something like this:
```cpp
static struct lb_dst **dsts;

struct lb_dst **dsts_cur;
struct lb_dst *dsts_tmp;

if ( (flags & LB_FLAGS_RANDOM) && (data->dst_no > 1) ) {
/* allocate space with pkg_realloc() in dsts */
dsts_cur = dsts;
} else {
dsts_cur = &dsts_tmp;
dsts_size_cur = 1;
}
```

so no malloc's happen when LB_FLAGS_RANDOM flag unset.

bogdan-iancu:
@shripchenko , it is ok with me, please make pull request for master. 
And please do not forget about updating the docs too ;)
Thanks and regards,
Bogdan

PR #395, opened 2014-12-16
Newbranch

davesidwell:
The RTP Proxy module also requires correction in moving the session parameters from Branch AVP to Dialog variables. The "setid" field is an integer and therefore requires special handling.


razvancrainea:
This was pushed on the master branch.

PR #408, opened 2015-02-04
httpd: Make sure httpd can recognize Content-Type headers with params

staskobzar:
Content-Type headers like "text/xml; charset=utf-8" was not recognized. Some libs (like built-in ruby xmlrpc client) force these additional params.


ovidiusas:
Please re-work the patch to keep the code TAB indented.

staskobzar:
It is done. Please close this pull request and check #411 
https://github.com/OpenSIPS/opensips/pull/411
Thank you

PR #414, opened 2015-02-18
modified source and target table for registrant table migration

jbonicioli:


bogdan-iancu:
Thank you very much @jbonicioli !!

Issue #415, opened 2015-02-18
RTP Engine transport-protocol bug


eldonogithub:
The following is a patch for rtpengine.c, the code originally was attempting to get the index of the transport array from the ng_flags.transport bitfield. It was using 0x5 which happens to be the length of the transports array but not the correct value for a mask of the index. Instead, the lowest 3 bits are the index, not just the 1st and 3rd bit ( which is all 0x5 gives ). So a mask of 0x7 is required here.

We found this when attempting to change the protocol to RTP/SAVPF (0x03). 

diff --git a/modules/rtpengine/rtpengine.c b/modules/rtpengine/rtpengine.c
index 6c7b3e6..056c835 100644
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -1272,9 +1272,10 @@
         bencode_dictionary_add(ng_flags.dict, "flags", ng_flags.flags);
     if (ng_flags.replace && ng_flags.replace->child)
         bencode_dictionary_add(ng_flags.dict, "replace", ng_flags.replace);
-    if ((ng_flags.transport & 0x100))
+    if ((ng_flags.transport & 0x100)) {
         bencode_dictionary_add_string(ng_flags.dict, "transport-protocol",
-                transports[ng_flags.transport & 0x005]);
+                transports[ng_flags.transport & 0x007]);
+    }
     if (ng_flags.rtcp_mux && ng_flags.rtcp_mux->child)
         bencode_dictionary_add(ng_flags.dict, "rtcp-mux", ng_flags.rtcp_mux);

etamme:
Oops - looks like an oversight on my part.  Thanks for catching it, created a merge request with the fix.

bogdan-iancu:
Fixed, by accepting the pull request #416 
Thank you @eldonogithub  and @etamme

PR #416, opened 2015-02-18
fixed bitmask to cover all 3 bits vs. 1st and 3rd.  Thanks to eldonogith...

etamme:
fixes issue #415


bogdan-iancu:
Thank you @etamme  and @eldonogithub

PR #423, opened 2015-03-05
add event support to presence module.

razvancrainea:
The event E_PRESENCE_PUBLISH will be triggered each time the presence module receives a PUBLISH

The parameters of the event are:
user::AOR
domain::domain
event::event_type
expires::int_time_to_expire
etag::etag_string
body::string_with_xml_body_of_publish


razvancrainea:
Can you also write documentation for this new event triggered? After that I will merge it in the master branch.

Thanks,
Răzvan

PR #429, opened 2015-03-10
expose the DTLS passive flag to allow for passive relay through rtpengine

etamme:
not much else to say really.


bogdan-iancu:
Thank you

PR #431, opened 2015-03-17
dialog: Update dlg lifetime whe re-evaluating the AVP after the callback...

dsandras:
This completes the previous commit. Without this fix, when the SST module is activated and the routines are activated with the appropriate setflag, the calls will be interrupted after the sst timeout even when there is no sst support at all from both peers.

### DIFF
--- a/modules/dialog/dlg_handlers.c
+++ b/modules/dialog/dlg_handlers.c
@@ -1231,6 +1231,7 @@ after_unlock5:
 		/* update timer during sequential request? */
 		timeout = get_dlg_timeout_update(req);
 		if (timeout != 0) {
+			dlg->lifetime = timeout;
 			if (update_dlg_timer( &dlg->tl, dlg->lifetime )==-1)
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @dsandras , I also updated 1.10 - it suffers from the same problem.

dsandras:
No problem... What about 1.11 and master?
(I did not check them)


Le lundi 20 avril 2015 à 02:23 -0700, Bogdan Andrei IANCU a écrit :

> Thank you @dsandras , I also updated 1.10 - it suffers from the same
> problem.
> 
> —
> Reply to this email directly or view it on GitHub.
> 



-- 
Damien Sandras <dsandras@seconix.com>

bogdan-iancu:
Starting with 1.11, there is a different approach on updating the lifetime - the modules do access directly the dlg->lifetime to it (the AVP lifetime was dropped):
  - modules (for internal stuff) use the dlg->lifetime
  - script uses the dedicated script variable for lifetime.

PR #437, opened 2015-03-19
RTPEngine.c cleanup

rgagnon24:
- Fix prototypes to match the documentation (remove extra params from function definitions)
- Change function prototypes to use gparam_p instead of char *


liviuchircu:
A solid coding style is always nice. Thank you, @rgagnon24 !

PR #440, opened 2015-03-23
Documentation updates

rgagnon24:
Modernize the CSS for documentation as well as match the style of main opensips.org website

Update rtpproxy documentation to replace has_sdp() examples with has_body("application/sdp")

Update rtpengine documentation to elaborate on where to use rtpengine_delete()


liviuchircu:
First of all, thank you for the initiative! I completely agree that the module doc .css needs some fine tuning!

However, drastical changes usually have these kind of results: either a big *WoW!* or a *is it really better?!*. In my opinion, some of your ideas are good, others are not. While keeping in mind that I have little knowledge of CSS, here's a summary of _my_ first impression of the new style:

*Bad*
* imo, box/text width should be fixed regardless of display resolution
* Module header looks bad. No delimiting lines or stylish images and not that great of a style choice for the title
* gray text / white bg is harder to read than black text / white bg
* lots of wasted space in code block border areas (more than before)

*Good*
* emails are now displayed for each author/editor
* slightly bigger font size for all text
* rounded corners for code blocks
* nice choice of text and on-mouseover colors (apart from the gray thingy mentioned above)

We definitely need to hear what everyone has to say about this before making the big move. All in all, I'd say this is a damn good first try!

rgagnon24:
Reasons for the changes that you've marked as "bad":

Modern HTML should adjust to a user's monitor.  Some people are using tablets, some a phone, others still have tiny monitors, while I for example have one that is 22".  Hence the restriction on the width of the page was changed from a fixed 1200px to a percentage of the screen width.

The color choices were selected to match the main opensips.org website.  Those can be changed back if you guys want the page to look different from that.

The module headers are using the same gradient colors as that of the home page/menu area from opensips.org.

As for the code-block border sizes.  Probably the left and right could be narrowed out.  I just used 10, 15, or 20px as a simple number to get the look started.

Examples:

**10px  x 15px borders:**
example1015

**15px x 20px borders (as proposed)**
example1520

rgagnon24:
For this I probably could setup a temporary public webserver so it would be easier for commenters to see how things would look for their favorite modules, which might be easier overall to be able to see the changes without having to figure out how to build the HTML.

What do you think?

rgagnon24:
I've posted the results of building the documentation with this .css file to http://opensips.telepointglobal.com/html

To be thorough, I also re-built the README files and put those here:
http://opensips.telepointglobal.com/readme

PR #442, opened 2015-03-27
Fix get_redirects(max, reason)

rgagnon24:
get_redirects(max, reason) did not properly send the value of reason to the accounting function resulting in a NULL reason code, and a reason of "Unspecified" in the accounting log, db, aaa or other storage.

This properly sends the reason to the accounting function as ```pv_elem_t```, instead of the ```acc_param```

Documentation is also updated to reflect how ```reason``` should be formatted.


bogdan-iancu:
@rgagnon24 , thank you for your patch - indeed, a good catch !!!

PR #446, opened 2015-03-31
Registrar: Add contact-based aliasing when the remote UA supports it.

dsandras:
If an endpoint supporting TCP connection reuse registers to OpenSIPS, we
now create TCP connections aliases for all IP/ports announced by the UA.

From my point of view, it is more a fix than a new feature.

Without this patch, OpenSIPS will only reuse connections targetted at the IP / ports in the Via, which correspond to the actual currently active connection. But most of the time, the proxy will not use those IP/ports when communicating with the UA. It will instead use the registered IP/ports. We should automatically reuse the active connection for all communications with those IP/ports instead of creating new connections, especially in 1.8 where it is a blocking operation.

If the connection can not be reused, then OpenSIPS will create a new one.

### DIFF
--- a/modules/registrar/save.c
+++ b/modules/registrar/save.c
@@ -458,6 +458,13 @@ static inline int insert_contacts(struct sip_msg* _m, contact_t* _c,
 				} else {
 					e_max = e;
 				}
+				if (_m->flags&tcp_persistent_flag && _m->via1->alias && tcp_accept_aliases) {
+
+					if (tcpconn_add_alias(_m->rcv.proto_reserved1, uri.port_no ? uri.port_no : SIP_PORT, uri.proto))
+						LM_ERR("tcp alias failed\n");
+					else
+						LM_DBG("tcp alias contact port %d\n", uri.port_no ? uri.port_no : SIP_PORT);
+				}
 			}
 		}
 #endif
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
We had some long discussions about TCP and the aliasing ...here are some conclusions:

1) You have 2 types of entities involved: network level (ip:port) and SIP level (SIP uris).

2) in SIP, the next hop (upstream or downstream) can be different on network and SIP level: even if I received the INVITE from proxyA, because of RR (or missing RR from A), on SIP level, for sequential requests, I may talk to proxyB ; or during a registration via an edge proxy, the contact may still point to the end-point, but the TCP conn may come from the edge proxy.

3) the existing aliasing (for VIA) is done at network level: network info from VIA is aliased on the TCP conn (everything is network here)

4) what you did in USRLOC is a mixing (in terms of aliasing) of SIP (the contact URI) with net info (the TCP conn) - I understand it fixes your problem, but it is not the correct way to do it. Think about if end point registers via edge proxy (using mpath) -> you will add to the tcp conn (to edge proxy) and alias with the port of the end-device :P.

5) I'm not sure if what you are trying to do is correct (as forcing the tcp conn re-usage) in your particular case: it is clear that the end point is aware of the 2 ports (the uac ephemeral one 30000 and the uas listening one 5060) as it using 30000 in VIA and 5060 in Contact. And this (for me) is an indication that the device explicitly wants to be contacted on 5060, otherwise it should use 30000 too in the Contact URI. So you may try to fix something that is not broken :) (of course, considering non NAT scenarios)

6) still, there is another problem left, when opensips behaves as a proxy (and not as registrar) : A -> proxy -> B , where A and B are end points. If A use 30000 in VIA (and net level) and contact 5060 (as in your case), proxy will create a new conn when routing a BYE from B to A (as B will use A's contact, with 5060 port). and if there is no registration involved (like doing presence), there will be alias between 30000 and 5060......

Until we will come up with a complete and correct solution, we decided not to accept your patch because:
     - it is mixing the SIP and network levels (works only for particular cases)
     - it is bogus when there is an edge proxy with mpath between the uac and registrant
     - it is not complete (does not cover the case when doing routing without registration).

We do not what to upload the temporary hack (which may prove into wrong direction) until there is a clear and complete solution.

In the mean while, what I recommend you as workaround (that at least is consistent with not mixing the SIP and network level and has no drawback) -> before save() use fix_nated_register(), see:
    http://www.opensips.org/html/docs/modules/1.11.x/nathelper.html#id294034

This will not change your message, but it will create a SIP aliase between the Contact URI and the outbound URI (where the Register came from):  sip:ip:5060-> sip:ip:30000, so it should solve your problem.

bogdan-iancu:
@dsandras , please open a bug report on this matter, so we have it on the radar for fixing.

PR #456, opened 2015-04-08
Added pike mi command  for unblocking IP addresses

jarrodb:
```
# opensipsctl fifo pike_list
78.46.64.50
# opensipsctl fifo pike_rm 78.46.64.50
# opensipsctl fifo pike_list
# opensipsctl fifo pike_rm 78.46.64.50
404 Match not found
# opensipsctl fifo pike_rm 78.46.
500 Bad IP
```

### DIFF
--- a/modules/pike/doc/pike_admin.xml
+++ b/modules/pike/doc/pike_admin.xml
@@ -1,15 +1,15 @@
 <!-- Module User's Guide -->
 
 <chapter>
-	
+
 	<title>&adminguide;</title>
-	
+
 	<section>
 	<title>Overview</title>
 	<para>
 		The module provides a simple mechanism for DOS protection - DOS based
-		on floods at network level. The module keeps trace of all (or selected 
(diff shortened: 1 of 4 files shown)


etamme:
Awesome!

bogdan-iancu:
Shouldn't be a better approach to reset the red flag and the counters instead of actually removing the tree node?
Thanks, bogdan

etamme:
This came up when I was doing some rollout to production with pike module.  Some actual customers were getting pike blocked, and I wanted to unblock them.  Jarrod decided to code it up after we talked about it.

We ended up having to put an iptables rule in to drop the customer entirely, then remove it, then drop them, then remove it until we could get over the pike threshold.

I think pike_rm should be that literal - this is not a soft unblock, it is when you want to remove the ip in the pike tree.

If there was a "soft unblock" command, then I agree it would be better to reset counters etc.  But for the situation where traffic has been blocked that you do not want blocked - I want to remove the ip from the tree.

bogdan-iancu:
Thank you !!!

PR #462, opened 2015-04-13
Verify the dialog mi parameter is not null before passing to dlg_hash

jarrodb:
Reported (and verified) by Satish Patel on the developers mailing list that issuing the following command crashes opensips:

opensipsctl fifo dlg_list to_uri::

The p1 value is never verified to not be NULL before calling dlg_hash which crashes the program.  Perhaps there is a better way to resolve this problem, but here is my pull request.


satishdotpatel:
I have test above patch and it works!! 

[root@sip opensips]# opensipsctl fifo  dlg_list to_uri::
400 Invalid Call-ID specified

Issue #420, opened 2015-03-02
Allow TCP without transport=tcp header


oteren:
Hi.

I am pondering if Opensips might be enforcing some stuff that makes RFC compliancy a bit inconsistent.

The problem is that the major video platform Cisco VCS does not by default consider an URI with and without transport=tcp the same. At first I figured that had to be RFC breaking, but the RFC actually states in RFC3261 19.1.4:

``` 
        A URI omitting any component with a default value will not
        match a URI explicitly containing that component with its
        default value.  For instance, a URI omitting the optional port
        component will not match a URI explicitly declaring port 5060.
        The same is true for the transport-parameter, ttl-parameter,
        user-parameter, and method components.
```

As far as I can see, there is nothing in any of the RFCs that it is *disallowed* to send a tcp request *without* the transport=tcp header.

Meaning that the Cisco VCS that uses TCP first and foremost will send an INVITE on tcp without the transport header, but anything going via an Opensips-instance will have transport=tcp added onto it.

This makes the two systems very hard to keep in step.

I figured that instead of having a philosophical debate on whether or not this and that is RFC compliant, if we could for opensips 2.1 have either a config that makes tcp (without the transport header) the first class citizen or a method similar to kamalios to_tcp() (which is implementation wise *not* what we want, since what it does is just to tack on transport=tcp) that makes opensips use the tcp srv record instead of the udp one.

We would still like to keep a sort of cascading behavior so that for example we can prefer that (with no NAPTR records) it first tries TLS, then TCP and then UDP srv records.

I know that if you have the right priorizations in the NAPTR records, opensips will go tcp before udp, but as an operator with a diverse client base we have no way of ensuring that our clients have NAPTR capable dns providers (not even Amazon Route53 currently supports NAPTR). Not to mention that NAPTR use is almost non-existent 'out in the wild' of external platforms our users will call that we have no control over at all.

In general, this change of behavior would also encompass the changes in RFC5630 that has deprecated the use of transport=tls header (since tls should be transport-independant between sctp and tcp) if a to_tls() method was exposed.

oteren:
A follow up on this issue:

If you set up the tcp and sips srv records but **not** the udp one, opensips will reply `SIP/2.0 476 Unresolvable destination (476/TM).` and give up.

It does not cascade down to tcp and sips at all.

Tested with a simple INVITE and t_relay().

For what its worth: using t_relay(tcp:somedomain:5060) works as intended and uses tcp without the transport header.

[...]

PR #474, opened 2015-04-22
resolve inet6 records for hosts and prefer them if dns_try_ipv6 is set

jarrodb:
Previously, if `dns_try_ipv6` was set, resolvehost would only attempt to resolve the ipv6 address if the ipv4 lookup returned no results.  Now if `dns_try_ipv6` is true, both address families are attempted and ipv6 is preferred (this seems to be the logic of most software with ipv6 support enabled).

Also, now that both are resolved, I envision being able to override the default family from the script as well as enable failover should the ipv6 endpoint timeout.  This would follow with future pull requests.

### DIFF
--- a/resolve.c
+++ b/resolve.c
@@ -447,54 +447,58 @@ query:
 
 inline struct hostent* resolvehost(char* name, int no_ip_test)
 {
-	static struct hostent* he=0;
+        static struct hostent* he=0;
 #ifdef HAVE_GETIPNODEBYNAME
-	int err;
-	static struct hostent* he2=0;
+        int err;
+        static struct hostent* he2=0;
 #endif
-	struct ip_addr* ip;
-	str s;
+        struct ip_addr* ip;
+        str s;
 
-	if (!no_ip_test) {
-		s.s = (char*)name;
-		s.len = strlen(name);
+        if (!no_ip_test) {
+                s.s = (char*)name;
+                s.len = strlen(name);
 
-		/* check if it's an ip address */
-		if ( ((ip=str2ip(&s))!=0)
-			|| ((ip=str2ip6(&s))!=0)
-		){
-			/* we are lucky, this is an ip address */
-			return ip_addr2he(&s, ip);
-		}
-	}
+                /* check if it's an ip address */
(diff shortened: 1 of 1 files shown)


etamme:
very much looking forward to discussion/merging of this commit :)

bogdan-iancu:
@jarrodb , the idea is good, but isn't that a bit of a performance killer by doing 2 DNS queries (if v6 enabled) ?
Actually not sure how to read your "both address families are attempted and ipv6 is preferred" - you actually do v4 and v6 and then look into response ? or first v6 and if no response -> v4 (so sequential).

etamme:
It looks like the commit actually attempts both - ipv4 first, then if ipv6 is enabled do ipv6.  I think @bogdan-iancu is right and the logic should be serial as follows:

 if (ipv6_preferred==true){
   do IPV6 lookup
   if (no IPV6 result){
     do IPV4 lookup
   }
} else {
   do IPV4 lookup
   if (no IPV4 result && IPV6 enabled) {
     do IPV6 lookup
   }
}

jarrodb:
I guess I had intended on providing the results for both address families down the road to enable user to have more granularity with controls, but now that you mention it, I agree that should not be the default behavior.  I can update to @etamme's suggestions

jarrodb:
This should take care of the performance issue.  Also, we were talking and wondered about the best way to set the chosen address family for the destination so we can offer the proper SDP from the branch_route.  Perhaps adding a new pvar and setting it once this determination is made?

jarrodb:
With this patch, I can perform the following to determine (most likely) which destination will be chosen, either v4 or v6, by doing the following:

```
$var(x) = $rd;
$(var(x){ip.resolve})
```

Of course this will result in a double query, but it works for the time being.

bogdan-iancu:
OK, let me look into the patch

bogdan-iancu:
Thank you @jarrodb . By the next release we need to look into a way to set priorities (family) from script. For sure we need to do a wider thinking on that, as same kind of priorities (+fallback) is needed for the protocol (see #420)

Issue #471, opened 2015-04-18
modules/db_sqlite: error loading drouting 


jarrodb:
##### opensips.cfg 
```
# ----- drouting -----
modparam("drouting", "db_url", "sqlite://var/db/opensips")
```
##### opensips.log
```
Apr 18 15:39:39 localhost opensips: INFO:drouting:dr_init: Dynamic-Routing - initializing
Apr 18 15:39:39 localhost opensips: ERROR:db_sqlite:db_sqlite_query: failed to prepare: (no such table: version)
Apr 18 15:39:39 localhost opensips: ERROR:db_sqlite:db_sqlite_get_query_rows: failed to prepare query
```
#####  opens incorrect sqlite db handler:
```
# ls -al /var/db
-rw-r--r--  1 root root      0 Apr 18 15:28 opensips?  <-- this is created
```

##### gdb
```
(gdb) break modules/db_sqlite/dbase.c:162
(gdb) print _h->url
$2 = {s = 0x7ffff7431bd0 "sqlite://var/db/opensips8", len = 24}
(gdb) print query_holder
$3 = {s = 0x852740 <sql_buf> "select table_version from version where table_name='dr_gateways's", len = 64}
167		if (ret!=SQLITE_OK)
(gdb) s
168			LM_ERR("failed to prepare: (%s)\n", sqlite3_errmsg(CON_CONNECTION(_h)));
(gdb) print ret
$4 = 1
```

##### Synopsis
Seems like the pointer arithmetic in val2str for type DB_STR (modules/db_sqlite/val.c:229) is off (the trailing s is from the previous sql_buf that dialog populates however the length is correct), but the other issue is its not accessing the proper db_url string that is set correctly.  It's weird that the other modules do not seem to have this issue.  

For example, dialog is the first to load, and the gdb printout of _h->url returns the same data for its query and everything else seems ok.  Perhaps there is an invalid or lacking CON_<X> macro call?

bogdan-iancu:
@jarrodb , do you have the "version" table created in your DB ? as the drouting module is looking into it to check the versions of the dr related tables.

jarrodb:
Absolutely, every other module finds its version properly and the SQLite query built (query_holder) runs successfully on my database. 

However, the real problem is its trying to access the wrong SQLite path as noted by the newly created file in the same path as my actual database when the SQLite library attempts to read.

ionutrazvanionita:
[...]

PR #475, opened 2015-04-22
append '\0' null character to the end of db_url in macro init_db_url

jarrodb:
I guess most of the time this doesn't matter since other libraries also ask for the length along with the `db_url` string, but libsqlite3 is unfortunately not one of those libraries.  This resolved issue #471 for me, but it changes a macro used everywhere else.

### DIFF
--- a/db/db.h
+++ b/db/db.h
@@ -454,6 +454,7 @@ int estimate_available_rows( int payload_size, int column_count);
 		} else {\
 			_db_url.len = strlen(_db_url.s); \
 		} \
+		_db_url.s[_db_url.len+1] = '\0'; \
 	}while(0)
 
 


db/db.h
```diff
@@ -454,6 +454,7 @@ int estimate_available_rows( int payload_size, int column_count);
 		} else {\
 			_db_url.len = strlen(_db_url.s); \
 		} \
+		_db_url.s[_db_url.len+1] = '\0'; \
```
razvancrainea: I think this is a bit broken:

1. _db_url.s - is already NULL terminated, because otherwise strlen() would not work
2. you are adding '\0' after the end of the string (the '\0') - this is a buffer overflow.
3. there is one branch where _db_url.len is not initialized - if there is no DB specified and no default one

I think we'd better investigate the root cause :)

ionutrazvanionita:
fixed https://github.com/OpenSIPS/opensips/commit/0b6402cf96727f4fefc268fc8d5d7069c01b1327

db/db.h
jarrodb: Lol, where is your sense of adventure @razvancrainea ???

Issue #482, opened 2015-04-28
sip_resolvehost srv logic


jarrodb:
This may be a non-issue, but I have been using `sip_resolvehost` in an attempt to create fall-through scenarios based on protocol preference, so I am just trying to understand the logic.

When a protocol is specified (`tls` for example), but a port is omitted when calling `sip_resolvehost`, an attempt is made to resolve the SRV record (`_sips._tcp.domain.com`).  If the SRV record does not exist, it simply tries to resolve the A record for the domain, but then preserves the attempted protocol and default port for that protocol in the return.

Here is a pseudo example of the flow:

1) sip_resolvehost(*name="opensips.org", *port=0, *protocol=PROTO_TLS)
2) attempt to find srv record _sips._tcp.opensips.org
3) no record exists
4) set the port to 5061 (the default for PROTO_TLS)
5) resolve the A record for opensips.org
6) return the results

It seems it would be better, if falling through to an A (or AAAA) record to set the port and protocol to a default:

```
diff --git a/resolve.c b/resolve.c
index 86ac6e5..8a5686c 100644
--- a/resolve.c
+++ b/resolve.c
@@ -1835,7 +1835,7 @@ do_srv:
                LM_WARN("domain name too long (%d),"
                        " unable to perform SRV lookup\n", name->len);
                /* set defaults */
-               if (port) *port = (is_sips)?SIPS_PORT:SIP_PORT;
+                if (port) *port = SIP_PORT;
                goto do_a;
        }
 
@@ -1876,7 +1876,7 @@ do_srv:
        LM_DBG("no valid SRV record found for %s, trying A record lookup...\n",
                tmp);
        /* set default port */
-       if (port) *port = protos[*proto].default_port;
+       if (port) *port = SIP_PORT;
 
 do_a:
        /* do A record lookup */
```

Or, if requesting a **specific** protocol, which was **not** found, then return 0 so a decision could be made upstream.

I may be way off base, just trying to see if this was the intended result.

bogdan-iancu:
@jarrodb , the sip_resolvehost() takes the proto parameter as a must. Whatever proto is requests by upper layer must be obey. This function must not change the proto (for whatever reasons) .
If the upper layer wants TLS (because of forced from script, or because the the URI indication), the function must find (DNS based) a destination for TLS. If an SRV record is not present should be an error - maybe there is a simple setup where SRV is not used. But the service accepts TLS via A record + default SIP port.
[...]

PR #483, opened 2015-04-28
(discussion only) uri_resolve method build dest uri with SRV support

jarrodb:
This pull request is definitely not finished and probably not even the right direction so I did not want to spend a lot of time on it, but wanted to continue the discussion from previous PR #474.  This exposes do_srv_lookup from resolve.c poorly because of issue #482.

This PR is attempting to solve 2 issues per the discussion:
1) Build destination uri from a preferred protocol list that will fall-through until it finds a result.
2) Be able to identify the address family of the host based destination within the script for offering IP6 or IP4 SDP.

The first argument to the method will be populated with the phostport;protocol string as a result from the queries.

The second argument is a semi-colon separated list of protocols to try.  As the protocols are tried, they are removed from the list so that the remaining can be retried from a failure_route.  If not SRV records are found, it tries to simply resolve the A or AAAA record for the host.

Here is an rudimentary example of how I've been testing it:

```
branch_route[rtpengine] {
  $avp(plist) = "tls;tcp;udp";
  if (uri_resolve("$avp(du)", "$avp(plist)")) {
    $du = $avp(du); # set the $du to our result
    t_on_failure("1");

    if ($du =~ "^sip:[0-9]{1,3}\..*$") {
      $dlg_val(uas_ipv)="IP4";
    } else {
      $dlg_val(uas_ipv)="IP6";
    }
  }
  ...
}

failure_route[1] {
  if (!t_check_status("(5..)|(6..)|(403)")) {
    xlog("L_NOTICE","$var(request_log) $rs not acceptable failure\n");
    exit;
  }

  if ($avp(plist)) {
      # The first protocol failed, go to the next one.
      if (uri_resolve("$avp(du)", "$avp(plist)")) {
        t_on_failure("1");
        $du = $avp(du);
      }
  } else {
    exit;
  }
}

```

### DIFF
--- a/modules/uri/methods.c
+++ b/modules/uri/methods.c
@@ -1,0 +1,144 @@
+/*
+ * Various URI checks and Request URI manipulation
+ *
+ * Copyright (C) 2001-2003 FhG Fokus
+ *
+ * This file is part of opensips, a free SIP server.
+ *
+ * opensips is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ * opensips is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "../../error.h"
+#include "../../pvar.h"
+#include "../../resolve.h"
+#include "../../socket_info.h"
+#include "db_checks.h"
+#include "methods.h"
+
+#define PLIST_DELIM ";"
+#define PHP_BUFFER_SIZE 4+2+INET6_ADDRSTRLEN+1+5+10+5
+
+/* Resolves R-URI host based on sip_resolvehost logic */
+int uri_resolve(struct sip_msg* msg, char* _php, char* _plist) {
+        static char resolve_buf[INET6_ADDRSTRLEN];
+        static char php_buf[PHP_BUFFER_SIZE];
+	static char tmp[MAX_DNS_NAME];
+	char pbuf[5];
+	char* save = NULL;
+	char* p;
+	unsigned short port;
+	int proto;
+	pv_value_t pv_plist;
+	str* name;
+        struct hostent* he;
+	int_str php;
+	int_str plist;
+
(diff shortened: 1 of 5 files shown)


jarrodb:
To help explain the pull request and further the discussion...  I suppose some of this type of logic could be incorporated into the current sip_resolvehost method where it referenced a global variable for the preferred protocol ordering attempts

However, there are some benefits of having a script pvar populated with this list and used for the resolving logic:
1) If you are keeping track of the protocols and what's been attempted, it's easy to failover to the next protocol in the list.
2) If, for example, TLS is first in the list but a domain does not have TLS records, the end user could choose to cache an updated protocol list for the destination host so that it skipped TLS on subsequent attempts.  It does seem inefficient to prefer TLS and attempt it on every SRV record, but have 90% of the destinations not support it.
3) Furthering the first point, it provides a mechanism for failing over, not just for protocols (tcp->udp) but also accounts for weight and priority within the same protocol.

As an example for point 3, consider the following results of an SRV query with multiple weighted records:

_sip._tcp.opensips.org.	3600	IN	SRV	0 50   5060 romania.opensips.org.
_sip._tcp.opensips.org.	3600	IN	SRV	0 100 5060 usa.opensips.org.
_sip._tcp.opensips.org.	3600	IN	SRV	0 150 5060 outerspace.opensips.org.

Since tcp returned results, the tcp protocol would be shifted from the pvar protocol list, but since there were multiple results returned for the SRV record (hostent->(char **)h_addr_list) we would rewrite the list to something like this "tcp:0:100;tcp:0:150;udp".  If romania timed out and failure_route was entered, we would pass the updated protocol list pvar back to the command and it would know the next result to try would be the tcp record with 0 priority and 100 weight and so on.

Lastly, we need to be able to determine the destination uri IP address **before** the request_route or branch_route (or failure_route?) so we know the address family of the destination.  Again, this aids in offering IPv6 or IPv4 SDP in the initial INVITE request which I believe is necessary.  Perhaps there are other benefits for having this information as well?

Anyway, those are my thoughts to continue the discussion and sheds light on why I chose to build a new script command as opposed to modifying the existing sip_resolve logic.  Although I do think sip_resolvehost will need to be modified in order to accomplish this.  If I can get some help in determining a decent plan that everyone likes, I can come up with an initial pull request to try to accomplish this.

etamme:
I agree with points 1-3, as well as the follow on that it would allow for multiple attempts against weighted SRV records.

I am not sure I follow that we need to know the destination protocol/address family before the main request route - I would say that would not be useful as the request could end up forking via append_branch mechanisms.  Knowing in branch_route IS crucial, and any processing that is happening based on a destination should happen in the branch route father than the request route block.

In the case of failure route, you need to know what AF/protocol you had attempted, but this can be achieved via transactional storage mechanisms: avp, flags etc.  If you t_relay from failure_route it will hit the branch route again for further branch processing.

I assume in your example that in the failure route you have modified plist to remove the previously attempted protocol from the list before passing it to uri_resolve?

It seems like a reasonable approach to me in general.

bogdan-iancu:
Hi all ! Following the discussion on IRC, let's re-evaluate all the possibility for the proto / family failover (DNS only, signaling auto, signaling manual) and come up with a consistent proposal that cover all the cases. I will try to put together a starting point.

PR #517, opened 2015-05-19
Make sure httpd can recognize Content-Type headers with parameters.

staskobzar:
With improved memory management patch from bogdan-iancu.


bogdan-iancu:
@staskobzar , you need to keep the strcasecmp() instead of strncasecmp(). If you get a Content-Type: text/xmlX ; bla=foo ; after param stripping and trimming, the "value" will be 'text/xmlX', that will match "text/xml" with strncasecmp() and will not with strcasecmp(). And it should not match.

staskobzar:
@bogdan-iancu Right! Thank you for explaining that in details. I have updated functions.

Issue #525, opened 2015-05-28
NULL support in cachedb raw interface


jayesh1017:
The cachedb raw queries currently ignores the NULL values that gets returned from a NoSQL Database. For eg. if you do the following:
cache_raw_query("redis:myRedis", "HMGET MYKEY one two three", "$avp(result)");
And if the the key "two" doesn't exist in redis, the avp array element 1 does not return with a NULL value, instead the value at element 2 is substituted at element 1 and if we do value matching in the script it has possibilities of error.
It'll be great, if NULL returns back so that value matching of array elements in $avp(result) is done properly.

vladpaiu:
Hello,

This has been fixed in the master branch ( thanks to Jarrodb ) - please update your sources and retest.

Best Regards,
Vlad

PR #527, opened 2015-05-28
Add support for hiredis type REDIS_REPLY_NIL in redis_raw_query

jarrodb:
Initial attempt to address #525


vladpaiu:
Hello,

Thanks for the patch - merged ( also fixed a small typo ).

Best Regards,
Vlad

jayesh1017:
Hello,
Although this works, but if I query for a key that doesn't exist in redis, opensips throws out a error in logs something like this:
ERROR:core:new_avp: id: 85 s- NULL VALUE!
This is harmless as far as I tested, but just fills up the logs unnecessarily.

PR #530, opened 2015-05-29
add width string transformation

jarrodb:
this adds a string transformation primarily used for making pretty logs.  for example:

```
xlog("L_INFO","$var(prefix): $(cs{s.width,7}): $(rm{s.width,9}): request from uac behind nat\n");
```

i looked into use `$(cs{s.fill.right, " ", 7})` but it wasn't meant for that it seems.


liviuchircu:
The **s.fill** transformations have been tuned a bit to also work exactly as the proposed **s.width**. 

Performance-wise, it will work exactly as **s.width** (memset if possible). Moreover, you can do left-alignment with spaces as well.

jarrodb:
Would it also be beneficial to have width as well, since it can also truncate longer strings to a specified width?

liviuchircu:
IMO, as long as they have decent applicability (i.e. pretty-logs for Call-IDs in this case), such transformations are nice additions. Also, **s.width** is shorter to write.

PR #534, opened 2015-05-30
fix rpm build error on centos

seanchann:
README.python unpack  when build rpm package


bogdan-iancu:
The python module had no README file until the time of the 2.1 release when it was added.

bogdan-iancu:
@seanchann , thank you for your fix !

PR #539, opened 2015-06-03
Remove generating error message on creating avp with null value

jarrodb:
#527 creating a null AVP should not generate an error since we are now supporting that type - AVP_VAL_NULL.


jarrodb:
@vladpaiu resolved with commit eea8476b54a80704904e33d0fa9332d2a82dcf0f

PR #545, opened 2015-06-09
fixes for Debian packaging

dynamicpacket:
* Check for freeradius libs in Makefile
* Check for json lib include paths in Makefile
* Update debian/changelog to reflect 2.1.0 stable
* Update debian/control with new dependency names and standards
* Update debian/rules to remove deprecated dpatch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,13 @@
 #
 
 #FREERADIUS=1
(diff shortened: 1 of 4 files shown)


bogdan-iancu:
Many thanks @dynamicpacket

PR #548, opened 2015-06-09
fixes for Debian packaging

dynamicpacket:
* Check for freeradius libs in Makefile
* Check for json lib include paths in Makefile
* Update debian/control with new dependency names and standards
* Update debian/rules to remove deprecated dpatch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,13 @@
 #TLS=1
 #SCTP=1
 #FREERADIUS=1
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Many thanks @dynamicpacket 
Even if this branch (1.10) is not supported any more, the packaging specs should be in place as this version may still be in use.

PR #562, opened 2015-06-23
Fixed counter value when the key is not found.

dsandras:
It can happen that a counter value is queried before it has been set. In
that case, we should assume a default value of 0.

A typical use case is the dialog module where we could call
get_dialog_profile for a specific key before set_dialog_profile has been
called for that key. There should be no error in that case and we should
simply return a value of 0.


dsandras:
This is indeed a far better fix. Thanks !

PR #586, opened 2015-08-03
add 'force_send_address' to support 'pseudo-variables'

was4444:
add `force_send_address' to support `pseudo-variables'


bogdan-iancu:
Hi @was4444 , is this force_send_address() function intended to be the same as force_send_socket() but with variables as param ?

was4444:
Yes, @bogdan-iancu, that's all I want.
But I don't know whether this function will reduce the performance of OpenSIPS, so I write another one to not effect the original too much.
I have two interfaces, when use **mhomed** parameter in configure file, I find OpenSIPS could not select the right interface, at least for me, when relay ACK requests(I use **record_route** to make all in-dialog messages sent to OpenSIPS).

bogdan-iancu:
Have you looked at $fs variable ? it does exactly what you need : 
   http://www.opensips.org/Documentation/Script-CoreVar-2-1#toc44

was4444:
Oh, yes, that's what I need.
I should upgrade to 2.1 some time.

bogdan-iancu:
@was4444 , $fs is available in older versions too, so no need to upgrade. 
I will close this PR for now, as it brings no new addition as functionality.

Issue #593, opened 2015-08-07
Updating parts of $du results in malformed duri


brettnem:
So I'm using d9eec5868da2501a5071185aaa4b0e65cf386bdc 

If I do this:
$du=$ru;
xlog("du->$du");

du looks right: say like this:
sip:15125551212@1.2.3.4:5060;user=phone

** THERE IS NO EXTRA COLON IN THERE,  I PROMISE **

However, if I modify $du piece parts like this:
$du=$ru;
$dp=5061;

dp is changed, but du is now malformed
sip:15125551212:@1.2.3.4:5061;user=phone

Note: port 5061 as expected, but a COLON is inserted before the @ 
Note2: If I make 5061 a string like '5061' it does the exact same thing

Similarly if I change the domain ONLY  I get the same weird colon

$du=$ru
$dd='example.com'

gives me:
sip:15125551212:@example.com:5060;user=phone

I have no idea how the colon gets in there

It's always ONE COLON before the @. I haven't see any other case

Now if I change BOTH $dd and $dp:
$du=$ru
$dp=5061
$dd='example.com'

The exact same problem:
sip:15125551212:@example.com:5061;user=phone

I don't doubt the possibility of some dynamic other variable or other part of my script causing this to happen, but I can't; figure out how or why

I've tried to debug on level 6, but because of a different bug, more than half my AVPs are NULL during the script and everything fails.

liviuchircu:
All three scenarios confirmed. You can even replicate this within a startup route!

liviuchircu:
Looks like it's printing the **password** as well. However, the password is NULL, so it seems like it's just appending a colon.

Nevertheless, the resulting uri "sip:15125551212:@example.com:5061;user=phone" seems valid according to http://www.in2eps.com/fo-abnf/tk-fo-abnf-sipuri.html.

brettnem:
Even tho this may pass abnf, when all this happens, it appears that t_relay doesn't actually send the call out. Note that I'm manipulating my $du in branch_route

brettnem:
Also in reviewing the abnf, isn't it suggesting that if you have a : you must have a password? 

Also  [userinfo] itself is optional. but if you have a [userinfo] you have to have a user.. then user can optionally have :password 

But I don't see how : <no password> without user info or an actual password would validate?

liviuchircu:
Because **password** is defined as ***** (**zero** or more) of its specific characters.

brettnem:
[...]

PR #598, opened 2015-08-12
Do not append colon if uri password is NULL

jarrodb:
I agree with @brettnem's assessment of the problem with #593.  As @liviuchircu pointed out, this pull requests addresses the problem in `do_action.c:1032` where a colon is blindly added if the user exists without consideration for the presence of an actual password... it attempts a memcpy on a NULL pointer no less (albeit with a length of 0).

I think if `parse_uri(tmp, len, &uri)<0)` finds an actual semi colon after the user and the password IS indeed blank, uri.password.s should be initialized to an empty string `""` instead of NULL so that `do_action.c:1032` check passes and the current "problematic" behavior will occur.  But otherwise, if the password is NULL, the colon should not be appended.

This should satisfy the userinfo requirements as well as fix the issue presented.

Thoughts?

### DIFF
--- a/action.c
+++ b/action.c
@@ -1029,7 +1029,7 @@ int do_action(struct action* a, struct sip_msg* msg)
 			/* passwd */
 			tmp = uri.passwd.s;
 			len = uri.passwd.len;
-			if (user || tmp) {
+			if (tmp) {
 				if (crt+len+1>end) goto error_uri;
 				*crt++=':';
 				memcpy(crt, tmp, len);


liviuchircu:
Although current behavior barely passes RFC requirements (i.e. it's **correct**), a SIP server should try to be as permissive as possible when *receiving*, yet strictly adhere to recommended RFC behavior when *sending*. And appending unnecessary colons to URI usernames doesn't seem to be too recommended.

The fix looks great, since it still propagates existing empty-string passwords, if script writer *really* needs this. Many thanks, @jarrodb !

PR #602, opened 2015-08-17
Send socket fixups

eseanucristian:


bogdan-iancu:
Dropped as included in #607

PR #608, opened 2015-08-24
two bugs fixed in db_text module

shripchenko:
+ incorrect use of strncasecmp()
+ lost pkg_malloc


bogdan-iancu:
Thank you @shripchenko ! Your patch was accepted and backported to 2.1 and 1.11 too.

PR #609, opened 2015-08-24
New MI command dbt_reload - causes db_text module to reload cached tables from disk.

shripchenko:
Depending on parameters it could be a whole cache or a specified database or a single table.
If any table cannot be reloaded from disk - the old version
preserved and error reported.


bogdan-iancu:
Thank you @shripchenko !

PR #610, opened 2015-08-24
A new flag for LB module lb_start() function to pick a rand()'om destinations with equal load

shripchenko:
...if multiple destinations selected.

This could help to offload an excessive load from first destination and distribute load in situations when calls duration is near to zero (like failed calls)
and counts in resources are almost always zero and does not reflect an actual calls flow.

Already discussed in PR: OpenSIPS/opensips#345


bogdan-iancu:
Thank you @shripchenko

PR #619, opened 2015-08-26
other fixes on db value interpretation in dialog module

qmphan:
@bogdan-iancu : There are other places in dlg_db_handler.c where VAL_STR().len is used. This one fix them all.

### DIFF
--- a/modules/dialog/dlg_db_handler.c
+++ b/modules/dialog/dlg_db_handler.c
@@ -573,7 +573,7 @@ static int load_dialog_info_from_db(int dlg_hash_size)
 			/* script variables */
 			if (!VAL_NULL(values+18))
 				read_dialog_vars( VAL_STR(values+18).s,
-					VAL_STR(values+18).len, dlg);
+					strlen(VAL_STR(values+18)), dlg);
 
 			/* profiles */
 			if (!VAL_NULL(values+19))
@@ -1378,7 +1378,7 @@ static int sync_dlg_db_mem(void)
 				/* script variables */
 				if (!VAL_NULL(values+18))
 					read_dialog_vars( VAL_STR(values+18).s,
-						VAL_STR(values+18).len, dlg);
+						strlen(VAL_STR(values+18)), dlg);
 
 				/* profiles */
 				if (!VAL_NULL(values+19))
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @qmphan for the follow up on the issue. 
The problem is more complex as the values may contain a 0 byte, so strlen() may return shorter strings. The original fix for 1.8 covered only mysql unfortunately. We will follow with another fix to make the value B64 encoded to avoid problems.
Starting with 1.11 this was solved by moving to BLOD data type.

qmphan:
@bogdan-iancu: Would the use of DB_STR type and VAL_STR(...).len resolve the problem?

bogdan-iancu:
Not really, as the the .len (for DB_STR) is not populated by the db_postgres driver.

qmphan:
If you pull in the  #618, db_postgres driver will populate the .len ;)

bogdan-iancu:
Maybe I'm missing something, but that PR changes only the type of the column from STRING to STR, but the actual value returned in "res" is the same....without the .len filled in.

qmphan:
When values are read from DB, they are converted to C structures by going through these functions:
db_postgres_fetch_result  --> db_postgres_convert_rows  --> db_postgres_convert_rows  --> db_postgres_str2val
Inside the last function (db_postgres_str2val), if the DB type is DB_STR, the len field will be filled.

The PR #618 set the DB type to DB_STR so that db_postgres_str2val will fill the .len

bogdan-iancu:
@qmphan , you are right!  The only issue I see with that approach is that we will break many other modules. Most of them do expect to get a DB_STRING after a DB query - and they actually check the DB type. If we change to DB_STR, all these modules will get broken.
What can be done is to : 
1) revert the so far changes in dialog module
2) make db_postgres to populate the .len even if the DB_STRING (in db_postgres_str2val).

qmphan:
Yeah, that's the safest way! I didn't see that there are checks for DB_STRING type in other modules.

qmphan:
These changes are made in #622

bogdan-iancu:
Thanks @qmphan , let me review and upload.

PR #621, opened 2015-08-27
lock_ops.h: added support for debugging locks

ionel-cerghit:
with the flag DBG_LOCK set(from menuconfig for example), with each
lock will be stored information regarding the place from where the
lock was aquired (file, function, line)


liviuchircu:
Really looking forward to using this. It's the type of easily overlooked feature that can turn out to be a huge time saver if the developer/platform maintainer is aware of it!

PR #623, opened 2015-08-28
Non printable characters are encoded in hex format

eseanucristian:


modules/dialog/dlg_db_handler.c
```diff
@@ -898,6 +924,77 @@ static inline unsigned int write_pair( char *b, str *name, str *name_suffix,
...
+}
+
+static int encode_string(str *dst, str *src)
+{
+	unsigned char x;
+	int j = 0;
+	char *at = dst->s;
+	char *p = src->s;
+	while (j <= src->len) {
+		if (*p < 32 || *p > 127 || *p == '#' || *p == '|' || *p == '\\') {
```
razvancrainea: 127 is not printable, so it should be caught by this branch. You could test *p against 127, (*p == 127), since *p can never be greater than 127 (max char).

modules/dialog/dlg_db_handler.c
```diff
@@ -905,19 +1002,21 @@ static str* write_dialog_vars( struct dlg_val *vars)
 	static int o_l=0;
 	struct dlg_val *v;
 	unsigned int l,i;
+	static str dst = {NULL,0};
 	char *p;
-
 	/* compute the required len */
-	for ( v=vars,l=0 ; v ; v=v->next) {
+	for ( v=vars,l=0; v ; v=v->next) {
 		l += v->name.len + 1 + v->val.len + 1;
```
razvancrainea: The code below is no longer necessary, since the control chars will be encoded anyway.

PR #629, opened 2015-09-04
Tls rework

eseanucristian:


modules/proto_tls/proto_tls.c
```diff
@@ -1343,7 +256,7 @@ static int tls_conn_init(struct tcp_connection* c)
 		if (!avp) {
 			LM_DBG("no TLS client doman AVP set, looking "
 				"for socket based TLS client domain\n");
-			dom = tls_find_client_domain(&c->rcv.src_ip, c->rcv.src_port);
+			dom = tls_mgm_api.find_client_domain(&c->rcv.src_ip, c->rcv.src_port);
 			if (dom) {
 				LM_DBG("found socket based TLS client domain "
 					"[%s:%d]\n", ip_addr2a(&dom->addr), dom->port);
```
razvancrainea: The tls_client_domain_avp should not be visible in the proto_tls module. The avp search should be done internally in the tls_mgm module and the "domain" should be returned directly.

modules/tls_mgm/tls_mgm.c
```diff
@@ -0,0 +1,964 @@
...
+	{{"tls_peer_notAfter", sizeof("tls_peer_notAfter")-1},
+		850, tlsops_validity, 0,
+		0, 0, pv_init_iname, CERT_NOTAFTER },
+
+	{ {0, 0}, 0, 0, 0, 0, 0, 0, 0 }
+
+};
+
+struct module_exports exports = {
+	PROTO_PREFIX "tls_mgm",  /* module name*/
```
razvancrainea: Why use PROTO_PREFIX? This is not a transport protocol, its just a helper!

modules/proto_tls/proto_tls.c
```diff
@@ -118,23 +116,6 @@ static cmd_export_t cmds[] = {
```
razvancrainea: The is_peer_verified() function should also be moved in the tls_mgm module.

modules/proto_tls/proto_tls.c
```diff
@@ -325,7 +134,7 @@ struct module_exports exports = {
 	params,     /* module parameters */
```
razvancrainea: Since proto_tls now depends on the tls_mgm, a dependency should be put here.

razvancrainea:
don't forget to migrate documentation too!

PR #632, opened 2015-09-10
alias_db module changed to allow alias_db_find() to be called from startup route

shripchenko:
* allow alias_db_find() to be called from startup route (f.e. to pre-populate db_text cache for a table)
* allow alias_db_find(,input,) parameter to be string or PV or mix of them instead of just PV


bogdan-iancu:
Thank you @shripchenko , patch uploaded.

PR #634, opened 2015-09-14
xcap: make the 'doc' column a LONGBLOB on MySQL

saghul:
A normal-sized XCAP document with external references easily exceedes
the default column size, leading to all sorts of problems.

### DIFF
--- a/db/schema/pr_xcap.xml
+++ b/db/schema/pr_xcap.xml
@@ -46,6 +46,7 @@
         <name>doc</name>
         <type>binary</type>
         <description>doc</description>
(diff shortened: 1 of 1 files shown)


saghul:
@OpenSIPS/contributors please review.

Once it lands, mind if I backport it to 1.11?

razvancrainea:
I noticed that you've only changed the type for mysql. After a little bit of digging it looks like Mysql's blob limit is 65535 bytes, whereas for the other backends the limit is higher (more than 1GB).
I'll merge the PR and also generate the SQL schema.

saghul:
Thanks, Razvan!

PR #642, opened 2015-09-18
don't save ruri to Route in handle_sr if there is no next hop

doome:
In after_loose, behave much like in after_strict: don't insert the Route header with our own address if the next hop is the endpoint, and not a strict router.

### DIFF
--- a/modules/rr/loose.c
+++ b/modules/rr/loose.c
@@ -405,9 +405,11 @@ static inline int handle_sr(struct sip_msg* _m, struct hdr_field* _hdr, rr_t* _r
 	int rem_len;
 
 	/* Next hop is strict router, save R-URI here */
-	if (save_ruri(_m) < 0) {
-		LM_ERR("failed to save Request-URI\n");
-		return -1;
+	if (_r->next) {
+		if (save_ruri(_m) < 0) {
+			LM_ERR("failed to save Request-URI\n");
+			return -1;
+		}
 	}
 
 	/* Put the first Route in Request-URI */


bogdan-iancu:
Hi @doome , sorry for late handling on this patch.
I checked a bit the whole logic in handling the routing and IMHO, your change is not correct.

The handle_sr() function is called from after_loose() when the next hop (indicated by Route hdrs) is marked as strict router. Coming via the after_loose() route, it means in RURI we have something else than our own IP (most probably the IP of the end point).
Going back to handle_sr()....as next hop is strict , we need to place its IP in RURI; and not to loose what we have in RURI (the end point), this must be added as a last Route to the message.
You say this insertion (of current RURI as Route) should be do only if there is another Route to follow, but this has no impact actually.

Let's take this example:
>RURI: end_point
>Route: myself;lr
>Route: proxy_sr

after going through after_loose(), the local Route is dropped and the next route will point to "proxy_sr" which is strict route:
>RURI: end_point
>Route: proxy_sr

Now, the handle_sr() should re-arrange the message as the next hop is a strict router. If you do it as per your patch, the RURI will not be saved into Route any more, and only have the proxy_sr moved into RURI:
>RURI: proxy_sr

Which is wrong as you lost the last element in the routing set - the "end_point" ; the next proxy will have no idea where to send the message further.

Currently, without your patch, the result after handle_sr() is:
>RURI: proxy_sr
>Route: end_point

so proxy_sr will know that it has to fwd the request to end_point.

Am I missing anything on your patch ?

Regards,
Bogdan

doome:
Hi Bogdan,

thanks for replying!

Actually, when I made the attached patch, I didn't dig too deep into the handling of strict and loose routing, I just had a problem, which the patch seemed to solve.

The scenario was:

We had an asterisk behind an opensips proxy, and tried to communicate with an other telco. Our asterisk created an INVITE like this:

    INVITE sip:<opensips_address> SIP/2.0
    Route: <sip:phonenumber@telco_address>

After getting through opensips, the packet looked like:

    INVITE sip:<phonenumber@telco_address> SIP/2.0
    Route: <sip:opensips_address> 

The other endpoint was also a proxy, and handled this incoming message wrong: it forwarded the message to some of the their SIP servers, and changed the RURI to `<sip:proxy_address>` which it could not handle. As I mentioned, I didn't dig myself deep into the world of SIP routing, so there is a chance that this problem was actually caused by their proxy's bug.

However, with the patch above, our proxy never send out `Route:` headers to our outer legs. This fixed the problem above, and it seems that nothing else got broken (we use this in production since I attached the patch).

bogdan-iancu:
Hi @doome 
In your case Asterisk acts as strict router (next hop in RURI - opensips - and following hops in Route hdrs - the telco). The bogus routing happened because (for whatever reasons) opensips failed to recognize its own ip in RURI.
If the server IP is in RURI -> after_strict() and next Route hdr  will simply be uploaded in the RURI (which is what you want)..

So the key question is why opensips does not recognize its IP in RURI - do you have in RURI and IP or FQDN ? it is and "advertised" address or a real listener of opensips ?

bogdan-iancu:
Closing as invalid.

doome:
Hi!

The address in the RURI is the very same address which opensips is configured to listen on. It is there as IP, no an FQDN.

PR #662, opened 2015-10-01
load balancer: new parameter to lb_start/next

shripchenko:
Added optional parameter [dsts_cnt] to lb_start/next family of functions that holds a string containing an integer PV (pseudo-variable) where the total number of matching destinations (regardless of their capacity) will be stored.

This could be used to distinguish between 'no capacity' and 'no destinations left' situations upon a failure to find out an available destination.


bogdan-iancu:
Hi @shripchenko  - sorry for the late handling of this PR. Aside making a difference between no capacity' and 'no destinations left', do you see any other usage case of this new "dsts_cnt" param ?

shripchenko:
mmm.
not sure i could provide you more scenarios.
for our usage scenario we had to know, whenever the call failed for 'no capacity' or for 'no destinations left' scanarios, to adjust fail-over scenario.

bogdan-iancu:
@shripchenko , in this case, what about propagating the info "no channels available" or "bad resources" via return code rather than a variable ?

bogdan-iancu:
@shripchenko , any opinion on this ? if not, I will rework the patch to get the functionality based on the return code.

shripchenko:
sure! no prob.
thanks!

PR #674, opened 2015-10-13
packaging/debian: change MI_XMLRPC to MI_XMLRPC_NG in packaging

telematico:
mi_xmlrpc has some problems with multithreading and has been
superseeded by mi_xmlrpc_ng module. Aditionaly needs a very old
version of libxmlrpc-c3 that is not present in debian wheezy or
jessie

This commit change updates Debian packaging files

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -10,7 +10,7 @@ Homepage: http://www.opensips.org/
 Package: opensips
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}, adduser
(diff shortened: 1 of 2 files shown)


razvancrainea:
Hi, Carlos!

I am afraid I do not entirely agree with this pull request request for 1.11. Using a new package (xmlrpcng) rather than the old one (xmlrpc) looks more like a feature/enhancement, and not a bugfix. Although this might help people installing new OpenSIPS 1.11 instances, it will definitely annoy the ones that have stable 1.11 installs for more than 2 years and now need to update their config script if they want to update their version.

However, I think it is a good idea to have a separate package for mi_xmlrpc_ng (probably called opensips-xmlrpcng-module) for people that want to use it. But we need to support the old package (opensips-xmlrpc-module) for backwards compatibility.

My 2 cents,
Răzvan

telematico:
Hi Răzvan!

I think you're definitely right, It's a better idea to support both modules. I close this pull request now and will make a new one with support for creating two diferent packages. Thanks for your help!

PR #678, opened 2015-10-20
fix: some sh(busybox) do not support syntax of $(< file)

was4444:
BusyBox v1.20.2 (2014-12-11 14:58:46 CST) multi-call binary.


liviuchircu:
Thanks, @was4444 !

PR #691, opened 2015-11-05
net_udp: add configurable max number of fds

ionel-cerghit:
added script variable udp_max_sockets to configure max number of udp sockets allowed


bogdan-iancu:
This is to be reworked in a total different way, closing this one.

PR #693, opened 2015-11-09
Maintain database result integrity by supporting null column values

jarrodb:
The avp population of column values from avp_db_query equaling NULL are skipped, sacrificing the integrity of the actual database result.

Basically, the old method checked if the column value structure's nul member was 1 and skipped to the next avp:

```
if (RES_ROWS(db_res)[i].values[j].nul)
        goto next_avp;
```

I moved this check to column values of type DB_INT, since that is the default type, and instead of skipping the avp, assign the value type to AVP_VAL_NULL:

```
avp_type |= AVP_VAL_NULL;
```

### DIFF
--- a/modules/avpops/avpops_db.c
+++ b/modules/avpops/avpops_db.c
@@ -477,9 +477,6 @@ int db_query_avp_print_results(struct sip_msg *msg, const db_res_t *db_res,
 		crt = dest;
 		for(j = 0; j < RES_COL_N(db_res); j++)
 		{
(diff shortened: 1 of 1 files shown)


jarrodb:
I did this too quickly, this is making some assumptions about the data it receives (i.e. that any column value equaling nul with be of type DB_INT) and I am not sure if that is true.

There is probably a safer solution that I'll need to think about, but the issue stands that assigning db results to avps shouldn't skip NULL values.

PR #701, opened 2015-11-24
uac_registrant: avoid bulk re-registrations on large servers

nikbyte:
Keeps call-id, tag and state of registration records on servers where reg_reload used too often. Let avoid hanged registrations and ban for often re-registrations with another call-id on remote servers.


nikbyte:
Thank you, Ovidiu! I've just removed unnecessary loop.

opensips-github:
Hello Nick,

I merged your changes.
Please keep in mind that it would be helpful to perform a rebase on
your branch to keep the logs clean:
git fetch origin; git pull --rebase origin master
or
git fetch origin; git rebase origin/master

This will move all your commits on top and merge commits will be avoided.
Not a big deal, but helps keeping the logs clean.

Thank you for your contribution.

-ovidiu

On Mon, Dec 14, 2015 at 12:20 AM, Nick Altmann <notifications@github.com> wrote:
> Thank you, Ovidiu! I've just removed unnecessary loop.
>
> —
> Reply to this email directly or view it on GitHub.
>
>
> _______________________________________________
> Devel mailing list
> Devel@lists.opensips.org
> http://lists.opensips.org/cgi-bin/mailman/listinfo/devel
>



-- 
VoIP Embedded, Inc.
http://www.voipembedded.com

modules/uac_registrant/registrant.c
```diff
@@ -824,6 +824,34 @@ static struct mi_root* mi_reg_list(struct mi_root* cmd, void* param)
...
+		new_rec->registration_timeout = old_rec->registration_timeout;
+		new_rec->state = old_rec->state;
+	}
+	return 0;
+}
+
+int run_find_same_rec(void *e_data, void *data, void *r_data)
+{
+	reg_record_t *new_rec = (reg_record_t*)e_data;
+	int i = (int*)data;
```
razvancrainea: @ovidiusas could you review this a bit - while looking at the [Travis log](https://travis-ci.org/OpenSIPS/opensips/builds/97166562) I noticed this warning, which actually may lead to a bug.

`data` is a pointer to an integer, that is passed to an integer. This may be a _huge_ value, that is used below to index the `reg_htable`.

My guess is that the line should be `int i = *(int *)data;`, but please give it a look too.

ovidiusas: Thank you for spotting this out!

PR #708, opened 2015-11-26
TLS: CRL+CA list refresh MI command added to tlsops

ph4r05:
TLS subsystem (TLS engine + tlsops module) was extended with functions to reload CA + CRL lists on-the-fly for all configured domains.  

TLSOPS module was extended to provide a new MI command "refresh_crl_ca", calling refresh of CRL and CA files. New connections will be validated with the newly loaded CA + CRLs.

Previously, OpenSIPS does not reload CA list when running. The same holds for CRL in 2.x branch. 

Reloading CRL is of high importance as OpenSIPS can be running several weeks/months without being restarted. If TLS with client certificates is used it is crucial to periodically refresh CRL files and to reload them by OpenSIPS without need to restart it.

### DIFF
--- a/cfg.lex
+++ b/cfg.lex
@@ -357,6 +357,7 @@ TLS_CERTIFICATE	"tls_certificate"
 TLS_PRIVATE_KEY "tls_private_key"
 TLS_CA_LIST		"tls_ca_list"
 TLS_CA_DIR		"tls_ca_dir"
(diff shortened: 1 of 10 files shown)


ph4r05:
as new features are not accepted to 1.11 branch, should I do the same changes also on 2.x branch? Will it be accepted then?

bogdan-iancu:
see #709

PR #709, opened 2015-11-26
TLS: Client authentication with client certificate CN

ph4r05:
TLSOPS module was extended to add support for checking correspondence between FROM/TO URIs and CN of the client certificate used for TLS connection = client authentication via client certificates.

For clients using TLS client certificates this patch can save bandwidth and messages up to 50% for REGISTER, MESSAGE and INVITE requests compared to traditional www_authorize authentication. This improvement is especially important for clients connected via mobile networks (higher packet loss / latency). 

Example configuration for TLS client authentication:
```
# authenticate the REGISTER requests 
if (proto==TLS && is_peer_verified()){
    xlog("L_NOTICE","[$pr:$fU@$si:$sp]: Processing '$rm' auth trusted cert '$tls_peer_subject_cn'\n");

    # Doing pretty serious stuff here, check if to matches CN.
    if (!tls_check_to())
    {
        xlog("L_NOTICE","[$pr:$fU@$si:$sp]: Processing '$rm' auth TO check failed\n");
        sl_send_reply("403","Forbidden auth ID");
        exit;
    }
}
else {
    # TLS validation could not be applied - use challenge response
    $var(auth_code) = www_authorize("", "subscriber");
    xlog("L_NOTICE","[$pr:$fU@$si:$sp]: Processing '$rm' auth: '$var(auth_code)'\n");

    if ( $var(auth_code) == -1 || $var(auth_code) == -2 ) {
        xlog("L_NOTICE","Auth error for $fU@$fd from $si cause $var(auth_code)");
    }
    if ( $var(auth_code) < 0 ) {
        www_challenge("", "0");
        exit;
    }
    if (!db_check_to())
    {
        sl_send_reply("403","Forbidden auth ID");
        exit;
    }
}
```

### DIFF
--- a/cfg.lex
+++ b/cfg.lex
@@ -357,6 +357,7 @@ TLS_CERTIFICATE	"tls_certificate"
 TLS_PRIVATE_KEY "tls_private_key"
 TLS_CA_LIST		"tls_ca_list"
 TLS_CA_DIR		"tls_ca_dir"
+TLS_CRL_DIR		"tls_crl_dir"
 TLS_DH_PARAMS		"tls_dh_params"
 TLS_EC_CURVE		"tls_ec_curve"
 TLS_CIPHERS_LIST	"tls_ciphers_list"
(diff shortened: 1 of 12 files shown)


ph4r05:
as new features are not accepted to 1.11 branch, should I do the same changes also on 2.x branch? Will it be accepted then?

bogdan-iancu:
Please migrate your PR to master (2.2) in order to be accepted. The feature looks great and I will love to have it committed, but we can do this only in 2.2/master branch

PR #715, opened 2015-12-03
ASYNC_CHANGE_FD status + db engine async state parameter + db_virtual async

ionutrazvanionita:
This pull requests adds the following:

* added a new status - ASYNC_CHANGE_FD which allows changing the file descriptor in the reactor if something goes wrong in the resume function, such that you can do failover
* added a state parameter for database async query - database engines can now use a parameter to hold state between the async call (async_raw_query) and the async resume (async_raw_resume)
* adapted sipcapture and avpops to the new functionalities
* db_virtual can now use the async raw query functionality; since mysql is the only engine that supports this, it will be the only one supported


bogdan-iancu:
@liviuchircu , please review this patch

async.h
```diff
@@ -36,7 +36,8 @@
  * NOTE: all values in this enum must be negative
  */
 enum async_ret_code {
-	ASYNC_NO_IO = -5,
+	ASYNC_NO_IO = -6,
+	ASYNC_CHANGE_FD,
```
liviuchircu: I'd suggest moving `ASYNC_CHANGE_FD` two spots below, since it's still an async operation (but on a new DB backend/host), along with `ASYNC_CONTINUE`.

modules/db_virtual/dbase.c
```diff
@@ -512,3 +512,180 @@ int db_virtual_insert_update(const db_con_t* _h, const db_key_t* _k,
 {
     db_generic_operation2(insert_update(handle->con, _k, _v, _n),1, 1, 1);
 }
+
+#define CURRCON(__ah__) (__ah__->current_con)
+
+#define db_generic_async_operation(__h,__ah, _resume_f, FUNC_WITH_ARGS)         \
```
liviuchircu: What's the deal with all the underscores in macro params? To prevent upper function variable overriding? If yes, working out the output of the following code before running it yourself will definitely be good exercise:

```
main() {
    int a = 1, b = 2;
    #define foo(a) a = 2
    foo(b);
    printf("%d %d\n", a, b);
}
```

As a side note, prefixing a C function with `__` is actually **useful** when intending to mark that function as *"a low-level function you should use with caution"*. Take for example `set_proc_log_level`, which makes use of `__set_proc_log_level` behind the curtains.

modules/db_virtual/dbase.c
```diff
@@ -512,3 +512,180 @@ int db_virtual_insert_update(const db_con_t* _h, const db_key_t* _k,
...
+                                                                                \
+			/* try next*/                                                       \
+			rc = -1;                                                            \
+			CURRCON(__ah)  = (CURRCON(__ah)+1)%p->size;                         \
+		}                                                                       \
+		LM_DBG("curent_con = %i\n", CURRCON(__ah));                             \
+	} while ((__ah)->cons_rem); /* should never exit here */                    \
+                                                                                \
+	return rc;                                                                  \
+}while (0);
```
liviuchircu: [This](https://sourcemaking.com/files/sm/images/spagett.jpg) is all I can say about the runaway macros in `db_virtual`.

modules/db_virtual/dbase.c
```diff
@@ -512,3 +512,180 @@ int db_virtual_insert_update(const db_con_t* _h, const db_key_t* _k,
...
+
+
+int db_virtual_async_raw_query(db_con_t *_h, const str *_s, void **_data)
+{
+	handle_async_t* _ah;
+    handle_con_t * _handle;
+    handle_set_t * _p = (handle_set_t*)_h->tail;
+
+	if (_s->len > 16384) {
```
liviuchircu: Please define this in some header, makes code much more **readable** and **easy to change**

modules/db_virtual/dbase.c
```diff
@@ -512,3 +512,180 @@ int db_virtual_insert_update(const db_con_t* _h, const db_key_t* _k,
...
+		LM_ERR("query exceeds buffer size(16384)!\n");
+		return -1;
+	}
+
+	if ((_ah=pkg_malloc(sizeof(handle_async_t)+_s->len)) == NULL) {
+		LM_ERR("no more pkg\n");
+		return -1;
+	} else {
+		/* for RR next time the script shall trigger this,
+		 * other dest shall be called*/
```
liviuchircu: I literally had to read the code in order to understand the comment, rather than vice-verse!

Rule of thumb: when it comes to formulating comments, two things should come to mind: **well-written** vs. **badly-written** (language, acronyms, doc templates, coding style, etc.). **insightful** vs. **impeding** (usefulness of the comment itself: do you really have to *explain how round-robin works*, or simply that this code just does a *"round-robin increment"*?)

modules/sipcapture/sipcapture.h
```diff
@@ -62,5 +62,6 @@ struct hep_ip6hdr {
         struct in6_addr hp6_dst;        /* destination address */
 };
 #endif
+typedef void* sc_async_param_t;
```
liviuchircu: Yet nobody includes `sipcapture.h` anyway!

modules/tm/async.c.orig
```diff
@@ -0,0 +1,258 @@
+/*
```
liviuchircu: Original or fake, `async.c.orig` will have to go.

Issue #721, opened 2015-12-09
SIGSEGV in pkg_malloc/fm_malloc


shimaore:
This is using opensips 2.1 at commit ee99ac71d8718c93c29e28e6a5266287491f17a5 in a production environment. OpenSIPS is configured to generate coredump, [core.txt](https://github.com/OpenSIPS/opensips/files/56763/core.txt) has `bt full` + `print` of variables in last 3 stack entries.

The issue in the code is relatively simple to understand,

    pf = n->prev;

is followed by

    *pf = n->u.nxt_free;

(lines [167](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L167) and [171](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L171) of [mem/f_malloc.c](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c)) without checking whether `n->prev` actually exists. However this is fairly old code so there must be some assumptions that made the test unnecessary at the time — but maybe not anymore, or maybe there's some environment dependency I need to check.

shimaore:
FWIW there are two locations in mem/f_malloc.c that might have problems with `n->prev` being NULL before calling `fm_remove_free`:
- The first one is [line 353](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L353) if `n` === `frag`.
- The other one is the one where this issue goes through, [line 385](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L385). The loop at [lines 332-337](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L332) never checks `frag->prev`.

Also there are only three places were `->prev` is set to NULL:
- in [`fm_remove_free`](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L178)
- in [`fm_malloc_init`](https://github.com/OpenSIPS/opensips/blob/2.1/mem/f_malloc.c#L288) for `qm->last_frag` and `qm->first_frag`.

Not sure I can help much more. :]

shimaore:
PR #722 has a workaround for the condition experienced in this case. However this might lead to fragments not being re-used iff `->prev` is already NULL, which might lead to memory leak / memory exhaustion.
This would need to be reviewed by someone more conversant with memory management. :)

razvancrainea:
[...]

PR #722, opened 2015-12-09
Prevent crash in `fm_remove_free`.

shimaore:
This is a workaround for #721 
This doesn't address the base question (why would a fragement have `->prev` set to NULL), so this isn't a fix, but it should prevent crashes for this case.


razvancrainea:
Unfortunately this issue is a bit more delicate. The problem is that ```->prev``` should _never_ be NULL,  otherwise it wouldn't have been in the ```qm->free_hash[hash]``` hash.
This is a concurrency problem we are tracing for a while, but unfortunately we couldn't find a fix for it.

PR #724, opened 2015-12-10
mem/module_info: adapt code to shared memory initialisation change

ionel-cerghit:
the statistics vector is exteded as more memory groups are declared
if SHM_SHOW_DEFAULT_GROUP is defined allocations done util grups are
defined are added to the default group

### DIFF
--- a/mem/module_info.c
+++ b/mem/module_info.c
@@ -70,75 +70,100 @@ int set_mem_idx(char* mod_name, int  mem_free_idx){
 	return 0;
 }
 
-inline void update_module_stats(long mem_used, long real_used, int frags, int group_idx){
+int init_new_stat(stat_var* stat) {
+#ifndef DBG_MALLOC
+	#define MALLOC_UNSAFE(_size)  MY_MALLOC_UNSAFE(shm_block, _size)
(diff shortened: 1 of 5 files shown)


mem/module_info.c
```diff
@@ -59,4 +60,61 @@ inline void update_module_stats(long mem_used, long real_used, int frags, int gr
...
+#ifndef SHM_SHOW_DEFAULT_GROUP
+	groups = mem_free_idx - 1;
+#else
+	groups = mem_free_idx;
+#endif
+
+	one_full_entry = 3 * (sizeof(stat_var) + sizeof(stat_val));
+	size_prealoc = groups * sizeof(struct module_info) + groups * one_full_entry;
+
+#ifndef DBG_QM_MALLOC
```
razvancrainea: I think this flag is obsolete. Please check if this flag is still present in the latest version.

razvancrainea: Any updates on this?

ionel-cerghit: It got obsolete indeed. Now seems to be working fine. Thanks for the observation!

PR #728, opened 2015-12-17
Slot Based Taildrop algorithm for ratelimit

ionutrazvanionita:
In this new algorithm a window is held per every pipe,
which is defined by two parameters: window_size(seconds) and
slot_period(microseconds). The number of slots is
window_size*1000/slot_period. The window is updated with every
call received and it's implemented as a circular vector, which
modifies it's start based on the elements that were dropped.

How is the window updated:
* if no message comes in 2*window_time from when the window start
was set , it means we can't use the values from any of the slots
and we set the value from any of the slots to 0; the window start
time is updated;
* if the message comes in t=[start+window_time; start+2*window_time)
it means that we can keep the values in slots in the interval
[t-window_time;start+window_time) and drop the ones in
[window_time; t-window_time); the window start time is updated;
* it the message come in t=[start; start+window_time) we don't
do any update do the window, only to the value of the slot corresponding
to time t;

The algorithm also shares this value through bin interface, allowing
having the values from all the replicas. Also, before sending the
value through the interface, the replica checks the window in order
to send only the values in valid slots. Two different functions
were made because of efficiency reasons. When replicating
we only do some very short loops or maybe nothing.

### DIFF
--- a/modules/ratelimit/ratelimit.c
+++ b/modules/ratelimit/ratelimit.c
@@ -68,6 +68,9 @@ int rl_repl_cluster = 0;
 int accept_repl_pipes_timeout = 10;
 int repl_pipes_auth_check = 0;
 
+int rl_window_size=10;   /* how many seconds the window shall hold*/
+int rl_slot_period=200;  /* how many milisecs a slot from the window has  */
+
 static str db_url = {0,0};
 str db_prefix = str_init("rl_pipe_");
 
@@ -140,6 +143,8 @@ static param_export_t params[] = {
 	{ "replicate_pipes_to",		INT_PARAM,	&rl_repl_cluster		},
 	{ "accept_pipes_timeout",	INT_PARAM,	&accept_repl_pipes_timeout	},
 	{ "repl_pipes_auth_check",	INT_PARAM,	&repl_pipes_auth_check		},
+	{ "window_size",            INT_PARAM,  &rl_window_size},
+	{ "slot_period",            INT_PARAM,  &rl_slot_period},
 	{ 0, 0, 0}
 };
 
@@ -346,7 +351,7 @@ static int mod_init(void)
 
 	if (repl_pipes_auth_check < 0)
 		repl_pipes_auth_check = 0;
-		
+
 	if ( (rl_repl_cluster || accept_repl_pipes) && load_clusterer_api(&clusterer_api) != 0 ){
 		LM_DBG("failed to find clusterer API - is clusterer module loaded?\n");
 		return -1;
@@ -417,8 +422,8 @@ static int mod_init(void)
 		LM_ERR("cannot init bin replication\n");
 		return -1;
 	}
-	
-	
+
+
 
 	return 0;
 }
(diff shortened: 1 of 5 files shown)


modules/ratelimit/ratelimit.c
```diff
@@ -477,6 +482,122 @@ int hash[100] = {18, 50, 51, 39, 49, 68, 8, 78, 61, 75, 53, 32, 45, 77, 31,
...
+ * the algorithm keeps a circular window of requests in a fixed size buffer
+ *
+ * @param pipe   containing the window
+ * @param update whether or not to inc call number
+ * @return number of calls in the window
+ */
+static inline int hist_check(rl_pipe_t *pipe)
+{
+	/* Window ELement*/
+	#define WEL(__el__) pipe->rwin.__el__
```
razvancrainea: This is really bad programming - use the structure and its field!

modules/ratelimit/ratelimit.c
```diff
@@ -477,6 +482,122 @@ int hash[100] = {18, 50, 51, 39, 49, 68, 8, 78, 61, 75, 53, 32, 45, 77, 31,
...
+	int first_good_index;
+	int rl_win_ms = rl_window_size * 1000;
+
+
+	unsigned long long now_total, start_total;
+
+	struct timeval tv;
+
+	/* first get values from our beloved replicated friends */
+	count = rl_get_all_hist_counters(pipe);
```
razvancrainea: You should use the ```counter``` field of the pipe.

modules/ratelimit/ratelimit_helper.c
```diff
@@ -911,6 +923,7 @@ void rl_rcv_bin(int packet_type, struct receive_info *ri, int server_id)
 		/* set the destination's counter */
 		destination = find_destination(*pipe, server_id);
 		destination->counter = counter;
+		destination->hist_counter = hist_counter;
```
razvancrainea: why not using the counter, but the hist_counter instead?

modules/ratelimit/ratelimit_helper.c
```diff
@@ -920,6 +933,59 @@ void rl_rcv_bin(int packet_type, struct receive_info *ri, int server_id)
...
 }
 
+/*
+ * same as hist_check() in ratelimit.c but this one
+ * only counts, no updates on the window ==> faster
+ */
+static inline int hist_count(rl_pipe_t *pipe)
+{
+	/* Window ELement*/
+	#define WEL(__el__) pipe->rwin.__el__
```
razvancrainea: Again, bad programming.

modules/ratelimit/ratelimit_helper.c
```diff
@@ -995,6 +1061,9 @@ void rl_timer_repl(utime_t ticks, void *param)
 			if (bin_push_int((*pipe)->limit) < 0)
 				goto error;
 
+			if (bin_push_int(hist_count(*pipe)) < 0)
```
razvancrainea: Just compute the last counter and push in the counter variable.

Issue #742, opened 2016-01-05
failing to loose route sequential requests that are proto wss


etamme:
On master commit 968a4e004ebf85687f864b885a1ad5dc6632f41a

If I place a call from a WSS client to a UDP client, then the UDP client hangs up, opensips will show the following log indicating that it cant identify the protocol.
```
Jan  5 12:35:23 voipphreaker /usr/local/opensips/sbin/opensips[9885]: DBG:rr:after_loose: Topmost route URI: 'sip:159.203.234.107;r2=on;lr;ftag=s6992f0ms9;did=e6a.94fb3f15' is me
Jan  5 12:35:23 voipphreaker /usr/local/opensips/sbin/opensips[9885]: WARNING:core:grep_sock_info: unknown proto 9
Jan  5 12:35:23 voipphreaker /usr/local/opensips/sbin/opensips[9885]: WARNING:rr:after_loose: no socket found to match 2nd RR [9][159.203.234.107:443]
```

I dug around in sock_info.c and net/trans.c to find how the protos list was populated, but I'm not able to immediately see whats wrong.  get_sock_info_list appears to be returning 0, but it looks like the wss proto is in the list as proto 9 as the log indicates, so I'm not clear on whats going on.

etamme:
This has been tracked down to parse_uri.c not parsing wss - looking into how to fix it now

etamme:
im close on a patch that fixes this, but im having trouble because WS and WSS are the same letters up till the last S.  Im not sure how to avoid the transport_fin call for WS as i basically need to look ahead.

here is my current diff https://gist.github.com/etamme/953cb286cb73b6a62216
It you uncomment 

```
//transport_fin(VW_S_FIN, PROTO_WS);
```

It will fail to compile, but i tested it as is with WSS and got clean loose routing on a sequential request towards a WSS client.

bogdan-iancu:
Thanks to @jarrodb , this is fixed via #744

PR #744, opened 2016-01-05
add parse_uri support for transport=wss

jarrodb:
Try this patch for #742


bogdan-iancu:
Thank you @jarrodb , work in progress to integrate (some changes are still required)

PR #746, opened 2016-01-06
rtpproxy: patch: has_sdp() does not exist

lemenkov:
Not sure how this slipped through, but we don't have a function called has_sdp.


lemenkov:
Just a note - this is just a docfix, not a code change.

bogdan-iancu:
Thank you @lemenkov !!

PR #748, opened 2016-01-08
dispatcher: add ds_list full mi command, update documentation

ionel-cerghit:
optional parameter 'full' added for ds_list full, which will display
the weight, priority and description of a destination
role of weight and priority briefly explained in documentation, FAQ section

### DIFF
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -148,7 +148,7 @@ void ds_destroy_data(ds_partition_t *partition)
 
 
 int add_dest2list(int id, str uri, struct socket_info *sock, int state,
-							int weight, int prio, str attrs, ds_data_t *d_data)
+							int weight, int prio, str attrs, str description, ds_data_t *d_data)
 {
 	ds_dest_p dp = NULL;
 	ds_set_p  sp = NULL;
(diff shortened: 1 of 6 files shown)


modules/dispatcher/dispatch.c
```diff
@@ -2110,6 +2123,27 @@ int ds_print_mi_list(struct mi_node* rpl, ds_partition_t *partition)
 				if(node1 == NULL)
 					goto error;
 			}
+
+			if (flags & 1) {
```
bogdan-iancu: Please define the flag "1" as a macro with the meaning of "full listing"

modules/dispatcher/dispatcher.c
```diff
@@ -1365,7 +1377,7 @@ static struct mi_root* ds_mi_list(struct mi_root* cmd_tree, void* param)
 				9, part_it->name.s, part_it->name.len);
 
 		if (part_node == NULL
-			|| ds_print_mi_list(part_node, part_it) < 0) {
+			|| ds_print_mi_list(part_node, part_it, explicit) < 0) {
```
bogdan-iancu: as third param of ds_print_mi_list() is a bitmask, use here the predefined flags as macro (see the other comment). Maybe "explicit" is not the best name - better use a "flag" variable to collect through the code all the flags you want to pass to ds_print_mi_list().

ionel-cerghit:
Thank you for the comments Bogdan! I made the changes.

PR #752, opened 2016-01-12
[siptrace] all types of tracing are done using sip_trace function

ionutrazvanionita:
* added trace_id parameter with which one is able to define
database urls and trace uris (HEP and SIP); the trace id is given
as an argument to sip_trace() function;
* multiple trace_id s can be defined under the same name which
will alow users to do multiple types of tracing using only one
function call;
* sip_trace() now accepts flags to specify what to trace: message,
transaction or dialog; in stateless mode only message trace does
a valid type of trace;
* traced_user_avp now is as an argument to sip_trace() function;
it's name now changed to trace_attrs and also the column in the
database and table version were modified;setting it won't duplicate
entries in the database; it will do only one insert setting the
trace_attrs column accordingly;
* most of the parameters in the modparam section(db_url, table,
duplicate_uri, duplicate_with_hep) are now part of a trace_id
called 'default'; removed the old ones completely; all that remains
now is the trace_on parameter which can be set/unset with mi
(read the docs for more info), trace_local_ip and trace id;
* the stateless part of the sip_trace module was completely
removed; this also caused that the ACKs that are forwarded
by OpenSIPS are caught only when they come(input) and not when
they leave(output); this was done using the stateless forward
callback but this is highly ineffective(registering a callback
that would be called for all the requests for only one type of
messages);

### DIFF
--- a/db/schema/sip_trace.xml
+++ b/db/schema/sip_trace.xml
@@ -9,7 +9,7 @@
 
 <table id="sip_trace" xmlns:db="http://docbook.org/ns/docbook">
     <name>sip_trace</name>
-    <version>4</version>
+    <version>5</version>
     <type db="mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
         <db:para>This table is used to store incoming/outgoing SIP messages in database. More informations can be found in the siptrace module documentation at: &OPENSIPS_MOD_DOC;siptrace.html.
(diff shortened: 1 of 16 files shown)


bogdan-iancu:
Good work @ionutrazvanionita  !! Really great stuff this enhancement in siptrace.

PR #781, opened 2016-02-02
[dialog] flag for replicating profiles

ionutrazvanionita:
added a new flag which allows replicating over the bin interface
only the profiles which have it set

### DIFF
--- a/modules/dialog/dlg_db_handler.c
+++ b/modules/dialog/dlg_db_handler.c
@@ -388,7 +388,7 @@ void read_dialog_profiles(char *b, int l, struct dlg_cell *dlg,int double_check,
 	char *end;
 	char *p,*s,*e;
 	char bk;
-	int use_cached;
+	unsigned repl_type;
 
 	end = b + l;
 	p = b;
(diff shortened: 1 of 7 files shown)


modules/dialog/dlg_db_handler.c
```diff
@@ -1176,8 +1178,8 @@ str* write_dialog_profiles( struct dlg_profile_link *links)
 		for( i=0 ; i<link->value.len ; i++ )
 			if (link->value.s[i]=='|' || link->value.s[i]=='#'
 					|| link->value.s[i]=='\\') l++;
-		if (link->profile->use_cached)
-			l+=cached_marker.len;
+		if (link->profile->repl_type/*==(CACHEDB||PROTOBIN)*/)
```
razvancrainea: Better check if the replication type is different than REPL_NONE.

modules/dialog/dlg_profile.h
```diff
@@ -54,10 +54,11 @@ struct dlg_profile_link {
 
 struct repl_prof_novalue;
 
+enum repl_types {REPL_NONE=0, CACHEDB=1, PROTOBIN};
```
razvancrainea: I think prefixing all with REPL_ sounds better. Like `{REPL_NONE, REPL_CACHEDB, REPL_PROTOBIN};`

PR #793, opened 2016-02-15
add ds_probing_list parameter to specify which sets to probe

jarrodb:
Add support for a parameter that specifies which destinations this proxy is responsible for probing:

```
modparam("dispatcher", "ds_probing_list", "1")
```

I have a replicated database that includes all of my destination sets which multiple proxies load, but I want to be able to specify which sets a specific proxy is responsible for probing.

If this is something that is eventually accepted, I will update with the necessary documentation.

### DIFF
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -67,6 +67,7 @@
 #include "../../rw_locking.h"
 
 #include "dispatch.h"
+#include "ds_fixups.h"
 #include "ds_bl.h"
 
 #define DS_TABLE_VERSION	7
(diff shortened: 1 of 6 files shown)


bogdan-iancu:
Hi @jarrodb , is this backward compatible ? like not setting this new param at all will make dispatcher probe all the destination (with probing active, of course) ?

jarrodb:
Absolutely it is backwards compatible, namely by this check within the timer function:

```
if ( (!ds_probing_list || in_int_list(ds_probing_list, list->id)==0) &&
```

The condition passes for all destination sif ds_probing_list is NULL.

bogdan-iancu:
I see no reason not to accept this PR - could please attach to it the documentation part, so we can upload it all from one move ?

jarrodb:
I believe I have added the necessary lines for defining the additional parameter in the documentation.

bogdan-iancu:
Thank you @jarrodb  !!

Issue #696, opened 2015-11-12
Wrong rule selected in dialplan translation when both string and regexp rules are matched


paolodepa:
Let's say we are translating a destination matching two dialplan rules, the first working on exact string match and the second working on regexp match: I'm getting selected the exact match rule ONLY IF its priority is bigger (>) than the regexp rule; the regexp rule is selected otherwise.
I would have expected the opposite behaviour.

I'm facing this issue on 1.11.5 version but, looking at  the source code, the problem should be present also in 2.1.X.
In detail, I suppose that replacing this code snippet into the function "translate" (dp_repl.c):

        /* pick the rule with lowest table index if both match and prio are equal */
        if ((string_res | regexp_res) == 0) {
                if (rulep->pr < rrulep->pr) {
                        rulep = rrulep;

with:

        /* pick the rule with lowest table index if both match and prio are equal */
        if ((string_res | regexp_res) == 0) {
                if (rrulep->pr < rulep->pr) {
                        rulep = rrulep;

would make the job.

bogdan-iancu:
@paolodepa, if there is no clear ordering based on priorities, there is no "expected" behavior . To avoid any overlapping or incertitude, use priorities

paolodepa:
@bogdan-iancu that's exactly the reason why I submited this issue: the "expected" behaviour should be to select the lowest priority rule, regardless of the type of matching (regexp vs exact), isn't it?

Instead "I'm getting selected the exact match rule ONLY IF its priority is bigger (>) than the regexp rule".

BTW, working side by side with @Parantido, I know that his reworking in 2.1.2 and subsequent versions fixes this issue.

bogdan-iancu:
@paolodepa, that;s my point - there is not expected behavior in such case. And we strongly advice to solve such conflicts with priorities and not with "expected" behavior.
I do not remember to have in docs any statements of a such expected behavior - if I'm missing something here, please correct me.
Yes, I'm aware of the patches from @Parantido and we have scheduled them to be reviewed for 2.3 release.

paolodepa:
Neither I could find any specification in the docs on the priority field in dialplan module, unless of arbitrarily supposing it to be similar to the same field in cc_flows table, which is ascending (lower value means higher priority).

You are telling me instead that the priority order has to be considered descending into the dialplan module (at least in this piece of code), right?

liviuchircu:
[...]

PR #794, opened 2016-02-15
MOD DIALPLAN v2.1.2: New Dialplan Features

Parantido:
- Rules preload/translate alghoritm reworked: no more needs to have different buckets for REGEXP and BEST MATCH ones. All rules belong from same bucket. This way opens to many different scenarios and future development without afflicting performances (this rework will also fix issue #696 https://github.com/OpenSIPS/opensips/issues/696).
- New "match_var" database column introduced. This column will override "src/dest" dp_translate parameter when filled. It respects same syntax (wrong syntax will cause rule to being discarded at preload time).
- New "matched_pvar" AVP introduced. It will be filled with last matched rule id. AVP name can be set by modparam directive:

		modparam("dialplan", "matched_pvar", "$avp(dp_ruleid)")

- New "continue_search" database column introduced. When filled to "1", rule matching will continue through other rules. ATTRS and MATCHED_PVAR will be updated as soon as rule matches (next rule match will works on PVAR updated values). After a dp_translate ATTRS and MATCHED_PVAR will be updated with last rule match values.

All new features and code rework default behaviour respects old syntaxes. No database entries or opensips.cfg scripts needs to be changed on module
upgrade

### DIFF
--- a/modules/dialplan/dialplan.c
+++ b/modules/dialplan/dialplan.c
@@ -22,7 +22,6 @@
  *  2007-08-01 initial version (ancuta onofrei)
  */
 
-
 #include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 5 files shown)


Parantido:
Sorry I totally missed database schema

CREATE TABLE IF NOT EXISTS `dialplan` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `dpid` int(11) unsigned DEFAULT NULL,
  `pr` int(11) NOT NULL,
  `match_op` int(11) NOT NULL,
  `match_exp` char(64) NOT NULL,
  `match_flags` int(11) DEFAULT NULL,
  `match_len` int(11) NOT NULL,
  `subst_exp` char(64) NOT NULL,
  `repl_exp` char(32) NOT NULL,
  `disabled` int(11) NOT NULL DEFAULT '0',
  `attrs` char(32) DEFAULT NULL,
  `timerec` varchar(255) NOT NULL DEFAULT '',
  `match_var` char(32) DEFAULT NULL,
  `continue_search` tinyint(1) NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`),
);

Parantido:
Hi Bogdan,

in which opensips version this feature will be introduced?

Thank you

PR #822, opened 2016-03-07
free disposition pkg_mem when parse_disposition returns parse_errors (-2)

jarrodb:
free disposition pkg_mem when parse_disposition returns parse_errors (-2)


bogdan-iancu:
@jarrodb , thanks for spotting this mem leak. Still, the fix was not correct (on -1 it was doing double free). Also, the correct meaning of the return codes is : -1 is error with no mem allocated ; -2 is error during parsing with mem allocated.

PR #823, opened 2016-03-08
Add JSONOID to pg_type.h for supporting JSON return types as DB_STRING

jarrodb:
I receive the following error when my PostgreSQL stored procedures return JSON type or I use row_to_json in an avp_db_query:

```WARNING:db_postgres:db_postgres_get_columns: unhandled data type column (server_detail) type id (114), use DB_STRING as default```

This just defines JSONOID as DB_STRING in `db_postgres_get_columns` as opposed to letting it default to DB_STRING and generate the warnings.


bogdan-iancu:
Thank you @jarrodb !

PR #825, opened 2016-03-10
free_disposition for any negative return error code

jarrodb:
I believe #822 is still occurring and that the disposition memory should be freed when ANY negative error code is received.  The package memory was still allocated just a few lines prior and a return code of -1 from `parse_disposition` only means that a disposition parameter was unable to allocate pkg memory.


bogdan-iancu:
@jarrodb, many thanks for keep looking into this issue. After paying more attention into this, the massive leak comes from not freeing the "disp" structure (along with all its parameters) on the success branch. Even if the Disposition body is ok, the "disp" is neither attached, not freed anywhere else in the code. So, I extended your fix to free the disp all the time, on failure and success.

PR #828, opened 2016-03-15
Avoid segfault and return NULL value for $si if empty

rrb3942:
After upgrading some systems from 1.10 to 1.11 I started receiving segfaults that I traced back to accessing $si when it was not populated in the msg and a failure to correctly check the return value of ip_addr2a(). I believe it only happened on internally generated requests.

Not sure what changed between 1.10 and 1.11 to trigger this, but this patch fixes the crash and seems to match 1.10 behavior.

### DIFF
--- a/pvar.c
+++ b/pvar.c
@@ -1046,7 +1046,17 @@ static int pv_get_srcip(struct sip_msg *msg, pv_param_t *param,
 	if(msg==NULL)
 		return -1;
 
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @rrb3942 for your fix - I just pushed a larger one (covering several aspects) . Please confirm that the  commit 2d0cebc really fixes the issues for you.

Regards,
Bogdan

PR #832, opened 2016-03-18
Wildcard operator for partitions in some drouting functions

ionutrazvanionita:
do_routing() and is_from_gw() functions now support wildcard(*) operator for partitions


bogdan-iancu:
Thank you John !!

PR #836, opened 2016-03-22
Make s.int also operate on floats/doubles by truncating them to an int

rrb3942:
This actually makes two changes:

1. It makes it so that when {s.int} is performed on an empty string it returns 0. This was the old/original behavior that was changed by https://github.com/OpenSIPS/opensips/commit/4fdb55bfab9b7cfcc3dfbf15093e2becafe7ec4b

Not sure if it was intended to affect the transformation or not. While I think the change is good for the primary use of that function, for the script level transformation we should be more lax.

2. {s.int} now also accepts a double/float and returns an int. It always truncates the number (e.g. 2.9 is 2, 2.3 is 2, -2.9 is -2, -2.3 is -2), which I think is a sane default. If someone needs more control over the rounding they can use the mathops module.

See issue #788 for some a bit of discussion on making {s.int} more permissive.

Not sure where to update the documentation for the transformations. If someone can point me at the right spot I am glad to update the documentation as well to reflect these changes.

Thanks.

### DIFF
--- a/transformations.c
+++ b/transformations.c
@@ -29,6 +29,8 @@
 #include <time.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <errno.h>
+#include <math.h>
 
 #include "dprint.h"
 #include "mem/mem.h"
@@ -162,6 +164,7 @@ int tr_eval_string(struct sip_msg *msg, tr_param_t *tp, int subtype,
 		pv_value_t *val)
 {
 	int i, j;
+	double d;
 	char *p, *s;
 	str st;
 	pv_value_t v;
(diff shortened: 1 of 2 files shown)


transformations.c
```diff
@@ -182,8 +185,25 @@ int tr_eval_string(struct sip_msg *msg, tr_param_t *tp, int subtype,
...
+					d = strtod(val->rs.s, NULL);
+
+					if(errno != 0) {
+						return -1;
+					}
+
+					//Use truncation for getting int value. Ensures consistent treatment
+					//trunc(1.5) is 1.0 and trunc(-1.5) is -1.0
+					//mathops gives users more rounding options if they require them
+					val->ri = trunc(d);
```
liviuchircu: The only benefit to adding additional error handling by the use of `strtod` and `trunc` is to parse the following values, which `str2sint` is not able to:

* signed numbers, prefixed by `+` or `-`
* `INF` or `INFINITY`
* `NAN` (not a number)

Apart from the above negligible benefits, `str2sint` will successfully parse and truncate all input that's fed to it. Coupled with the fact that we now have to link against `lm`, I think we should rather:

* update the function comments on `str2sint`, so that the truncation / partial parse (just like `strtod` and friends) is obvious to all devs
* avoid further error handling for `str2sint` and just use its result, after it parsed all it could from the given input

liviuchircu:
First of all, thanks for the follow-up!

Some quick feedback:

* I completely agree that parsing a NULL / empty string should output 0, and not throw some weird errors (standard `strtol` behaviour). Furthermore, "errors" (or 0 values) are still detectable at script level, should the writer decide to catch them.
* Added a suggestion on improving the patch
* The manual, along with Transformations wiki page, of OpenSIPS 2.2 can be found [here](http://www.opensips.org/Documentation/Manual-2-2). All pages are globally editable - you just need to register an account.

rrb3942:
My only question would be how should completely bogus input be handled (say the letter 'Z', with no numbers). It will return a -1 and set the int to 0, which would be the same result as the input '0.1231'. Should they be differentiated at script level?

liviuchircu:
Well, with your initial patch, it would return `0`, while hiding the error. I consider this to be the correct approach from a scripting perspective, especially since transformation errors are quite cumbersome.

Input errors can be caught beforehand, if needed, through the use of the regex operator, `=~`.

rrb3942:
Replacing this pull request with #842

PR #841, opened 2016-03-25
drop_accounting function for acc

ionutrazvanionita:
This function allows unsetting parameters set by do accounting
function. It has the same parameters excepting the table name.

### DIFF
--- a/modules/acc/doc/acc_admin.xml
+++ b/modules/acc/doc/acc_admin.xml
@@ -1058,10 +1058,10 @@ modparam("acc", "acc_created_avp_name", "call_created_avp")
 
 	<section id="cancel_accouting">
 		<title>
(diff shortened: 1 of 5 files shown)


liviuchircu:
I would suggest renaming it to `drop_accounting` just to avoid any collisions/confusions related to the SIP CANCEL method. To us it may seem obvious, to others it may not - less ambiguity is always better.

modules/acc/doc/acc_admin.xml
```diff
@@ -1057,6 +1056,96 @@ modparam("acc", "acc_created_avp_name", "call_created_avp")
 
 	</section>
 
+	<section id="cancel_accouting">
```
liviuchircu: You left out this little thingie!

PR #869, opened 2016-04-27
bugfix in rest_client module: avoid waiting too long for libcurl connection

qmphan:
The error is in the loop:

long retry_time, check_time = 5;
for (i = 0; i < busy_wait; i += check_time) {
    ....
    sleep(1000 * check_time)
}

We sleep 5 **miliseconds** but increase only 5 **microsecond** in each loop.

### DIFF
--- a/modules/rest_client/rest_methods.c
+++ b/modules/rest_client/rest_methods.c
@@ -127,7 +127,7 @@ int start_async_http_req(struct sip_msg *msg, enum rest_client_method method,
 	fd_set rset, wset, eset;
 	int max_fd, fd, i;
 	long busy_wait, timeout;
-	long retry_time, check_time = 5; /* 5ms looping time */
+	long retry_time, check_time = 5000; /* 5000 micro seconds = 5ms looping time */
 	int msgs_in_queue;
 	CURLMsg *cmsg;
 
@@ -259,7 +259,7 @@ int start_async_http_req(struct sip_msg *msg, enum rest_client_method method,
 				}
 			}
 
-			usleep(1000UL * check_time);
+			usleep(check_time);
 		}
 	}
 


liviuchircu:
Did this patch actually fix anything for you? Because it looks like it's only going to cause that polling loop to run only once (since `libcurl` usually doesn't come back with a wait time larger than `5000`), which is not ok - the idea behind that code is that `curl_multi_fdset()` will really return a `max_fd` value of `-1` quite often the first couple of times we call it.

Also, your last statement isn't true. We sleep 5 **millliseconds**, and we increase looping by 5 **milliseconds** (both `busy_wait` and `check_time` contain ms values), which is totally ok.

PS: Thanks for reviewing the module! :-)

qmphan:
> Also, your last statement isn't true. We sleep 5 milliseconds, and we also increase looping by 5 milliseconds (both busy_wait and check_time contain ms values), which is totally ok.
No, busy_wait and check_time contain microsecond value. Please double check.

(I did this fix because my opensips wait 10 seconds before actually sending out the http request)

qmphan:
From libcurl's documentation, it it true that busy_wait and check_time should be in miliseconds! However, when I debug on Centos7(up to date), curl_multi_timeout set the timeout value to 20000, which made me believe that it is in microsecond.

It seems to be a bug in libcurl, not in your module. Let me check more and get back to you.

qmphan:
In fact, Centos7 is still shipped today with a buggy version of libcurl: https://bugs.centos.org/view.php?id=8076
Upgrading libcurl manually fixed the problem. Sorry for blaming it on your code.

By the way, I love your rest_client module and the way async function work in OpenSIP 2. Keep on great work!

PR #874, opened 2016-05-03
Updated opensipsctl.base to also search /sbin

renicok:
The md5 utility is located at /sbin/md5 on FreeBSD 8.3. An additional common location was added to allow the script to accommodate this.


liviuchircu:
Thank you for this subtle contribution, @renicok !

Issue #568, opened 2015-07-06
xlog - add option to prefix log messages with log level


ar45:
Just like the core log messages it would be nice to have a option to have xlog statements prefix the Log Level to the logs.

nikbyte:
Why not to do it by configuring syslog?

liviuchircu:
Not really sure if I understood right, but if you're asking for this:

    xlog("L_INFO", "Hi, this is INFO level debug\n");

then it has been in OpenSIPS since forever.

ar45:
@liviuchircu the log level now only controls at which level it should be emitted.  After the message is logged there is no way to know what level that came from.  Having the log level in the message helps you analyze the logs more easily.

liviuchircu:
Very useful. I would also like to suggest adding the "current_route" to the log line.

ar45:
Then it would make more sense to have a global param to set the common template format for all xlog messages, something like

```
xlog_format = "[%L_LEVEL - %ROUTE_TYPE(%ROUTE_NAME) Line:%LINE_NO] %MSG"

xlog("L_ERR", "Failed to route to PSTN attempt($avp(relay_count))\n");
```

resulting in a syslog message like
```
Jul  7 05:49:52 hostname opensips[31929]: [ERR - failure_route(HANDLE_FAILURE) Line:954] Failed to route to PSTN attempt(3)
```

ar45:
related to #663

ionel-cerghit:
Is https://github.com/OpenSIPS/opensips/pull/915 ok for the basic feature?

PR #915, opened 2016-06-23
xlog.c: add $xlog_level pseudovariable

ionel-cerghit:
using the $xlog_level pseudovariable in the xlog function will insert
in the message generated the log level, example:
xlog("$xlog_level some text\n") will generate the line
ERROR: some text


bogdan-iancu:
This PR has been merged into master, see a7f26e.
Closes #568

PR #920, opened 2016-06-30
Changed from VAL_STR to VAL_STRING in case DB_STRING

Danfx:
it's a suggestion to standard

### DIFF
--- a/modules/db_sqlite/row.c
+++ b/modules/db_sqlite/row.c
@@ -40,7 +40,7 @@
  */
 int db_sqlite_convert_row(const db_con_t* _h, db_res_t* _res, db_row_t* _r)
 {
-	int col;
+	int col,len;
 	db_val_t* _v;
 	const char* db_value;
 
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Hi @Danfx ! the existing code returns STR, but the value can be used as STRING or STR in the same time (on a similar model for DB_TEXT driver). With your change the type and the value are STRING only.
By "standard" you mean the fact that most modules (using a DB backend) expect STRING and not STR ?

Danfx:
No, I'm analyzing only this case ('case DB_STRING:') from switch. If you are using a VAL_TYPE(_v) equal DB_STRING, I guess it would be more elegant to use a structure of the DB_STRING ('VAL_STRING(_v)')
like as in other software points:
[https://github.com/OpenSIPS/opensips/blob/master/modules/db_sqlite/val.c#L95](https://github.com/OpenSIPS/opensips/blob/master/modules/db_sqlite/val.c#L95)
[https://github.com/OpenSIPS/opensips/blob/master/db/db_row.c#L68](https://github.com/OpenSIPS/opensips/blob/master/db/db_row.c#L68)
[https://github.com/OpenSIPS/opensips/blob/master/modules/db_oracle/res.c#L343](https://github.com/OpenSIPS/opensips/blob/master/modules/db_oracle/res.c#L343)

ionutrazvanionita:
merged manually

jarrodb:
I'm pretty sure you just broke clusterer which is one of the reasons it was changed: https://github.com/OpenSIPS/opensips/pull/844

Forgive me if I'm wrong...

PR #937, opened 2016-07-26
fix: pthread_mutex should be process-shared

was4444:


bogdan-iancu:
Hi @was4444 , according to the man page (http://linux.die.net/man/3/pthread_mutexattr_init), it definitely looks like you are right in this.
Do you have any idea if this was a recent changed (having a default "private" for the process-shared attribute) ?

was4444:
@bogdan-iancu, I just found this when I try to use the pthread lock option(see #859 ), so no more idea than a google.
But I don't think this was a recent changed, since the manuals were written years ago(http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutexattr_getpshared.html).

bogdan-iancu:
Thank you @was4444 !

PR #941, opened 2016-08-03
tls_mgm: load actual files from db not paths to files

ionel-cerghit:
the certificate, private_key, ca_list and dh_params columns from database
now need to contain the actual files to be used by tls, path to files are no
longer supported for this columns
domains can be set either by database or by script but not both at the same time
to avoid errors
default domains can be now loaded from database (their type needs to be different from
client and server type)

### DIFF
--- a/modules/tls_mgm/tls_domain.c
+++ b/modules/tls_mgm/tls_domain.c
@@ -101,19 +101,15 @@ void tls_release_domain(struct tls_domain* dom)
 	lock_stop_write(dom_lock);
 }
 
-int set_all_domain_attr(struct tls_domain **dom, char **str_vals, int *int_vals)
+int set_all_domain_attr(struct tls_domain **dom, char **str_vals, int *int_vals, str* blob_vals)
 {
 	size_t len;
 	char *p;
 	struct tls_domain *d = *dom;
 	size_t cadir_len = strlen(str_vals[STR_VALS_CADIR_COL]);
-	size_t calist_len = strlen(str_vals[STR_VALS_CALIST_COL]);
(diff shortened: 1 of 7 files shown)


modules/tls_mgm/tls_domain.c
```diff
@@ -221,11 +219,138 @@ int set_all_domain_attr(struct tls_domain **dom, char **str_vals, int *int_vals)
...
+		d->pkey.s = p;
+		d->pkey.len = blob_vals[BLOB_VALS_PK_COL].len;
+		memcpy(p, blob_vals[BLOB_VALS_PK_COL].s, blob_vals[BLOB_VALS_PK_COL].len);
+		p = p + d->pkey.len;
+	}
+
+	return 0;
+}
+
+int set_all_default_domain_attr(struct tls_domain *d, char **str_vals, int *int_vals, str* blob_vals)
```
razvancrainea: It seems to be a lot of duplicate code with the previous version. Could you say why?

scripts/mysql/tls_mgm-create.sql
```diff
@@ -6,14 +6,14 @@ CREATE TABLE tls_mgm (
     method CHAR(16),
     verify_cert INT(1),
     require_cert INT(1),
-    certificate CHAR(255),
-    private_key CHAR(255),
+    certificate BLOB,
```
razvancrainea: Did you manually change the values here? You should update the database [schema](https://github.com/OpenSIPS/opensips/blob/master/db/schema/tls_mgm.xml) and re-generate it for all backends (`make dbschema`).

bogdan-iancu:
Manually merged in 5009fc2e7d8caac676725ad1b2327cc04d4cdddb

PR #942, opened 2016-08-04
rtpengine recording

etamme:
Update rtpengine module with new flags and documentation to expose call recording support that has been recently added to rtpengine: https://github.com/sipwise/rtpengine/pull/245

### DIFF
--- a/modules/rtpengine/README
+++ b/modules/rtpengine/README
@@ -36,6 +36,12 @@ Richard Fuchs
 
    Sipwise GmbH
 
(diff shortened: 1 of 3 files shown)


etamme:
Hmm actually i test build on the wrong branch - let me see if i cant get this cleaned up.

bogdan-iancu:
Hi @etamme , is this PR to be update ? Or, is it still of an interest ?
Thanks, Bogdan

Issue #938, opened 2016-07-29
Dialogs are not replicated when using clusterer/TCP binary interface


sysreq0:
Using the new TCP binary interface and the Clusterer module in 2.2.x between two servers sharing a floating virtual IP address; dialogs do not seem to be replicated from one server to another. When server 1 processes a call, a 'fifo dlg_list' issued on server 2 will not show anything.

The two servers "see" each other and are active from clusterer's point of view. They are also connected to each other through the TCP binary interface. A packet capture on the binary interface's port shows traffic being sent back and forth whenever a call comes in, and also when there are no active calls.

Here's a snippet of my configuration file. The only things that change from server 1 to server 2's configuration file is the "listen =" statement and "server_id" clusterer parameter.

```
listen = bin:XXX.XXX.XXX.XXX:XXXX

#### BIN protocol
loadmodule "proto_bin.so"

#### DIALOG module
loadmodule "dialog.so"
modparam("dialog", "dlg_match_mode", 1)
modparam("dialog", "default_timeout", 21600)  # 6 hours timeout
modparam("dialog", "accept_replicated_dialogs", 1)
modparam("dialog", "replicate_dialogs_to", 1)
modparam("dialog", "accept_replicated_profiles", 1)
modparam("dialog", "replicate_profiles_to", 1)
modparam("dialog", "accept_replicated_profile_timeout", 30)
modparam("dialog", "auth_check", 1)
modparam("dialog", "profiles_no_value", "inbound/b; outbound/b")
modparam("dialog", "profiles_with_value", "carrier/b")

#### CLUSTERER module
loadmodule "clusterer.so"
modparam("clusterer", "db_url", "mysql://XXXXXXXXXX:XXXXXXXXXX@localhost/opensips")
modparam("clusterer", "server_id", 2)

[...]
        set_dlg_profile("inbound/b");
        set_dlg_profile("carrier/b", "XXXXXXXXXX");
[...]
        set_dlg_profile("outbound/b");
        set_dlg_profile("carrier/b", "XXXXXXXXXX");
```

After having enabled debugging mode, here is the kind of chatter I see between the two servers on the binary interface:

```
Jul 29 11:40:01 [22521] DBG:core:call_callbacks: binary Packet CMD: 4. Module: dialog
Jul 29 11:40:01 [22521] DBG:clusterer:bin_receive_packets: received bin packet from source: XXX.XXX.XXX.XXX:39248
[...]

PR #955, opened 2016-09-06
Clusterer rework final

rvlad-patrascu:


46labs:
@liviuchircu Are any of these going to be cherry picked into 2.2?

razvancrainea:
@46labs no, this is a complete rework of the custerer module and will only be available starting from the next release.

bin_interface.c
```diff
@@ -314,14 +409,26 @@ int bin_register_cb(char *mod_name, void (*cb)(int, struct receive_info *, void
...
 
-void call_callbacks(char* buffer, struct receive_info *rcv){
+void call_callbacks(char* buffer, struct receive_info *rcv)
+{
 	str name;
 	struct packet_cb_list *p;
+	unsigned short pkg_len;
+
+	if (!rcv_buf) {
+		rcv_buf = pkg_malloc(BUF_SIZE);
```
razvancrainea: Why allocate a pkg buffer here and not having a static buffer?

bin_interface.c
```diff
@@ -143,6 +142,41 @@ int bin_push_int(int info)
...
 }
 
+/*
+ * removes @count intergers from the end of the send buffer
+ *
+ * @return:
+ *		0: success
+ *		< 0: error, no more integers in buffer
+ */
+int bin_alter_pop_int(int count) {
```
razvancrainea: could you please rename this function? Looking at the comment and at the name of the function, I don't really understand what it is supposed to do. And since it is in the bin interface API/headers, should be pretty clear what the function should do.

modules/clusterer/clusterer.c
```diff
@@ -21,1570 +21,1628 @@
...
+	int check_call_cbs_event[MAX_NO_CLUSTERS] = {0};
+	int no_clusters = 0;
 
-str description_col = str_init("description");
-static db_key_t *clusterer_cluster_id_key;
-static db_val_t *clusterer_cluster_id_value;
+	gettimeofday(&now, NULL);
 
-static db_op_t op_eq = OP_EQ;
+	lock_get(ref_lock);
```
razvancrainea: I don't see where this lock is released - if the function is supposed to have the lock taken after it ends - that should be documented in the function's header.

razvancrainea: Also, why is it named a `ref_lock`? Does it lock for references?

modules/dialog/dlg_replication.c
```diff
@@ -1048,24 +1040,25 @@ static void repl_prof_utimer_f(utime_t ticks, void *param)
 {
 #define REPL_PROF_TRYSEND() \
 	do { \
+		nr++; \
 		if (ret > repl_prof_buffer_th) { \
 			/* send the buffer */ \
-			dlg_replicate_profiles(); \
-			replicated = 1; \
+			if (nr) \
```
razvancrainea: Note that this has been changed in the latest version, to send a packet even if there are no profiles to be replicated. This has been changed to make sure that there is communication between nodes, event if there is no traffic. If this has been fixed on the clusterer module, using heartbeats, then ignore this comment.

sysreq0:
Is this rewrite going to fix https://github.com/OpenSIPS/opensips/issues/938?

bogdan-iancu:
@sysreq0, no, this PR is not related to #938 - this PR is about enhancing the clusterer module in order to create an cluster overlay (as nodes) completely independent from the IP network topology. 
Like the overlay may say node A is sending a message to node B, but on IP level, the message may go A->C->B (as there may be no direct route from A to B).

rvlad-patrascu:
@sysreq0 actually it does fix #938 but only starting from 2.3 because, the fix being a result of the greater rework, there is no commit that specifically addresses this issue that can be directly backported to 2.2. It will have to be done separately for that version.

PR #962, opened 2016-10-05
exec: Flush the pipe after writing the message

tavyc:
I found that when using exec_msg (on 1.11) sometimes the executed script doesn't receive anything on its standard input.
Flushing the pipe after writing the message fixed the issue.

Please backport to the stable branches.


bogdan-iancu:
Many thanks @tavyc for this patch !

bogdan-iancu:
Backported to 2.2 and 2.1 too .

PR #965, opened 2016-10-17
new module blackwhite

dronord:


dronord:
What you think about it?

bogdan-iancu:
Hi @dronord, please provide a brief description about the scope of this new module, so we can evaluate it.

dronord:
The module is designed to filter packets based on black and white lists.
   For users specify black or white, or both the list of IP addresses/networks.
   Black is access, which is prohibited. White - allowed.
   If we have white addresses, then black unneeded.
   If user is not present - pass packet.

   In DB table lists stored in records:
   username - ip+mask(net) - black/white flag
   1 - black, 0 - white.
   Table readed in memory on start and can be refreshed by MI bw_reload.

   Function blackwhite() do search message parameters in memory structure.
   The time of execution on core2duo 3GHz - 12 microseconds.
   Practically not depend on number of entries.

bogdan-iancu:
Thank you @dronord. Is there any major difference from the [permissions](http://www.opensips.org/html/docs/modules/2.3.x/permissions.html) module ? White scripting and permissions module you can easily implement per user IP white/black list too, also with caching (see the addresses table).

dronord:
I cant understand how to use permission module.
permission-create.sql: grp SMALLINT(5) - it cant be username
hash.h: PERM_MAX_SUBNETS = 128 - its small for internet

liviuchircu:
Feedback after reviewing the code:

coding issues
---
* bad practice: structure declarations in .h files (e.g. `cur_data`, `bw_data1`, `bw_data2`). Luckily enough, you are only using them in a single .c file (`blackwhite.c`), so it works
* `qsort()`'ing the usernames seems redundant, as they seem to be already sorted by the DB query. Please correct me if I'm wrong.
* `data_append()` could use some refactoring, with some function parameter passing instead of maintaining that hard-to-grasp global state inside the function.

conceptual issues
---
* your module allows a user to have both a blacklist and a whitelist at the same time. This is a broken idea: not only is it hard to grasp (filter(filter(username))), but it could lead to difficult provisioning (e.g. admin whitelists an IP, but it has no effect, because the IP was already blacklisted). In theory, any set of dual blacklist+whitelist rules for a user can be refactored into either a single bigger blacklist or a single bigger whitelist. Simple is always better.
* based on my review, the only novelty of this code is the binary searching of usernames. The "permissions" module performs IP hashing over 128 buckets, so it will perform worse if rule number goes high, well into the tens of millions. Even if this were a problem, we would just recompile OpenSIPS with a bigger hash size on these platforms.

Cheers!

dronord:
**liviuchircu**, thank you for spent time.
> # coding feedback

> bad practice: variable definitions in .h files (e.g. cur_data, bw_data1, bw_data2). Luckily enough, you are only using them in a single .c file (blackwhite.c), so it works

> data_append() could use some refactoring, with some function parameter passing instead of maintaining that hard-to-grasp global state inside the function.

variable definitions and data_append() can be rewritten.
> qsort()'ing the usernames seems redundant, as they seem to be already sorted by the DB query. Please correct me if I'm wrong.

qsort() for case where db charset differ from program. I should to be sure bsearch() will work correctly. qsort() use same compare function cmpstringp() as bsearch(). qsort() process prepared data, it quickly.
If qsort() redundant - can be omitted.

> # conceptual feedback

> your module allows a user to have both a blacklist and a whitelist at the same time. This is a broken idea: not only is it hard to grasp (filter(filter(username))), but it could lead to difficult provisioning (e.g. admin whitelists an IP, but it has no effect, because the IP was already blacklisted). In theory, any set of dual blacklist+whitelist rules for a user can be refactored into either a single bigger blacklist or a single bigger whitelist. Simple is always better. Leave no room for complications.

In module defined order, chapter 1.4.1. blackwhite():
2. Find source IP of message in black list
4. Find source IP of message in white list
Like iptables. One address can be in many rules, first triggered - this is the order.
Web interface can warning about it, program cant decide.
We can leave a note in the documentation about it.

> it seems all this logic can be achieved with the permissions module, with the "username" stored as a pattern. Instead of looking up by username and matching an IP, permissions would look up by IP and match a username pattern. Tomato, tomato.

For example, we have rules. For user1 - block 1.1.1.1, user2 - pass only 213.253.24.2. Please write how it will look in the records of the address table of the module permission.

> a cool novelty introduced by this module is the binary searching of usernames. The "permissions" module performs IP hashing over 128 buckets. Indeed, this will definitely perform worse if rule number goes high, well into the tens of millions. But even if this were a real-life problem, we would just recompile OpenSIPS with a bigger hash size on these platforms. (Hint: we can always do some development to make the permissions hash size configurable!)

P.S. I dont understand why need IP hashing.

I am ready to edit/extend blackwhite module, but not to the detriment of the logic of black and white lists. You also have every right.

liviuchircu:
```
mysql> select * from address;
+----+-----+---------------+------+------+-------+---------+--------------+
| id | grp | ip            | mask | port | proto | pattern | context_info |
+----+-----+---------------+------+------+-------+---------+--------------+
|  6 |   1 | 1.1.1.1       |   32 |    0 | any   | user1   | NULL         |
|  14 |   2 | 213.253.24.2  |   32 |    0 | any   | user2   | NULL         |
+----+-----+---------------+------+------+-------+---------+--------------+
```

Where **1** is a blacklist, and **2** is a whitelist.

dronord:
And user3 have not rules...

liviuchircu:
Who cares about `"user3"`? That's the whole beauty of white or black lists: *the default policy*. He will be blocked or not, depending whether our script uses a whitelist (grp **2**), or a blacklist (grp **1**).

dronord:
**liviuchircu**, sorry, forget about user3. For me permission module is hard to understand for achieving black/white task.
What i mean, which scenario: we have 3 clients. First with black address, second with white, third without properties. I think default policy - pass packet. How to implement it in scope of permission module?
In blackwhite module - pass by default. If client(user3) not found, then $rc=-3, and we can block.
May be need redundant record 0.0.0.0/0 in group 2. But DBG:permissions:reload_address_table: invalid ip field in address table, ignoring entry with id 3.

```
MariaDB > select * from address;
+----+-----+--------------+------+------+-------+---------+--------------+
| id | grp | ip           | mask | port | proto | pattern | context_info |
+----+-----+--------------+------+------+-------+---------+--------------+
|  1 |   1 | 1.1.1.1      |   32 |    0 | any   | user1   | NULL         |
|  2 |   2 | 213.253.24.2 |   32 |    0 | any   | user2   | NULL         |
|  3 |   2 | 0.0.0.0      |    0 |    0 | any   | user3   | NULL         |
+----+-----+--------------+------+------+-------+---------+--------------+
```

liviuchircu:
The "0.0.0.0/0" solution seems clean, and it should work imo. It seems this behavior was introduced in commit 599caa61, which did not allow "0.0.0.0" IP addresses anymore for some reason.

volga629:
Will be nice if this module will  have ability to setup with ipset and not only with database. Ipset can provide fast way to insert or fetch ip's

volga629:
Example:

````
root@casbc00 ~> [/etc/opensips]# ipset list
Name: mgmt_acl
Type: hash:net
Revision: 6
Header: family inet hashsize 1024 maxelem 65536
Size in memory: 472
References: 24
Number of entries: 2
Members:

````

**Comment by liviuchircu** [comment]
@dronord the permissions module issue with `0.0.0.0/0` subnets has been fixed in c75a7ddcd1. Please let me know if blackwhite can do other things that aren't already achievable with existing modules.

**Comment by dronord** [comment]
**volga629**, blackwhite access DB only in case: on start and for refresh.

> Table readed in memory on start and can be refreshed by MI bw_reload.

You can to make ipset from opensips.blackwhite database table. What is your use case?

**Comment by dronord** [comment]
No any "already".
1. Patch permissions module for size of unneeded hash.
2. Insert redundant rows in table and keep it in mind in developing web interface or something else.
3. Call two times check_source_address() in script.

It is not hard to grasp?

By the same way we can use load_balancer or dispatcher or any other module to achieve same capabilities.

Simple is always better. (c)

**Comment by liviuchircu** [comment]
> Insert redundant rows in table and keep it in mind in developing web interface or something else.

This is optional. Here is how you can implement an _"accept by default"_ policy with black-white overlapping permissions groups (although, if you think about it, it doesn't really make sense: _"Why do I do whitelisting if I'm already accepting non-blacklisted users by default?!"):_

```
if (check_source_address(1, ...)) {
    /* blacklisted */
}
...
if (check_source_address(2, ...)) {
    /* whitelisted */
} else {
    /* accept-by-default */
}
```

**Comment by liviuchircu** [comment]
All I'm trying to say is that blackwhite doesn't reduce the complexity of provisioning at all. If it solves your problem by adding **N** rows in _"opensips.blackwhite"_ table, then it's possible to achieve the same result with **N** rows in _"opensips.address"_.

**Comment by dronord** [comment]
> Why do I do whitelisting if I'm already accepting non-blacklisted users by default?!

The first client want to block one address, second - to accept only one, third nothing wants:
> First with black address, second with white, third without properties.

```
if (check_source_address(1, ...)) {
    /* blacklisted */
}
...
if (check_source_address(2, ...)) {
    /* whitelisted */
} else {
    /* accept-by-default */
}
```
will block packets from 154.239.25.100 by user2 with white?

**Comment by liviuchircu** [comment]
No, they will pass through - I didn't say it was good scripting, I just implemented what you asked for :)

Regardless, I'm not sure I'm following anymore. Let's get back to your suggested scenario for a bit:

* user1 blacklist
* user2 whitelist
* user3 pass-through
* user4... block

This is **completely** broken. The idea of whitelisting is that it makes the edge OpenSIPS act as a firewall, **primarily** focused on the incoming IP address, and maybe some other metric only afterwards. With the above rule system, if I'm an attacker, all I have to do is to write "user3" in the From header, and voilà! I can immediately start routing calls through your OpenSIPS - my random public IP will surely go through, thanks to the broken rule for user3!

Secondly, I'm not sure why you call the "0.0.0.0/0" rule in permissions "redundant". Such whitelist addresses for user3 are mandatory in blackwhite as well, if you want to obtain a `-2` retcode. Otherwise, it will return `-3`, which is wrong.

**Comment by dronord** [comment]
Yes, i want OpenSIPS act as a firewall.
Which scenario could you wish to examine? Disable user3?
- in blackwhite module: set true client address to white list;
- block a service for user3.

**Comment by liviuchircu** [comment]
To conclude, regarding your last feedback:

> Patch permissions module for size of unneeded hash.
I'd be happy to review a PR on this, if you need this specific optimization.

> Insert redundant rows in table and keep it in mind in developing web interface or something else.
As stated above, this seems okay. blackwhite would use the _exact_ same rows anyway, and cannot compact them any better while staying within the same requirements.

> Call two times check_source_address() in script.
That's the least of our problems. I'd be more concerned with the security of the design ("user3" vulnerability I mentioned above), rather than worrying about two functions instead of one. Moreover, you can use GNU `m4` and do something like this, to make it more readable:

```
if (check_source_address(ADDRESS_GRP_BL, ...)) {
    /* blacklisted */
}
...
if (check_source_address(ADDRESS_GRP_WL, ...)) {
    /* whitelisted */
}
```

All in all, this is a nice pull request, but it adds little to no value with regards to existing OpenSIPS features.

Lastly, the binary sorting idea behind blackwhite is worth mentioning. Looking at "permissions", it currently has O(n) lookup times, with a 128-sized hash table, which makes it blazing fast the majority of use cases. Again, if this algorithmic performance is not enough for you, I'd be happy to review a PR which refactors the permissions code to use AVL trees (check out `map.h`), so we cut IP address string lookup times down to O(log n).

PR #968, opened 2016-10-19
Code added for REST_PUT API call

AgalyaR:
Support for REST_PUT API call 
Files modified:
rest_client.c
rest_methods.c
rest_methods.h

### DIFF
--- a/action.c
+++ b/action.c
@@ -45,6 +45,7 @@
  * \brief OpenSIPS Generic functions
  */
 
+
 #include "action.h"
 #include "config.h"
 #include "error.h"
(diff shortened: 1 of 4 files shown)


liviuchircu:
Thank you for taking the time to make the PR, @AgalyaR !

Looks good to me. There were issues with the indentation, and I took the time to fix them, since it was your first pull request :) As a general rule, we indent with tabs, not spaces. You should tune your text editor to display tabs as being 4-space wide.

The final logical step is to add documentation in `rest_client/doc/rest_client_admin.xml` for `rest_put()`.

PR #983, opened 2016-11-03
Spread timer wakeups across threads to prevent thundering herd

rrb3942:
The current timer method causes a thundering heard effect every time a timer is fired. This is because all children listen on the same timer pipe. Every write to that pipe cause all children to wake up, even though only the first one is able to grab the timer job. This isn't a very big problem at low children and timer counts, but as both of these grow it gets much worse.

This patch instead creates a timer pipe per worker child, up to TIMER_MAX_PIPES, at which point children start sharing timer pipes. The extra process that prevents timer starvation listens on all timer pipes, so it is still able to perform work if all children are busy.

When dispatching timer events we round-robin through the active timer pipes for each event. This allows us to balance timer events among the children. In case of an error we advance to the next pipe, eventually trying all pipes before giving up.

In most cases this results in only two threads being woken (the one child worker and the timer process) per timer event. This is in contrast to every child worker + the timer process being woken per timer event.

Some numbers (Taken from a VM running Centos 6 w/6 cores and watching vmstat on an idle system):
(Note: No TCP workers, just udp)

Pre-Patch:
Children = 30
TM timer_partitions = 16
Results: 1.5k-2k Interrupts/Sec 8-12k Context switches/Sec

Children = 200
TM timer_partitions = 16
Results: 8k-13k Interrupts/Sec 60k-80k Context switches/Sec

Post-Patch:
Children = 30
TM timer_partitions = 16
Results: 170-220 Interrupts/Sec 400-480 Context switches/Sec

Children = 200
TM timer_partitions = 16
Results: 260-350 Interrupts/Sec 880-1100 Context switches/Sec

The 200 children example is a bit extreme for this vm, but it demonstrates the scaling issues.

I think spurious wake-ups could be reduced further on kernels >= 2.6.35 by giving the timer process a dedicated pipe with a limited buffer size and only waking children when that becomes full. At the moment my testing systems are still on Cent 6, so they don't support it, but if there is interest I could work on adding that later.

This patch works on all systems and takes care of the bulk of the issue.

Because of the current regressive behaviour it might be worthwhile to look at applying this to the 2.2 LTS as well.

Appreciate any input and additional testing on this.

### DIFF
--- a/cfg.lex
+++ b/cfg.lex
@@ -354,6 +354,7 @@ DISABLE_STATELESS_FWD	"disable_stateless_fwd"
 DB_VERSION_TABLE "db_version_table"
 DB_DEFAULT_URL "db_default_url"
 DB_MAX_ASYNC_CONNECTIONS "db_max_async_connections"
+CHILDREN_PER_TIMER	children_per_timer
 DISABLE_503_TRANSLATION "disable_503_translation"
 
 MPATH	mpath
@@ -650,6 +651,8 @@ IMPORTFILE      "import_file"
 									return DB_DEFAULT_URL; }
 <INITIAL>{DB_MAX_ASYNC_CONNECTIONS}	{	count(); yylval.strval=yytext;
 									return DB_MAX_ASYNC_CONNECTIONS; }
+<INITIAL>{CHILDREN_PER_TIMER}		{	count(); yylval.strval=yytext;
+									return CHILDREN_PER_TIMER; }
 <INITIAL>{DISABLE_503_TRANSLATION}	{	count(); yylval.strval=yytext;
 									return DISABLE_503_TRANSLATION; }
 
--- a/cfg.y
+++ b/cfg.y
@@ -392,10 +392,10 @@ static struct multi_str *tmp_mod;
 %token DB_VERSION_TABLE
 %token DB_DEFAULT_URL
 %token DB_MAX_ASYNC_CONNECTIONS
+%token CHILDREN_PER_TIMER
 %token DISABLE_503_TRANSLATION
 %token SYNC_TOKEN
 %token ASYNC_TOKEN
-
 
 
 
@@ -1080,6 +1080,8 @@ assign_stm: DEBUG EQUAL snumber
 		| DB_MAX_ASYNC_CONNECTIONS EQUAL error {
 				yyerror("integer value expected");
 				}
+		| CHILDREN_PER_TIMER EQUAL NUMBER { children_per_timer=$3; }
+		| CHILDREN_PER_TIMER EQUAL error { yyerror("integer value expected"); }
 		| DISABLE_503_TRANSLATION EQUAL NUMBER { disable_503_translation=$3; }
 		| DISABLE_503_TRANSLATION EQUAL error {
 				yyerror("integer value expected");
--- a/globals.h
+++ b/globals.h
@@ -140,6 +140,9 @@ extern int db_max_async_connections;
 
 extern int disable_503_translation;
 
+/*Timer behavior*/
+extern int children_per_timer;
+
 extern int enable_asserts;
 extern int abort_on_assert;
 #endif
--- a/main.c
+++ b/main.c
@@ -236,6 +236,9 @@ int gid = 0;
 int disable_core_dump=0; /* by default enabled */
 int open_files_limit=-1; /* don't touch it by default */
 
+/* timer config options */
+int children_per_timer=0;
+
 #ifdef USE_MCAST
 int mcast_loopback = 0;
 int mcast_ttl = -1; /* if -1, don't touch it, use the default (usually 1) */
@@ -1161,7 +1164,8 @@ try_again:
 	}
 
 	/* init timer */
-	if (init_timer(children_no + tcp_children_no)<0){
+	/* Enforce 1 global timer pipe based on config */
+	if (init_timer(children_per_timer <= 0 ? 1 : ((children_no + tcp_children_no)/children_per_timer))<0) {
 		LM_CRIT("could not initialize timer, exiting...\n");
 		goto error;
 	}
--- a/main.c
+++ b/main.c
@@ -1161,7 +1161,7 @@ try_again:
 	}
 
 	/* init timer */
-	if (init_timer()<0){
+	if (init_timer(children_no + tcp_children_no)<0){
 		LM_CRIT("could not initialize timer, exiting...\n");
 		goto error;
 	}
--- a/net/net_tcp.c
+++ b/net/net_tcp.c
@@ -1688,7 +1688,7 @@ int tcp_count_processes(void)
 
 int tcp_start_processes(int *chd_rank, int *startup_done)
 {
-	int r, n;
+	int r, n, timer_pipe_id;
 	int reader_fd[2]; /* for comm. with the tcp children read  */
 	pid_t pid;
 	struct socket_info *si;
@@ -1717,6 +1717,10 @@ int tcp_start_processes(int *chd_rank, int *startup_done)
 			LM_ERR("socketpair failed: %s\n", strerror(errno));
 			goto error;
 		}
+
+		//Timer pipe id for this child to listen on
+		//This will cause it to wrap in the event of more children than pipes
+		timer_pipe_id = (*chd_rank) % timer_pipe_count;
 
 		(*chd_rank)++;
 		pid=internal_fork("SIP receiver TCP");
@@ -1758,7 +1762,7 @@ int tcp_start_processes(int *chd_rank, int *startup_done)
 
 			report_conditional_status( 1, 0);
 
-			tcp_worker_proc( reader_fd[1] );
+			tcp_worker_proc( reader_fd[1], timer_pipe_id );
 			exit(-1);
 		}
 	}
--- a/net/net_tcp_proc.c
+++ b/net/net_tcp_proc.c
@@ -114,7 +114,7 @@ inline static int handle_io(struct fd_map* fm, int idx,int event_type)
 
 	switch(fm->type){
 		case F_TIMER_JOB:
-			handle_timer_job();
+			handle_timer_job(fm->fd);
 			break;
 		case F_SCRIPT_ASYNC:
 			async_script_resume_f( &fm->fd, fm->data);
@@ -291,7 +291,7 @@ static inline void tcp_receive_timeout(void)
 
 
 
-void tcp_worker_proc( int unix_sock)
+void tcp_worker_proc( int unix_sock, int timer_pipe_id)
 {
 	/* init reactor for TCP worker */
 	tcpmain_sock=unix_sock; /* init com. socket */
@@ -300,7 +300,7 @@ void tcp_worker_proc( int unix_sock)
 	}
 
 	/* start watching for the timer jobs */
-	if (reactor_add_reader( timer_fd_out, F_TIMER_JOB, RCT_PRIO_TIMER,NULL)<0){
+	if (reactor_add_reader( timer_fds_out[timer_pipe_id], F_TIMER_JOB, RCT_PRIO_TIMER,NULL)<0){
 		LM_CRIT("failed to add timer pipe_out to reactor\n");
 		goto error;
 	}
--- a/net/net_tcp_proc.h
+++ b/net/net_tcp_proc.h
@@ -27,6 +27,6 @@
 #define _NET_net_tcp_proc_h
 
 /* Loop implementing a TCP worker */
-void tcp_worker_proc( int fd);
+void tcp_worker_proc( int fd, int timer_pipe_id);
 
 #endif
--- a/net/net_udp.c
+++ b/net/net_udp.c
@@ -261,7 +261,7 @@ inline static int handle_io(struct fd_map* fm, int idx,int event_type)
 			update_stat( pt[process_no].load, -1 );
 			return n;
 		case F_TIMER_JOB:
-			handle_timer_job();
+			handle_timer_job(fm->fd);
 			return 0;
 		case F_SCRIPT_ASYNC:
 			async_script_resume_f( &fm->fd, fm->data);
@@ -286,7 +286,7 @@ inline static int handle_io(struct fd_map* fm, int idx,int event_type)
  * \see main_loop
  * \return -1 for errors
  */
-int udp_rcv_loop( struct socket_info *si )
+int udp_rcv_loop( struct socket_info *si, int timer_pipe_id)
 {
 
 	/* create the reactor for UDP proc */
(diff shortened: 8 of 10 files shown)


wdoekes:
Nice work!

I'm not intimate with the code, so I won't review it. But I'll point out the `s/heard/herd/` spelling error.

liviuchircu:
Very nice before/after analysis!

I reviewed the patch, and my only note is related to its behavior with multiple UDP interfaces (say, 4-5 listeners, which is a fairly common usage). Even in that case, fd assignment would only wrap around `TIMER_MAX_PIPES` a couple of times, leading to a maximum of 3-4 processes sharing the same pipe fd, and being all woken up by a single timer event. In other words, it seems to work well even in the roughest of conditions!

Overall, great work and a nice addition. LGTM.

rrb3942:
@wdoekes Can't believe I missed that in my commit message AND the pull request. It has been fixed.

@liviuchircu You actually need over 100 children (between udp and tcp) before it starts to wrap. I figure at that scale a few extra wake ups aren't a big deal. I just didn't want to spawn pipes without bounds. In the case of less than 100 children (which I believe is most use cases), they each get their own pipe. At least, that is what is supposed to happen.

bogdan-iancu:
Hi @rrb3942 , let me share some thoughts on the how balancing the timer jobs to all the processes.

There are 2 main requirements here:
1) to be sure to you find all the time a free process able to handle the new timer job
2) the execution of the timer job must be with minimum of delay (buffering).

So, the idea of having a single pipe read by all processes guarantees 100% that the job will be picked up by any idle worker process. It is important to be picked up by an idle process, so we can guarantee 2) - an realtime timer triggering. With the unique pipe, does not matter which proc is available - if we have only one ready to read from pipe, the timer job will be picked up.
This guarantee that we find all the time the available procs to handle the timer job on the spot, with no delays. 
I was fully aware of the context switching storm that happens when same "read" event is delivered to all processes (listening on the pipe), but only one doing the actual read. But this is a drawback under low load, where it is not a problem. The unique pipe approach will provide the maximum of efficiency (this drawback will disappear) under high load as:
* more and more procs are busy doing other tasks, so less actually listening on the pipe (at a certain time), so the context-switching storm will become less and less with higher and higher load
* more and more timer tasks are generated under high loads, so even if task N generates a read event on the pipe in several procs, one will read task N, but other already awaken may read the following M timer task.
So basically we make opensips more efficient under heavy load on the expense of the low load inefficient processing :) 

Now, looking at the approach you optimize the low load (reduces the amount of context switches when you have many idle processes), but:
1) when you write a task to one of the pipes, you are not aware of the availability of the procs reading from that pipe - they may be busy or not; so you may end up pushing the task to a subset of procs that are busy, while others are idle -> it fails to find the available procs to take the job, breaking 1)
2) you stop pushing tasks on a pipe when the pipe get full (the write fails as it will block). That means you allow time jobs to be buffered/queued on the pipes, which breaks 2) - the jobs must be as realtime as possible, with zero buffering (as time as you have free workers).

Maybe there is something I missed on your patch, I will be more than happy see if there is a way to improve the current implementation.

Regards,
Bogdan

rrb3942:
@bogdan-iancu Thanks for the feedback.

To address your concerns:

1) There are always at minimum, two processes waiting on each pipe, the child worker process and the independent timer process (which listens on all timer pipes). I believe this is enough to ensure that the timer gets handled in a quick manner, but this could be adjusted. Lets say we make (children_no + tcp_children_no)/2 timer pipes, we would then have minimum 2 worker children + the timer process per pipe (for a total of 3). Worst case is that the timer processes has to handle the task, because the other child(ren) are busy, which is the current worst case as well. Also since we are evenly distributing the tasks among children, if everything is already busy we should see similar performance as compared to the current implementation under heavy loads.

2) Each timer gets pushed to a different pipe in a round-robin fashion, but we fail-over to a new pipe in the event of a write error. We are just not writing until the first pipe is full, then moving on to the next. We are advancing the pipe for every task. This gives us an even distribution of tasks. If we are at a point where pipes are filling up then nothing has been able to service timer tasks at all and things are really broken. This also lets us queue up tasks for the workers if many timer tasks fire at once to the point were we loop around the pipes, and allows for the same scenario in your second bullet point above (a thread that is already awake can handle multiple tasks that are waiting).

In general, I think it is a pretty bad experience to start up an application and see it immediately cause load while being idle. It can make judging actual load harder as well, as you have a baseline level of busy that can mask actual increasing load.

The other argument I would make is that the excess timer checking by all the processes is just busy work, taking CPU time from processes with actual work to do. While you are correct in saying that if a thread is already awake it can handle multiple jobs, it might also get kicked to the back of the run-queue because another process woke-up to check for that same job it just took, effectively delaying the timer. We are also not just reducing Context-Switching, but Interrupts as well, and no helpful work is being done in those interrupt handlers.

I don't have a load testing scenario currently setup. Some medium and high load scenarios might help determine if there is any regression at those levels, although I don't think there will be.

I think one important difference between the current implementation and my proposed one, is that the proposed one is adjustable. If TIMER_MAX_PIPES is set to 1, you will effectively have the same behaviour as the current implementation (one global shared pipe). What if I add a 'TIMER_CHILDREN_PER_PIPE' as well, that can be used to adjust the number of workers that listen per pipe? Between TIMER_MAX_PIPES and TIMER_CHILDREN_PER_PIPE, it could be adjusted for reduced load or for timer importance? Then users could make the decision (of course sane defaults would need to be determined).

rrb3942:
@bogdan-iancu Please review the latest commit.

It adds a 'children_per_timer' configuration option. It sets the minimum number of children that will be woken per timer event. The default (0) causes all children to listen on a single pipe, which is the same behaviour as the current implementation. Setting it to 1 causes each child to get its own pipe (unless TIMER_MAX_PIPES is exceeded). Setting it to 10 would cause at least 10 workers to listen per pipe, if there are less than 10 workers all children listen on the same pipe. 

This way it can be configured to the degree that the user wants to reduce load or increase timer priority.

Please note that in addition to the above, the separate timer process is always waiting on all timer pipes.

Numbers at idle for 200 children and 16 TM timer partitions:
Default (0):
Results: 8k-13k Interrupts/Sec 60k-80k Context switches/Sec

100:
Results: 1.5k-2k Interrupts/Sec 27k-30k Context switches/Sec

50:
Results: 600-1.6k Interrupts/Sec 12k-19k Context switches/Sec

10:
Results: 300-600 Interrupts/Sec 2.1k-3.6k Context switches/Sec

1:
Results: 250-360 Interrupts/Sec 800-1.1k Context switches/Sec

bogdan-iancu:
@rrb3942 , the problem we addressed with this timer pipe (starting with 2.1) was the timer bottleneck. Why squeezing all timer tasks in a single (or 2 or 3) processes (and note that the timer tasks sometimes are really time consuming) when you have so many worker processes that may sit and do nothing (idle processes in SIP handling).
So, to solve this issue, we must find ANY idle process to take over a time task in realtime, with no delay.

If the pipe is not shared between all processes able to handle a timer job, you cannot guarantee the "ANY" part of the above requirement. Distributing a job to a subset of processes will certainly fail (at some point) in locating a free/idle process:
Lets assume we have a pipe between P1, P2 and timer process. We also have P3 and P4 procs. If I put a timer job in the pipe and P1 and P2 are stuck in doing SIP handling (for INVITEs it may take some time) and the timer process is stuck in doing retransmissions or pings)...you will simply fail to use P3 or P4 (which may be idle) and compromise the timer accuracy.

You say that probabilistic speaking is ok, but it is not enough - we need a certainty. The certainty that even only one proc is available, we are able to push to it the timer job RIGHT NOW.

Another aspect which gets affected by multiple pipes - some timer jobs may be blocked in the pipe (as when dispatching the jobs you have no idea if there is any available proc to handle it). And the timer jobs are time-sensitive (as triggering order). So, with multiple pipes you may end up with a task triggered later to be executed long before one triggered earlier. 

On the other hand I understand your points in terms of unnecessary load due context switching. But IMO this is the small bad thing we have to live with with (during 0 or small load) in order to guarantee best performance under medium/high load. It is a compromise (which actually has no visible or problematic effect).

A similar problem we have right now with the dispatching of the TCP connection from the TCP main to the TCP workers - the current implementation relays on one-pipe-per TCP worker and it sucks big time. As we end up assigning new TCP conns to processes that are not idle (while others are sitting doing nothing).

I think you cannot do proper balancing without having a single central mechanism (in our case the pipe) which allows the workers to come (when available) and pick up the tasks to be executed.

rrb3942:
@bogdan-iancu Please see my timer_first_idle branch at https://github.com/rrb3942/opensips/tree/timer_first_idle 

That branch has each worker with their own pipe and they flag themselves idle right before waiting on the event loop. When they return from waiting for the event loop they flag themselves busy until they return to waiting on the loop. The timer process iterates through the workers looking for an idle worker and sends that one the job. This way we avoid sending timer work to any worker that is actively busy.

If no worker is idle it will put the timer job on a global timer pipe that all workers are listening on. This way the first worker that is spare is able to get to the timer.

The separate timer process still waits on all of the workers pipes and the global pipe to help prevent starvation.

I think this covers all of your concerns. There is some additional synchronization (this is our central mechanism) before and after waiting on the event loop, but waiting on the event loop is already "heavy" and the additional overhead should be minimal. I believe most of the polling methods also preserve the order that we register descriptors. Since we registers the timer descriptors first, even if additional events are waiting the timers should be processed first on the event wake-up.

The worker state could be used for your TCP worker issue as well.

I need to review the changes some more, but if you like that approach/direction let me know and I will close this pull and make a new one for that branch.

razvancrainea:
I'll drop some ideas here, hopefully they will be useful.

@rrb3942 In my opinion this approach definitely improves the initial problem, but does not solve it completely, because there is a small gap between the time you decide the worker is idle and you dispatch the command to him. During this gap, if that worker receives something huge (in terms of processing time) to do and the poll is triggered, you will delay the job until either him or the designated timer finishes.
Although this seems a corner case, I think it might happen pretty frequent during high traffic, especially because the `poll` will be triggered so fast. And we will end up having some jobs buffered, that could have been potentially processed by someone else, thus breaking 2.

IMO, our main focus should be optimizing during high load (that's when the timers started to drift, that's why we needed to spread timer jobs across across multiple processes), and not during idle times, where there's not such a big problem. Therefore, may I ask what is the impact/experience of the high number of context switches? Could we perhaps address it in a different way?

PS: note that the atomic builtin primitives (i.e. `__atomic_store_n`) have been introduced in [gcc 4.7](https://gcc.gnu.org/wiki/Atomic/GCCMM); for the final release, we need to work this out in a different way, perhaps enhance OpenSIPS's [atomic interface](https://github.com/OpenSIPS/opensips/blob/master/atomic.h).

Awesome work! Looking forward to get to a final solution :D!

rrb3942:
@razvancrainea I can look at alternatives for the atomics, that option was just readily available.

I think fixing that small gap would mean changing the event loop some. If, after flagging 'Busy' the loop always processed (whether an event was raised of not) its individual timer pipe immediately it would solve that. The question is if we are OK with the overhead of always checking that pipe on every wake-up. From the discussions it seems to me that we want to prioritize timers, so that might be acceptable. For the global timer pipe, should it also be processed on every wake-up, or only when triggered? If this is done, I think we could also make it so that the extra timer process only listens on the global timer pipe, as we have guaranteed that the worker will get the timer job.

From a sysadmin point of view: How do we know if the system is under heavy load if the load average is always high? :) I have monitoring configured to alert when load_avg > X. If I were to deploy the current 2.1+, I might always be over that even at idle.

The goal here is to make the scaling predictable (with minimal/no performance regression), so you can monitor system load and go 'Oh, I might be hitting a bottleneck soon' and plan accordingly, as opposed to "Well, my load average is always 5". Many of the tools on linux to check for performance issues report cpu usage, interrupts/sec, context switches/sec, disk io, io wait, etc because they are good indicators to determine what is happening on a system. Unnecessarily inflating some of of these metrics just makes it harder to diagnose a problem.

Also, what is good at heavy load can actually cause performance regression under light and medium load. So if we can optimize light and medium load, it may take longer to get to heavy load. Which of these sounds like it is doing less work from an OS/CPU usage point of view?

A) OS schedulers tend to favour processes that have been waiting on IO. This means if we can avoid sending a global signal, and only wake-up an idle process it will immediately get run time to handle the timer.

B) With a global signal, everyone wakes up, the first on gets the job, but can then be put to the back of the run queue as all the other processes check for IO. In effect this means the timer might get delayed. The only upside is that if the system is otherwise not busy that delay might not be too bad. The downside is that if other services are running on the system (or you are in a shared VM) you might be causing scheduler thrashing for them in attempts to wake-up all the children.

If we don't spend our time checking for work we won't get, we will have more time for real work (timers, sip messages, etc), which means it will take longer to hit a point where the global signal will be beneficial. In my general experience, when I see behaviour like this it often becomes a bottleneck, and when you can eliminate it you often find it is much harder to hit the 'heavy load' scenario and you have more headroom than you might otherwise have expected.

The current implementation also limits over-provisioning (say in VM's) and lets a user possibly overload their system just by spawning extra worker processes that would otherwise be idle.

If you look at it from an opensips point of view, it's best if we just spin at 100% cpu checking for work constantly. But if you look at it from an ecosystem point of view (OS, Other services, VMs) you want to be a good neighbour and eliminate unnecessary noise so everyone can get their work done in a timely manner.

I think the real issue that I might have uncovered (from the development standpoint) is that we don't have a good method to ensure timers are handled when we want them to be, so right now the global signal is the hammer we are using to help mitigate that problem.

liviuchircu:
According to [this article](https://lwn.net/Articles/667087/), thundering herd problems related to `epoll()` can be mitigated via the freshly added `EPOLLEXCLUSIVE` flag for `epoll_ctl()`. The flag is available starting with Linux kernel 4.5+.

I suggest we looked into coming up with a patch that introduces usage of this flag on appropriate glibc versions. While it wouldn't immediately take care of the idle resource usage on existing deployments, at least we can say that the problem will be fixed for future ones.

razvancrainea:
If you have a 4.5+ kernel available, just try out this patch and let us know how it behaves :):
https://gist.github.com/razvancrainea/1b61d595d4fc415aa122ca7967bd41b5#file-epoll_exclusive_trigger-diff

rrb3942:
Sorry I hadn't responded to this in a while. Holidays and other projects.

That looks really promising and does exactly what we want. Unfortunately I don't have a 4.5+ kernel system to test it on, but looks like it is worth exploring.

I was also curious if there would be interest in making timers work via a form of checkpoint preemption. This would mean having checkpoints to see if there was timer work to be done placed at strategic locations. Initially I think it could check after each sip message processed, and it could also check before actions that could take awhile (e.g. a blocking db query).

It would add some overhead to SIP message processing, but I think that can be kept low if the checkpoints are limited and placed well. The benefit would be that timers would fire faster and would not have to wait as long to be processed. This would also give timers even more priority handling.

bogdan-iancu:
@rrb3942 , thanks for pointing out the issue and for the great ideas resulting from this brainstorming. Razvan pushed a fix based on EPOLLEXCLUSIVE. 
Let us know if you have any feedback on this. 
I will close this PR as not applicable anymore.

Issue #992, opened 2016-11-15
SST module is not inserting Required: timer header on response


etamme:
See the last sentence of the second paragraph in https://tools.ietf.org/html/rfc4028#section-8.2

A "Require: timer" header must be inserted (or timer appended to an existing Require header).  The lack of a Require: timer header causes some user agents to not send any update requests at all.

PR #994, opened 2016-11-21
move appending require: timer to 200 OK where it belongs

etamme:
The logic for this fix originally put a Require: timer header on the request which is not what RFC 4028 specifies.  

> if the UAS does not support Session Timers, a proxy
can add the Session-Expires header initially negociated in the reply, but
must also add the "timer" extension in the Require header.

I have moved appending the Require: timer header to the section where the reply is processed and tested the change.  This resolves issue #992.


etamme:
It looks like I goofed when I ported this to 1.10 and thought it was a problem with the upstream commit.  Sorry about that.

PR #1010, opened 2016-12-26
Minor fix to the README to reflect the URL better

greenfieldtech-nirs:


razvancrainea:
Hi, Nir!

Thanks for helping the OpenSIPS community to get a better experience when reading OpenSIPS docs. Unfortunately the fix you provided is not complete. The reason is that the `README` file is automatically generated (by `make modules-readme`) from an XML file. Therefore when the documentation is re-generated, the fix will be overwritten.
To fix these kind of issues, we need to change the way URLs are displayed in a `README` file, or change the XML. Please file a bug request for this.

Best regards,
Răzvan

PR #1026, opened 2017-01-21
Enable curl handle and connection reuse for synchronous requests

rrb3942:
This patches makes all of the synchronous methods in the rest_client share and reuse a curl handle. This has the benefit that the curl handle will automatically reuse TCP connections (when possible) for subsequent HTTP requests, rather than opening a new TCP connection for each request. This can greatly reduce latency and overhead for HTTP requests to the same host, especially when RTT is not inconsequential.

I attached a couple example captures of the HTTP behaviour after applying this patch, demonstrating the reuse of the underlying TCP connection.

[examples.zip](https://github.com/OpenSIPS/opensips/files/720757/examples.zip)


liviuchircu:
This is an awesome optimization, @rrb3942 ! It seems that `curl_easy_cleanup()` and `curl_easy_reset()` handle the underlying TCP connections quite differently - first one shuts them down, while the other one does not touch them at all.

Patch looks good - I also tested it, even with multiple connections, and it behaved as expected. Merging into `master` for now, but this seems good enough of a fix to warrant a backport into `2.2`.

Issue #1039, opened 2017-02-03
drouting: high number of gateways can cause memory corruption


c-leroy:
If a route has more than 32 gateways, drouting can erroneously zero valid data when resizing destination list.

bogdan-iancu:
Hi @c-leroy - that is a good catch and a good fix - please create a pull request out of your commit, so we can accept it on the master tree.

c-leroy:
Thanks

PR #1040, opened 2017-02-04
drouting: Fix out-of-bound initialization

c-leroy:
Upon load/reload of drouting tables, if parse_destination_list() needs to
increase destinations array size, it asks for a new array twice the size of
the previous one.
It memset() the newly allocated capacity, skipping memory that is to be filled
with already learnt destinations, but using the array size instead of the
remaining capacity.
This patch fixes the size for this memset().

Closes OpenSIPS/opensips#1039


bogdan-iancu:
Thank you @c-leroy !

PR #1054, opened 2017-02-14
Make sure to call MD5 utility properly on FreeBSD ("md5" vs "md5sum" in linuxes).

sobomax:
This fixes build on FreeBSD.


razvancrainea:
Merged in upstream.

PR #1055, opened 2017-02-14
Added enable/disable replication to drouting and load_balancer

JeremyMartinez51:
I created four new opensips.cfg modparams replicate_lb_status_to, accept_replicated_lb_status, replicate_gw_status_to, and accept_replicated_gw_status. These commands use the clusterer module and were based off of the previously implemented clusterer commands used by the dialog and usrloc modules.

replicate_gw_status_to and accept_replicated_gw_status can be tested with the command:
opensipsctl fifo dr_gw_status

replicate_lb_status_to and accept_replicated_lb_status can be tested with opensips the command:
opensipsctl fifo lb_status

### DIFF
--- a/modules/drouting/dr_replication.c
+++ b/modules/drouting/dr_replication.c
@@ -1,3 +1,28 @@
+/*
+ * Copyright (C) 2017 OpenSIPS Project
+ *
+ * This file is part of opensips, a free SIP server.
+ *
+ * opensips is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * opensips is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ *
+ * history:
+ * ---------
+ *  2017-02-15 created by Jeremy Martinez
+ */
+
 #include "../../ip_addr.h"
 #include "../../ut.h"
 #include "prefix_tree.h"
--- a/modules/drouting/dr_replication.c
+++ b/modules/drouting/dr_replication.c
@@ -1,0 +1,49 @@
+#include "../../ip_addr.h"
+#include "../../ut.h"
+#include "prefix_tree.h"
+#include "dr_partitions.h"
+#include "dr_replication.h"
+
+str repl_dr_module_name = str_init("drouting");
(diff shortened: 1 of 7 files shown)


bogdan-iancu:
Thank you @JeremyMartinez51 for your contribution - I will review it shortly, but it definitely looks like a very nice addition. 
In the mean while, please add the license and copyright text to the newly added files.

modules/drouting/dr_replication.c
```diff
@@ -0,0 +1,74 @@
...
+{
+	if (bin_init(&repl_dr_module_name, REPL_GW_STATUS_UPDATE, BIN_VERSION) != 0) {
+		LM_ERR("failed to replicate this event\n");
+		return;
+	}
+
+	bin_push_int(clusterer_api.get_my_id());
+
+	bin_push_str(&gw->id);
+	bin_push_int(gw->flags);
```
bogdan-iancu: Pushing all the flags is not correct, as not all the flags are status/state related - see https://github.com/OpenSIPS/opensips/blob/master/modules/drouting/prefix_tree.h#L57.
So, only flags covered by `DR_DST_STAT_MASK ` should be replicated. Otherwise you may replicate the pinging state or the DB flushing state.

modules/drouting/dr_replication.c
```diff
@@ -0,0 +1,74 @@
...
+
+	bin_pop_str(&id);
+	bin_pop_int(&flags);
+
+	lock_start_read(head_db_ref->ref_lock);
+
+	gw = get_gw_by_id( (*head_db_ref->rdata)->pgw_tree, &id);
+	if (gw && (gw->flags != flags))
+	{
+		gw->flags = flags;
```
bogdan-iancu: Similar to the previous comment, only the status-related flags should be imported. Even more, if the imported flags make a difference in the gateway status, you need to :
- set DR_DST_STAT_DIRT_FLAG to force flushing of the new state to DB
- trigger the event for new state.

modules/drouting/drouting.c
```diff
@@ -537,6 +553,10 @@ static void dr_raise_event(pgw_t *gw)
 {
 	evi_params_p list = NULL;
 	str *txt;
+	if ((gw_status_replicate_cluster > 0) && gw)
```
bogdan-iancu: You should not trigger the replication from this event function - please see my comment on the replicate_gw_status_update() function, where you should trigger the event if the incoming replication changed the gw status. And in the current format this will lead to a replication storm :). Maybe dr_raise_event() should take one more flag (like do_replication) ; and when called from replicate_gw_status_update() to set it on FALSE.

modules/load_balancer/lb_data.c
```diff
@@ -1206,6 +1208,10 @@ void lb_raise_event(struct lb_dst *dst)
 {
 	evi_params_p list = NULL;
 
+	if ((lb_status_replicate_cluster > 0) && dst)
```
bogdan-iancu: All the issues mentioned for DR are also valid for LB too.

bogdan-iancu:
@JeremyMartinez51, your PR was manually merged in, together with my fixes to addressed all the mentioned issues - see bc5540
Thank you for your contribution.

PR #1056, opened 2017-02-15
More comprehensive fix for the fixup-related crashes in the python.so module (follow-up to #1050)

sobomax:
Prevent fixup function from trying to deallocate string allocated by the python interpreter by creating a copy of it and passing that down the pipe.

Improve failure handling by not leaking a memory while I am here.


sobomax:
Too old. GC.

PR #1057, opened 2017-02-15
Make it possible to listen at all interfaces again

andrey-vorobiev:
There was a nice feature in OpenSIPS 1.11. It used to listen at all available interfaces when no listeners are declared in configuration file. It's useful in case of multiple network interfaces because it's easy to provide default configuration when IP or network interfaces are not known in advance. Unfortunately this feature has been lost after SIP transport protocols rework. So since OpenSIPS 2.1 if no listeners are declared in configuration file via 'listen=...' statement OpenSIPs will not bind to any network interface.

I'd like to restore old behaviour, i.e. if some transport module is loaded via 'loadmodule' statement in configuration file and no listeners are declared for this module via 'listen=...' statement then listeners list will be populated with all network interfaces that are currently available.

### DIFF
--- a/net/trans.c
+++ b/net/trans.c
@@ -96,12 +96,7 @@ int trans_load(void)
 				return -1;
 			}
 			found_proto = 1;
-			/* check if there is any listener for this protocol */
-			if (!protos[pi.id].listeners) {
-				LM_DBG("No listener defined for proto %s\n", pi.name);
-				continue;
-			}
-
+			
 			/* check if already added */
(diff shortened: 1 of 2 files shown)


razvancrainea:
It's not correct to have opensips detect interfaces, because you don't know which ports they should be listening on. Instead, we've implemented a different approach, where you only have to specify the protocol you want to listen on all interfaces:
```
listen = udp:*
```
This has been pushed in 0271920abf2c38ebe44a33b0e8b41839ff2db75b

PR #1062, opened 2017-02-25
Replace *-dev dependencies with usual one for Debian packages

andrey-vorobiev:
Some Debian packages produced by OpenSIPS have '*-dev' packages as dependencies. For instance 'opensips-restclient-module' package depends on 'libcurl4-gnutls-dev' package. As a result when 'opensips-restclient-module' installed a lot redundant packages are being installed as 'libcurl4-gnutls-dev' dependencies such as: 'libkrb5-dev', 'librtmp-dev' and so on. And this is a minor problem.

There is also another problem: in Debian '\*-dev' package usually depends on non '\*-dev' package via '=' dependency so if I need patched version of some package on which 'libcurl4-gnutls-dev' depends via direct or transitive dependencies, for instance librtmp I can't install my custom version of librtmp and opensips-restclient-module at the same time.

So my proposal is to replace '*-dev' dependencies with usual one.

Following changes have been made:
- 'libssl-dev' package has been added as build dependency because some modules require libssl at build time
- 'libssl-dev' dependency has been replaced with 'libssl1.0.0' for following packages: 'opensips-tlsmgm-module', 'opensips-tls-module' and 'opensips-wss-module'.
- 'libcurl4-gnutls-dev' dependency has been replaced with 'libcurl3-gnutls' for 'opensips-restclient-module' package.
- 'libsctp-dev' dependency has been replaced with 'libsctp1' for 'opensips-sctp-module' package.

According to https://packages.debian.org/wheezy/libsctp1 libsctp1 is present both in Debian Wheezy and Jessie. The same is true for libssl1.0.0 (https://packages.debian.org/wheezy/libssl1.0.0) and libcurl3-gnutls (https://packages.debian.org/wheezy/libcurl3-gnutls).

And for Ubuntu: http://packages.ubuntu.com/search?keywords=libcurl3-gnutls, http://packages.ubuntu.com/search?keywords=libssl1.0.0, http://packages.ubuntu.com/search?keywords=libsctp1

### DIFF
--- a/packaging/debian/common/control
+++ b/packaging/debian/common/control
@@ -28,6 +28,7 @@ Build-Depends: bison,
                librabbitmq-dev,
                libsctp-dev,
                libsnmp-dev,
+               libssl-dev,
                libxml2-dev,
                libxmlrpc-core-c3-dev,
                pkg-config,
@@ -515,7 +516,7 @@ Description: Emerrgency call module for OpenSIPS
 Package: opensips-tlsmgm-module
 Architecture: any
 Multi-Arch: same
-Depends: libssl-dev,
+Depends: libssl1.0.0,
          opensips (= ${binary:Version}),
          ${misc:Depends},
          ${shlibs:Depends}
@@ -529,7 +530,7 @@ Description: TLS management module for OpenSIPS
 Package: opensips-tls-module
 Architecture: any
 Multi-Arch: same
-Depends: libssl-dev,
+Depends: libssl1.0.0,
          opensips (= ${binary:Version}),
          opensips-tlsmgm-module (= ${binary:Version}),
          ${misc:Depends},
@@ -544,7 +545,7 @@ Description: TLS transport module for OpenSIPS
 Package: opensips-wss-module
 Architecture: any
 Multi-Arch: same
-Depends: libssl-dev,
+Depends: libssl1.0.0,
          opensips (= ${binary:Version}),
          opensips-tlsmgm-module (= ${binary:Version}),
          ${misc:Depends},
@@ -559,7 +560,7 @@ Description: WebSocket Secure (WSS) transport module for OpenSIPS
 Package: opensips-sctp-module
 Architecture: any
 Multi-Arch: same
-Depends: libsctp-dev,
+Depends: libsctp1,
          opensips (= ${binary:Version}),
          ${misc:Depends},
          ${shlibs:Depends}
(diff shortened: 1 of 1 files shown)


razvancrainea:
I am not sure I can follow why you replaced the `libcurl4-gnutls-dev` dependency with `libcurl3-gnutls`. Is it because the first one comes with extra dependencies?

I've been working for a while with Debian Maintainers to add opensips in the Debian repository:
https://tracker.debian.org/pkg/opensips

I have not yet managed to merge all the changes I've done for the sid distribution into the upstream sources, but I was planning to do this soon. Basically the Debian approach is to remove all the `*-dev` Build dependencies and let the `${shlibs:Depends}` macro figure out what the necessary libraries. This is how the Debian Control file looks like:
https://tracker.debian.org/media/packages/o/opensips/control-2.2.2-3

What do you think about this approach? Is it suiting you?

andrey-vorobiev:
> What do you think about this approach? Is it suiting you?

It's much better than my proposal and completely suits my needs.

> Is it because the first one comes with extra dependencies?

Yes. And also there is another problem with declaring '*-dev' packages in 'Depends' section. As far as we can see from 'apt-cache show libcurl4-gnutls-dev' output
```
apt-cache show libcurl4-gnutls-dev
...
Depends: libcurl3-gnutls (= 7.26.0-1+wheezy13), libc6-dev | libc-dev, libgnutls-dev, libidn11-dev, libkrb5-dev, libldap2-dev, librtmp-dev, libssh2-1-dev, zlib1g-dev
...
```
'libcurl4-gnutls-dev' depends on 'libcurl3-gnutls' via '=' (exactly equal) relation. It means that in order to install 'opensips-restclient-module' package I need 'libcurl4-gnutls-dev' package as dependency which requires 'libcurl3-gnutls' exactly at version '7.26.0-1+wheezy13' as dependency. But if I need patched version of 'libcurl3-gnutls' from my private Debian repository for instance at version '10:7.26.0-mypatch1' I have a conflict because at one hand I need 'libcurl3-gnutls' at version '10:7.26.0-mypatch1' and at the other hand I need 'opensips-restclient-module' which needs 'libcurl3-gnutls' via transitive dependencies at '7.26.0-1+wheezy13' version. (Actually in my case problem relates to librtmp but I decided to use 'libcurl3-gnutls' as example in order to shorten reasoning)

So this PR is not actual anymore and can be closed. Thank you.

PR #1065, opened 2017-03-08
Update opensips.init on 2.1 to avoid init error

penguinjournals:
Added a whitespace on init script due to a failure on restarting on ubuntu.


razvancrainea:
Thanks for pointing this out. OpenSIPS 2.2 and master have this fixed, not sure how we missed it in 2.1.

PR #1067, opened 2017-03-08
Fix what seems to be a typo in the commit e2a6df0c16c42

sobomax:
Fix what seems to be a typo in the commit e2a6df0c16c42:

get_lock -> gen_lock_t

This unbreaks build with the USE_PTHREAD_MUTEX, which is when
this symbol is not defined.


razvancrainea:
Good catch, thanks!

Issue #1073, opened 2017-03-13
Topology Hiding: Wrong Callid in out of dialogy replies


rrb3942:
I am using topology hiding with force_dialog enabled and callid hiding.

When a request is received from the called party for a dialog that does not exist the generated reply contains the decoded callid, instead of the hidden one.

I have attached an example capture where two BYEs overlap. The BYE from the callee fails a topology_hiding_match() check, and when I try to send back a 481 the wrong callid is present in the reply.

[Overlap_BYE_Bad_Callid.zip](https://github.com/OpenSIPS/opensips/files/839153/Overlap_BYE_Bad_Callid.zip)

Opensips Ver:
version: opensips 2.2.3 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-FUTEX-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
git revision: 52813f4
main.c compiled on 16:33:40 Mar  7 2017 with gcc 5.2.1

rrb3942:
Looks like topo_callid_post_raw() only processes the reply if the dialog exists, where as topo_callid_pre_raw() will decode the callid so long as the callid prefix matches. I'm assuming topo_callid_pre_raw() doesn't check the dlg status because it is triggered very early in processing, before it would be available.

Seems like either topo_callid_pre_raw() needs to not decode the callid once a dialog is gone (which may not be possible to do in this call back) or we need to set a flag on the message so that either the callid can be re-encoded before it hits the script (or after topology_hiding_match() fails, the script writer will see the encoded callid) or pass the flag to the reply and have topo_callid_post_raw() check that flag if the dlg is not available and re-encode the callid on replies.

Does that seem correct? I'm not all that familiar with how all the callbacks work.

Thanks.

rrb3942:
See pull request #1076

rrb3942:
Sorry for the slow response, been out sick.

I don't mind re-writing the fix if you have some guidance on a better way to do it.

Maybe somehow flag the response for encoding if the request was decoded and the dlg check fails in the post_raw callback?

[...]

PR #1076, opened 2017-03-15
Fix callid in local replies for out of dialog topo hidden requests

rrb3942:
This patch fixes replies generated for out of dialog requests where topology hiding was used. Specifically, requests from the caller side that contain the encoded callid, which is decoded when the request is received. Currently if the dialog no longer exists the reply does not properly re-encode the callid in locally generated replies, which means they caller does not recognize our replies.

This patch causes us to keep a copy of the original callid if we decoded the callid for topology hiding and this is used for locally generated replies.

See issue #1073.

Initial testing looks good, but I plan to do some further testing.
If this fix looks acceptable it should be back-ported to 2.2 and probably 2.1.


rrb3942:
Is it possible to have the travis checks re-run? Looks like they failed due to issues in the main branch at the time.

Thanks.

bogdan-iancu:
@rrb3942 , I gave to reject this patch as it is too hackish - let's continue the discussion for a fix on #1073

PR #1145, opened 2017-06-20
Parameter "use_domain" not imported from usrloc

attermann:
Variable reg_use_domain is initialized to 0, but not imported from usrloc module as eluded in docs.  Added import of "use_domain" parameter from usrloc module to initialize reg_use_domain (taken from registrar module).


attermann:
Sorry about that pull request, it was sloppy.  Will resubmit the correct one.

PR #1147, opened 2017-06-21
Mid-registrar skipping appending of proxy ip:port to AOR usrloc use_domain is enabled

attermann:
In aor-throttling mode, when usrloc use_domain is enabled the aor already includes the domain so there is no need to append the proxy ip:port to the aor in the Contact header that is relayed to the main registrar.

### DIFF
--- a/modules/mid_registrar/save.c
+++ b/modules/mid_registrar/save.c
@@ -153,13 +153,20 @@ static int trim_to_single_contact(struct sip_msg *msg, str *aor)
 		return -1;
 	}
 
-	len1 = sprintf(buf, "<sip:%.*s@%s:%s>",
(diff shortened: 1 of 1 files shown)


liviuchircu:
LGTM, I should have spotted this in PR #1146 ...

attermann:
@liviuchircu There still appear to be some issues with the registration relayed to the main registrar with use_domain in contact-throttling mode.  I am just not familiar enough with how it all works yet to feel comfortable changing it.  It seems like it should also use the aor from usrloc which includes the domain (instead of proxy ip:port), but then again maybe just having the "rid" tag to associate it to the registration on the proxy is enough?  What do you think?

liviuchircu:
I will update/redo the testing scenarios and will let you know my findings asap

PR #1159, opened 2017-07-13
cgrates: add DisconnectCause for CDRs

bradleyjokinen:
This adds the DisconnectCause to the CDR data sent to CGRateS with the value set to the termination cause of the dialog.

### DIFF
--- a/modules/cgrates/cgrates_acc.c
+++ b/modules/cgrates/cgrates_acc.c
@@ -478,6 +478,7 @@ static json_object *cgr_get_cdr_acc_msg(struct sip_msg *msg,
 	char int2str_buf[INT2STR_MAX_LEN + 1];
 	static str cmd = str_init("SMGenericV1.ProcessCDR");
 	struct cgr_acc_sess *si = (struct cgr_acc_sess *)s->acc_info;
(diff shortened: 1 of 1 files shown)


razvancrainea:
Thanks for the patch, @bradleyjokinen! I was wondering though if the `DisconnectCause` field is used internally by cgrates? Because if it is not, I'd rather keep it out of the fields that are sent mandatory, and make it optional via script; the same behavior could be achieved before by doing in script:
```
$cdr(DisconnectCause) = $DLG_end_reason;
```

bradleyjokinen:
The `DisconnectCause` field is one of the primary fields for CDRs and gets used by some other CGRateS components. I initially tried setting the value in the script but I ran into some issues with accessing the CGRateS context at the time the dialog termination reason was available.

DigiDaz:
I am only seeing a DisconnectCause for billable calls, it seems that if BUSY or CANCEL etc, then {"DisconnectCause":""} seems to be sent to cgrates.

razvancrainea:
@DigiDaz please open a bug report for this. Thanks for spotting the issue!

Issue #1135, opened 2017-06-02
Error message when user location/dialog is stored in mongodb - Opensips 2.3


jalung:
[24348] CRITICAL:usrloc:dbrow2info: bad contact id
[24348] ERROR:usrloc:db_load_urecord: skipping record for 3333333333@domain.net in table location

Testing Opensips 2.3 with the cachedb_mongodb module, the above error appears when a UA's AOR is saved on registration. The AOR's are still saved however. Calls can be made and it appears to function normally.

The below errors also appear when dialogs are stored in mongo. They seem to appear at the BYE using the accounting module when calls are between UA's.

[24353] ERROR:cachedb_mongodb:mongo_db_update_trans: Failed to run query. Err = 0, 0 , 0
[24353] ERROR:cachedb_mongodb:mongo_db_update_trans: Fetched ERR key [shards]. Val = May 29 03:32:03 [24353] ERROR:cachedb_mongodb:mongo_db_update_trans: Fetched ERR key [shardRawGLE]. Val = May 29 03:32:03 [24353] ERROR:cachedb_mongodb:mongo_db_update_trans: Fetched ERR key [err]. Val = May 29 03:32:03 [24353] ERROR:cachedb_mongodb:mongo_db_update_trans: Fetched ERR key [ok]. Val = May 29 03:32:03 [24353] ERROR:dialog:update_dialog_dbinfo: could not update database info

Seems to function here normally too. Because the dialogs are being deleted at end of call.

FYI:
Documentation says to use mongo-c-driver 0.6 but that does not compile with 2.3 sips. 2.3 compiles with the 1.x versions of the driver but ultimately the cachedb_module is Dead On Arrival (at least for Mongo 2.6). To test a cachedb_mongodb module on sips 2.3, an older copy on the cached_mongodb module taken from sips 2.1.5 was compiled in and of course using the 0.6 driver.

Opensips 2.3
mongo-c-driver 0.6
MongoDB cluster 2.6.12

liviuchircu:
First of all, I have just updated the _cachedb_mongodb_ docs for OpenSIPS 2.3+, to reflect the driver update. What do you mean by "dead on arrival"? The new module should work with 2.6 Mongo servers, as per the driver docs, so we should start investigating that, rather than copy-pasting the old module.

jalung:
[...]

PR #1163, opened 2017-07-20
OpenSips cachedb_mongodb module et al patch

jalung:
#1135 Proposing this patch to resolve the issues encountered when using db_cachedb , cachedb_mongodb, usrloc, registrar and dialog modules together.
Versions Used: Opensips 2.3.1 , mongo-c-driver 1.6.3 , MongoDB 3.2.15

[opensips-2.3.1.txt](https://github.com/OpenSIPS/opensips/files/1160867/opensips-2.3.1.txt)


bogdan-iancu:
@jalung , your PR is wrong as it is based on master code, but merging into 2.3 . Please redo the PR against master.

liviuchircu:
@jalung: I have to say, this patch is an impressive work, showing just how complex OpenSIPS testing can get. After having replicated most of the bugs you suggested myself, here is my feedback:

1. Excellent catch! However, it seems like `mongoc_init()` is meant to be only called once (since it's using `pthread_once()` under the hood), so it belongs to `mod_init()`, rather than `mongo_new_connection()`. The underlying crash due to the un-initialized library is solved in both cases, nevertheless.

2. Even if you're lucky (or hard-working) enough that your system's libmongoc is compiled with `--disable-automatic-init-and-cleanup`, your script can still fail the startup phase if `tls_mgm` is loaded _before_ `cachedb_mongodb`. I suggest we add a `DEP_SILENT` module dependency, `cachedb_mongodb` -> `tls_mgm` to alleviate this.

3. This is the only item on your list that I cannot comprehend. My system worked well with the `pkg_malloc()` in place, and started leaking memory when I removed it - quite obvious from the logic too. Please give some more details on how it can misbehave or some backtrace, and we'll go from there.

4. Agree.

5. Thanks for catching this one! I suggest we refactor the fixed code to use a do-while loop, in order to avoid copy pasting that block of code.

6. Regarding `kvo_to_bson()`, it seems like I used `doc` all over the middle switch statement, instead of the obvious `child` object. Seemed to alleviate all crashes for me. Please let me know if your rewrite of `kvo_to_bson()` is actually meant to fix additional problems.

7. Agree.

8. Agree.

9. Although your `contact_id` fix helped a bit with `DB_ONLY` mode, an additional bug still seems to be lurking in there. To reproduce it: In DB_ONLY mode, unreg-dereg a single device a few times, so as to buffer up some deletion events. Soon enough, after the "flush" limit is hit, your device won't be able to register anymore (DB stays clean), as the system first inserts it, and then does a silly "flush" of some weird pending contact deletion queries, thus immediately removing what was just inserted:

```
             Aug  2 00:42:54 [15478] DBG:cachedb_mongodb:mongo_db_insert_trans: insert doc:
            { "contact_id" : 28077128926887936, "username" : "liviu", "contact" : "sip:liviu@192.168.0.101:48818;rinstance=1cb130aa06c2ac98;transport=UDP", "expires" : { "$date" : 1501623834000 }, "q" : -1, "callid" : "9dAnW_S54Sh0tlhB-Co5Tg..", "cseq" : 1, "flags" : 0, "cflags" : "", "user_agent" : "Zoiper rd82a609", "received" : null, "path" : null, "socket" : "udp:192.168.0.104:5060", "methods" : 5951, "last_modified" : { "$date" : 1501623774000 }, "sip_instance" : null, "attr" : null }
            Aug  2 00:42:54 [15478] DBG:cachedb_mongodb:mongo_db_delete_trans: remove doc:
            { "contact_id" : 28077128926887936, "contact_id" : 28077128926887936 }

```

Please re-open a proper Pull Request either from `2.3 -> 2.3` or `master -> master`, and feel free to include as many of my suggested changes as you'd like. FWIW, I'm willing to do all of them on top of your new PR :)

PR #1164, opened 2017-07-20
db_mysql: fix build against MariaDB 10.2

caringi:
Fixes the following build error while building against MariaDB 10.2
library:

```
my_con.c: In function 'db_mysql_connect':
my_con.c:68:10: error: 'MYSQL {aka struct st_mysql}' has no member named
'reconnect'
ptr->con->reconnect = 0;
          ^~
```

In this new version of the library the reconnect field was moved to
another struct. Setting it through API seems to be a better approach.

### DIFF
--- a/modules/db_mysql/my_con.c
+++ b/modules/db_mysql/my_con.c
@@ -34,6 +34,7 @@ extern struct tls_domain *tls_dom;
 
 int db_mysql_connect(struct my_con* ptr)
 {
+	my_bool reconnect = 0;
 	/* if connection already in use, close it first*/
 	if (ptr->init)
 		mysql_close(ptr->con);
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @caringi . Possible a stupid question, but how backward compatible is this mysql_options + MYSQL_OPT_RECONNECT ?

caringi:
@bogdan-iancu after a fast search on google I've found this: 

"introduced in 5.0.13"
https://lists.kamailio.org/pipermail/devel/2006-January/001651.html

So, I'd say that this option is at least 10 years old.

I've also found this PR for perl5-dbi: https://github.com/perl5-dbi/DBD-mysql/pull/133/files

Where someone suggests this #ifdefs:

https://github.com/perl5-dbi/DBD-mysql/pull/133/files#r121291081

bogdan-iancu:
@caringi , maybe the safest way will be to go for the #ifdef option. Do you want to rebuild the PR or should I do the change ?

caringi:
@bogdan-iancu done. I've amended the commit in this same PR.

bogdan-iancu:
Many thanks @caringi !! I will also do some backporting to the active versions too

PR #1170, opened 2017-07-31
Added functions to sipmsgops and support to handle Provisional Responses in b2b_entities.

fsantulli:
…remove_body, include_hf_value, exclude_hf_value, hf_value_existx


fsantulli:
I suppose to have added without note the original sipmsgops.c. Should i have to resubmit the request?

bogdan-iancu:
Imported from #1180 ( by @fsantulli ):

I've improved the ability to handle, forcing or striping the 100rel tag from calls in b2b top hiding scenario.
If a call arrives and the invite has the 100rel request, the b2b_entities module takes track of it. On handling the invite, when the remote party ring backs, if the caller required the provisional responses and they do not come back from the callee, the module adds itself the require header and a self generated rseq number.
There are also two exported dialplan functions that on the invite side, suppress (answering and removing) or forces (generating if not present) the Require and the RSeq headers.

Database structure has been updated adding 4 flags: caller and callee RSeq numbers and prov_resp flags on both dialog and leg.

In the commits have also been added some useful functions to sipmsgops.

Documentation has been updated as scripts.

Please merge the pull request if you can.

There is still a bug in the b2b module related to cseq and rseq numbers handled as signed integers so just up to the value of 32767 and not as rfc states in a range of 2**31 (unsigned long) number.
I didn't yet fixed the issue as the changes involves core functions to be implemented str2ulong and ulong2str.

bogdan-iancu:
Hi @fsantulli , could you split this PR and keep here only the b2b related part, as the functions for checking/removing/adding options to a list-header were committed (new code) minutes ago. See bbae925fe602bbbc813db659c7814aacc8d6a542 .

PR #1186, opened 2017-08-30
Fixes for mid-registrar memory corruption

attermann:
Fixes for memory corruption due to freeing of shared memory associated with mid_reg_info structure multiple times.


attermann:
Neglected to remove some debugging code.

PR #1187, opened 2017-08-30
Fixes for mid-registrar memory corruption

attermann:
Fixes for memory corruption due to freeing of shared memory associated with mid_reg_info structures multiple times.

### DIFF
--- a/modules/mid_registrar/save.c
+++ b/modules/mid_registrar/save.c
@@ -915,7 +915,7 @@ update_usrloc:
 			LM_DBG("UPDATING .....\n");
 			if (reg_mode != MID_REG_MIRROR) {
 				cti = c->attached_data[ucontact_data_idx];
-				cti ->expires_out = e_out;
+				cti->expires_out = e_out;
 				set_ct(cti);
 			}
 
--- a/modules/mid_registrar/save.c
+++ b/modules/mid_registrar/save.c
@@ -914,8 +914,9 @@ update_usrloc:
 
 			LM_DBG("UPDATING .....\n");
 			if (reg_mode != MID_REG_MIRROR) {
-				mri->expires_out = e_out;
-				set_ct(mri);
+				cti = c->attached_data[ucontact_data_idx];
+				cti ->expires_out = e_out;
+				set_ct(cti);
 			}
 
 			if (ul_api.update_ucontact( r, c, ci, 0) < 0) {
@@ -1136,9 +1137,10 @@ update_usrloc:
 
 			LM_DBG("UPDATING .....\n");
 			if (reg_mode != MID_REG_MIRROR) {
-				mri->expires = e;
-				mri->expires_out = e_out;
-				set_ct(mri);
+				cti = c->attached_data[ucontact_data_idx];
+				cti->expires = e;
+				cti->expires_out = e_out;
+				set_ct(cti);
 			}
 
 			if (ul_api.update_ucontact( r, c, ci, 0) < 0) {
@@ -1333,6 +1335,12 @@ static inline int star(udomain_t* _d, struct mid_reg_info *mri,
 }
 
 
+void mid_reg_trans_deleted(struct cell *t, int type, struct tmcb_params *params)
+{
+	struct mid_reg_info *mri = *(struct mid_reg_info **)(params->param);
+	mri_free(mri);
+}
+
 void mid_reg_resp_in(struct cell *t, int type, struct tmcb_params *params)
 {
 	struct mid_reg_info *mri = *(struct mid_reg_info **)(params->param);
@@ -1393,7 +1401,7 @@ void mid_reg_resp_in(struct cell *t, int type, struct tmcb_params *params)
 	LM_DBG("RESPONSE FORWARDED TO caller!\n");
 
 out_free:
-	mri_free(mri);
+	LM_DBG("not yet freeing mri: %p\n", mri);
 }
 
 /* !! retcodes: 1 or -1 !! */
@@ -1426,6 +1434,7 @@ static int prepare_forward(struct sip_msg *msg, udomain_t *ud,
 
 	if (parse_from_header(msg) != 0) {
 		LM_ERR("failed to parse From hf\n");
+		mri_free(mri);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


liviuchircu:
Unfortunately, it's still crashing on my setup, see below:

```
Core was generated by `./opensips -m64 -M16 -f etc/opensips-mid-reg.cfg -w .'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa65c0c113c in calc_contact_expires (_m=0x7fa65ea6de08, _ep=0x492d6c6c61430a0d, _e=0x7ffe285a1d80, _sctx=0x0) at save.c:86
86      if (!_ep || !_ep->body.len) {
(gdb) bt
#0  0x00007fa65c0c113c in calc_contact_expires (_m=0x7fa65ea6de08, _ep=0x492d6c6c61430a0d, _e=0x7ffe285a1d80, _sctx=0x0) at save.c:86
#1  0x00007fa65c0c5d65 in save_restore_req_contacts (req=0x7fa65ea6de08, rpl=0x7fa6625a9680, mri=0x7fa65ea74ef8, _a=0x7fa65ea74f68, r=0x7fa65ea758a8) at save.c:1049
#2  0x00007fa65c0c810a in mid_reg_resp_in (t=0x7fa65ea6c360, type=2, params=0x7ffe285a21b0) at save.c:1393
#3  0x00007fa65d8ad879 in run_trans_callbacks (type=2, trans=0x7fa65ea6c360, req=0x7fa65ea6de08, rpl=0x7fa6625a9680, code=200) at t_hooks.c:209
#4  0x00007fa65d878908 in t_reply_matching (p_msg=0x7fa6625a9680, p_branch=0x7ffe285a2330) at t_lookup.c:843
#5  0x00007fa65d878e2f in t_check (p_msg=0x7fa6625a9680, param_branch=0x7ffe285a2330) at t_lookup.c:918
#6  0x00007fa65d85b6cc in reply_received (p_msg=0x7fa6625a9680) at t_reply.c:1416
#7  0x000000000046c308 in forward_reply (msg=0x7fa6625a9680) at forward.c:495
#8  0x0000000000455127 in receive_msg (
    buf=0x8628e0 <buf> "SIP/2.0 200 OK\r\nVia: SIP/2.0/UDP 10.0.0.10:5060;branch=z9hG4bKef1.d62d21a3.0\r\nVia: SIP/2.0/UDP 10.0.0.10:54229;received=10.0.0.10;rport=54229;branch=z9hG4bKPj7e20b7b0-8d5b-4c9d-9e1f-24c781521382\r\nFrom"..., len=566, rcv_info=0x7ffe285a2520, 
    existing_context=0x0) at receive.c:257
#9  0x0000000000581dac in udp_read_req (si=0x7fa662577c40, bytes_read=0x7ffe285a25d0) at net/proto_udp/proto_udp.c:192
#10 0x00000000005617b8 in handle_io (fm=0x7fa66259b8e0, idx=0, event_type=1) at net/net_udp.c:259
#11 0x0000000000560050 in io_wait_loop_epoll (h=0x885c40 <_worker_io>, t=1, repeat=0) at net/../io_wait_loop.h:280
#12 0x0000000000562154 in udp_start_processes (chd_rank=0x84eda0 <chd_rank>, startup_done=0x0) at net/net_udp.c:389
#13 0x00000000004c2b77 in main_loop () at main.c:677
#14 0x00000000004c5ce6 in main (argc=7, argv=0x7ffe285a2938) at main.c:1283
(gdb) f 1
#1  0x00007fa65c0c5d65 in save_restore_req_contacts (req=0x7fa65ea6de08, rpl=0x7fa6625a9680, mri=0x7fa65ea74ef8, _a=0x7fa65ea74f68, r=0x7fa65ea758a8) at save.c:1049
1049            calc_contact_expires(req, __c->expires, &e, NULL);
(gdb) list
1044    
1045        for (__c = get_first_contact(req); __c; __c = __cn) {
1046            __cn = get_next_contact(__c);
1047    
1048            /* calculate expires */
1049            calc_contact_expires(req, __c->expires, &e, NULL);
1050            if (_c == NULL) {
1051                if (e != 0)
1052                    LM_ERR("200 OK from main registrar is missing Contact '%.*s'\n",
1053                           __c->uri.len, __c->uri.s);
(gdb) p __c->expires
$1 = (param_t *) 0x492d6c6c61430a0d
(gdb) x/s &__c->expires
0x7fa65ea72945: "\r\nCall-ID: b26fd6a6-524f-44d4-acab-ac9c7b20b29f\r\nCSeq: 1 REGISTER\r\nExpires: 30\r\nSupported: gruu\r\nUser-Agent: Blink 3.0.3 (Linux)\r\nContent-Length:  0\r\n\r\n"
```
This is quite a subtle bug, which has also present in the mainline `registrar` code since forever, but only on a 200 OK `save()` from script. Inside the mid_registrar's `mid_reg_resp_in()` and the registrar's `save_aux()` function, respectively, the initial SIP request is actually SHM-ized. Despite of this, both registrar and mid_registrar are attaching all sort of PKG-parsed structures to it, and working off of them. While this works, if two processes are running this code concurrently, they will each place their own PKG parsed pointers into the common SHM-ized initial request, easily leading to a crash.

To conclude the above, we either use the "save/skip" approach to retransmissions (see below), or protect the entire `mid_reg_resp_in()` callback with a big lock, if we want retransmit behavior to work as "save/update". I say we drop the latter idea for now, as it will almost serialize the REGISTER processing between the workers.

I will just add a shared `rw_lock_t` to protect the `*params->part = NULL` and `if (!mri)` code. While it is a sync section, it only gets called once per registration, and the majority of processing time is spent in other parts, so we are not introducing some sort of bottleneck.

attermann:
Just to be clear, this fix was not meant to resolve the crashes that we're seeing presently in `save_restore_req_contacts`.  I was having many more varied crashes before (see #1178) which this fix addresses.

Regarding these present crashes, I agree that "save/skip" is the best approach for handling retransmissions (can't even think of a reason to perform a "save/update" off-hand).  I modified your prior fix of setting `params->param` to null by placing it immediately after the check for null mri...

```
+       if (!mri) {
+               LM_INFO("SIP reply retransmission -> exit\n");
+               return;
+       }
+       *params->param = NULL; /* do not run this callback multiple times! */
```

This has stabilized opensips in my production environment.  I think that a sync section around this code would be a solid fix until something more elegant is available (perhaps from the tm module).

In the future, I think some way of detecting the first final-response vs. retransmissions in the tm module would be a valuable feature.  Either a new callback type that is only called once on final-response (TMCB_RESPONSE_FINAL?), or some indication in the TMCB_RESPONSE_IN callback that a given response is a retransmission.

liviuchircu:
Thanks for reviewing and correctly fixing commit 496e2dcc274 (I won't even begin commenting with regards to what I did there...).

I'll get everything backported for the upcoming `2.3.2` in the next minutes.

attermann:
@liviuchircu Do you plan to merge these commits that address memory corruption due to multiple frees of `mri` shared memory?  I did pretty extensive debugging of allocations and frees of `mri` memory to determine the places where pointers were being copied instead of duplicated, freed multiple times, etc.  I am keeping these changes in my fork for now because it seems to be stable, but I'd really like to know what you think about them.  Do you think they're unnecessary with the latest fixes from commit 496e2dc?

liviuchircu:
Yes, I definitely gave those some thought. If we go for "save/update", they are **mandatory**. Otherwise, the code works fine without them in a "save/skip" scenario, hence the reason why I didn't include them in the diff.

attermann:
OK, thanks for the clarification!

PR #1188, opened 2017-08-31
Misleading header directories when came to cross-compile

was4444:


razvancrainea:
Thanks for providing this fix, but unfortunately OpenSIPS 1.11 is no longer supported - can you please update this patch for master, OpenSIPS 2.3 and OpenSIPS 2.2 and open a new pull request?

PR #1191, opened 2017-09-08
fix memory leak in cpl-c write_to_db

etamme:
In write_to_db, the underlying db driver alloc's memory for a query, but it is never freed before return.


bogdan-iancu:
Thank you @etamme , I did also backports to 2.3 and 2.2

PR #1207, opened 2017-11-03
fix and optimize oracle fetch

dronord:


dronord:
Now in code OCI_FETCH_NEXT and then OCI_FETCH_PRIOR - only one first row will be taken. In any number of rows.
Here need OCI_FETCH_LAST instead NEXT:

static int get_rows(ora_con_t* con, db_res_t* _r, OCIStmt* _c, dmap_t* _d)
{
  // go to last row
  status = OCIStmtFetch2(_c, con->errhp, 1, OCI_FETCH_LAST, 0,
		OCI_DEFAULT);

  // current row number is count of rows in result
  status = OCIAttrGet(_c, OCI_HTYPE_STMT, &rcnt, NULL,
		OCI_ATTR_CURRENT_POSITION, con->errhp);

  // allocate memory for result
  db_allocate_rows(...);

  // get row data from last to first
 while (1) {
    do status = OCIStmtFetch2(_c, con->errhp, 1, OCI_FETCH_PRIOR, 0,
			OCI_DEFAULT);
  }
}

dronord:
OCI_ATTR_PREFETCH_ROWS
From Oracle documentation:
"In order to minimize server round trips and maximize the performance of applications, the OCI can prefetch result set rows when executing a query."
On 2000 SIP registrations in db, OpenSIPS starts 10 seconds.
When prefetch used, server starts immediately.

PR #1209, opened 2017-11-03
db_ora: null-terminating string is more safely, most modules use VAL_…

dronord:
…STRING, not VAL_STR.


razvancrainea:
Is there a bug, or a particular reason you are committing this?
I am asking because although `DB_STRING` seems more intuitive and easy to use, some db backends can't use it (see dbtext), therefore we still need to implement `DB_STR` in most of the places.

dronord:
Not a bug.
load_balancer, drouting, others not working without it, because
`check_val( ..., DB_STRING, ...);`
Yes, need to patch them.

PR #1230, opened 2017-12-11
modules/json: Properly detect json-c version and adaptions for v0.13

besser82:


besser82:
THe checks are failing because JSON_LIB_VERSION gets defined in an unpredictable way…

* If the json-c version is 0.11.99 ---> JSON_LIB_VERSION == 1199
* If the json-c version is 0.13 ---> JSON_LIB_VERSION == 13

That's somehow fishy and should not happen this way…

razvancrainea:
Thanks for digging into this!
You are right, the `JSON_LIB_VERSION` might not be consistent. This was a workaround for the times when the module was written and most of the distros were shipping 0.08 and 0.10 versions.

Nowadays most of the distros are shipping 0.11+ versions, but I wouldn't deprecate the older code yet.
I would propose the following selection:
* if the `/usr/include/json-c/json_c_version.h` exists, don't do anything in `Makefile`, and in the code do something like:
```
#ifndef JSON_C_VERSION_NUM
#include "json_c_version.h"
#endif
```
* if the version does not exist, fix the `Makefile` to use `pkg-config` and define the `JSON_C_VERSION_NUM` (instead of `JSON_LIB_VERSION`), but make sure it has the expected format
* if `pkg-config` is not found, assume the most commonly deployed version - 0.11.x

What do you think about the approach above? Do you see any flaws with it? If so, I am happy to discuss. If there are no issues, I can adjust the current files to make this work. Or if you are willing to do it, I'd me more than happy to review and merge your changes.

Best regards,
Răzvan

besser82:
@razvancrainea The rebased commit should address all versioning szenarios properly.

PR #1239, opened 2017-12-21
Oracle NUMBER not NUM in application

dronord:
From Oracle documentation https://docs.oracle.com/cd/B19306_01/appdev.102/b14250/oci03typ.htm
"NUMBER
You should not need to use NUMBER as an external datatype. If you do use it, Oracle returns numeric values in its internal 21-byte binary format and will expect this format on input."


dronord:
In PHP docs no SQLT_NUM!
https://www.php.net/manual/ru/function.oci-bind-by-name.php

PR #1256, opened 2018-01-19
replace /usr/local/lib/opensips to /usr/local/lib64/opensips in all files

Jeffrey2019:
Since most new OpenSIPS is running on 64 bit OS, is it reasonable to replace '/usr/local/lib/opensips' to '/usr/local/lib64/opensips' throughout the repository?

It will avoid update default config file generated every-time.


liviuchircu:
Resolution in #1261.  Closing.

PR #1345, opened 2018-04-19
Two minior issue which will stop compiling in unbuntu 18.04 with gcc7.3.0

wuhanck:
I want to use opensips in ubuntu 18.04 for future usage.
I found that the wrong usage of json_object_object_get. 
It is deprecated and even it is not, there is a BUG which jobj is nullified and use it in DEBUG info.

The other issue is that str_ip is not accually used and it block the comping too. (And some typo in the same file.)

I fix these two.

Also there is a sugguestion:
Can we make ip_addr2a A MACRO that allocate string on stack? 
It is not thread-safe.
I don't know whether it concerns.

### DIFF
--- a/modules/cgrates/cgrates_acc.c
+++ b/modules/cgrates/cgrates_acc.c
@@ -203,16 +203,16 @@ static int cgr_proc_start_acc_reply(struct cgr_conn *c, json_object *jobj,
 		return -1;
 
 	if (!cgre_compat_mode) {
+		json_object *jorig = jobj;
 		if (json_object_get_type(jobj) != json_type_object) {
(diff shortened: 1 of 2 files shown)


liviuchircu:
Good catches, @wuhanck! Both fixes look good to me.

Regarding the `ip_addr2a()` change, is there a real need for this? Currently, it's optimized for a single print & use style of usage. If you need to print multiple IPs before using them, we can update it to work similarly to `pv_printf_s()` in a different Pull Request / issue.

wuhanck:
I just begin to compile opensips, so I don't have any saying in ip_addr2a issue.
I have seen that in modules/acc/acc_extra.c some variable refer _ip_addr_A_buff directly.
There is  long story I think :).

Issue #1364, opened 2018-05-17
rtcp-mux-require missing from opensips


ryan-esty:
rtpengine now accepts rtcp-mux-require but opensips doesn't. WebRTC at least with chrome is moving to this scenario and it would be useful to support this.

ryan-esty:
I might not have everything covered. In rtpengine.c function parse_flags I added a rtcp-mux-require to case 16. After compiling and moving the module over rtpengine no longer has a problem with rtcp-mux-require. The only thing I'm wondering is if I have a rtcp-mux-require in the onreply shouldn't I see a rtcp-mux in the reply? This might be an rtpengine question but I'm trying to test this fix.

razvancrainea:
Can you share the code you've added? You have to set the `require` attribute to the `rtcp-mux` object in the bson, not directly in the main bson. Check how it's done for `rtcp-mux-accept` or `rtcp-mux-offer`.

ryan-esty:
--- modules/rtpengine/rtpengine.c       (revision 18227)
+++ modules/rtpengine/rtpengine.c       (working copy)
@@ -1228,6 +1228,8 @@
                        case 16:
                                if (str_eq(&key, "UDP/TLS/RTP/SAVP"))
                                        ng_flags->transport = 0x104;
+                               else if (str_eq(&key, "rtcp-mux-require"))
+                                       BCHECK(bencode_list_add_string(ng_flags->rtcp_mux, "require"));
                                else
                                        break;
                                continue;

I forked it and did a commit so it might be easier to look at:
https://github.com/OpenSIPS/opensips/commit/fd1124a2cca4d61c1b6493a13b5911e8441c835f

razvancrainea:
Yes, that looks OK. Can you also make a pull request for it :)?

I think you should also have `rtcp-mux-offer` or `rtcp-mux-require` in onreply_route to see it in the reply. Can you check this?

ryan-esty:
I haven't done much with github before. Hopefully you can use what I have if not I can try a pull if that is easier. I did run through all the different options. I would think accept would be the most likely candidate but I did do offer and require. Note that rtpengine had a comment saying that offer and probably it's cousin require shouldn't be in the reply they are only for the initial offers.

I'm not terminating in opensips though. In my scenario I'm using it as a gateway and I have a legacy SIP server that my phones are registered to. I'm not sure if that matters or not. I might need to ask the rtpengine people.

razvancrainea:
[...]

PR #1366, opened 2018-05-18
added rtcp-mux-require

ryan-esty:
I haven't been able to test this properly but I believe this is correct. The documentation is mostly taken from rtpengine. This is a fix for issue #1364.


bogdan-iancu:
Yes, it did, see the above commits on all branches up to 2.2

PR #1370, opened 2018-05-23
Rate Cacher module

vladpaiu:
Added a new module ( rate_cacher ) managing client and vendor ratesheets.
At startup, the module loads all the provisioned ratesheets that are assigned to all clients and all vendors.
Further on, the module exposes two types of functionalities :
* script-wise, it supports 
** filtering a list of vendors for a call coming from a specific client, based on the prices they offer for a specific destination, and a user desired profit margin, while keeping the initial provided user order ( see README )
** ordering a list of vendors for a call coming from a specific client, based on the prices they offer for a specific destination, and a user desired profit margin
* MI, it supports
** adding a new client
** assigning a new ratesheet or reloading the ratesheet of a client ( each client can have two ratesheets assigned, in order to support separate prices for separate offered qualities )
** querying for the price of a client on a specific destination
** removing a client
** adding a new vendor
** assigning a new ratesheet or reloading the ratesheet of a vendor
** querying for the price of a vendor on a specific destination
** removing a vendor


modules/rate_cacher/doc/rate_cacher_admin.xml
```diff
@@ -0,0 +1,636 @@
...
+		The  <emphasis>is_wholesale</emphasis> pseudo-var will contain either a 1 or a 0, depending on whether the call is wholesale or retail ( see client ratesheet provisioning ).
+		</para>
+		<para>
+		The <emphasis>vendors_csv</emphasis> pseudo-var contains a list of Vendors that need to be filtered based on the desired margin ( keep just those that match your desired percentage margin for this call )
+		</para>
+		<para>
+		The <emphasis>dialled_no</emphasis> pseudo-var contains the DNIS - the dialled number for the current call. It needs to be in E164 format, without the leading +
+		</para>
+		<para>
+		The <emphasis>desired_margin</emphasis> pseudo-var contains the minimum Integer margin that the script writer wants to achieve, based on the Client sell and Vendor buy prices. The formula used is : vendor_margin=(client_price - results[i])*100/client_price) . If the vendor_margin is higher than the desired_margin, then the Vendor is ok to use. The desired margin can be positive ( call will be profitable ) or negative ( the call will cause a loss ).
```
liviuchircu: 15. There are quite a few inaccuracies in this one sentence. For example:

* `The desired margin can be ... negative ( the call will cause a loss )`. This is not true, I can feed the function with `-3%` and still make profit from *all* my calls. More accurate would be: `the call may cause a loss`
* `The desired margin can be positive ... or negative`. What if I set it to zero? :)

I suggest:

```
The desired margin can be positive (call must be profitable), zero (call must not cause a loss) or negative (call may cause a loss).
```

modules/rate_cacher/doc/rate_cacher_admin.xml
```diff
@@ -0,0 +1,636 @@
...
+# to use based on the above call characteristics, and the desired margin
+# The order in $avp(carrierlist) does not matter, the vendors will be
+# ordered from most profitable to least profitable
+$avp(client_id)="testClient";
+$avp(is_ws)=1;  
+$avp(carrierlist)="testVendor,testVendor2";
+$avp(dnis)="40720018124";
+$avp(profit_margin)=0;
+
+if (cost_based_filtering("$avp(client_id)","$avp(is_ws)","$avp(carrierlist)","$avp(dnis)","$avp(profit_margin)","$avp(out_vendor_result)")) {
```
liviuchircu: 16. `if (cost_based_ordering(`

modules/rate_cacher/rate_cacher.c
```diff
@@ -0,0 +1,2195 @@
+#include <stdio.h>
```
liviuchircu: 17. Please add GPLv2 headers. Same for `rate_cacher.h`.

PR #1376, opened 2018-06-06
opensipsctl: Improve the "trap" command

liviuchircu:
* prevent "trap" from failing if the FIFO process is deadlocked
    * properly trap different OpenSIPS binaries/instances running on
      the same box

### DIFF
--- a/scripts/opensipsctl
+++ b/scripts/opensipsctl
@@ -2611,23 +2611,56 @@ opensips_trap() {
 		merr "'gdb' tool not found: set GDB variable to correct tool path"
 		exit
 	fi
-	DATE=`/bin/date +%Y%m%d_%H%M%S`
-	LOG_FILE=/tmp/gdb_opensips_$DATE
-	minfo "Trap file: $LOG_FILE"
-	$CTLCMD ps > $LOG_FILE
-	echo -n "Trapping OpenSIPS with gdb: "
-	PID_TIMESTAMP_VECTOR=`sed -e 's/.*PID=\([0-9]*\).*/\1/' $LOG_FILE`
-	minfo "$PID_TIMESTAMP_VECTOR"
-	for pid in $PID_TIMESTAMP_VECTOR
-	do
-		echo -n "."
-		PID=`echo $pid | cut -d '-' -f 1`
-		echo "" >> $LOG_FILE
-		echo "---start $PID -----------------------------------------------------" >> $LOG_FILE
-		$GDB opensips $PID -batch --eval-command="bt full" 2>&1 >> $LOG_FILE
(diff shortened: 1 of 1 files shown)


liviuchircu:
If we decide this solution looks good, I would even go as far as backporting it all the way down to `2.2`.

scripts/opensipsctl
```diff
@@ -2611,23 +2611,20 @@ opensips_trap() {
...
+	minfo "Trapping OpenSIPS with gdb: "
+	while read procline; do
+		BIN=$(echo "$procline" | grep -o "\<.*opensips\s" | awk '{print $NF}')
+		PID=$(echo "$procline" | awk '{print $2}')
+		echo -n "$PID "
+		echo >>$LOG_FILE
+		echo "---start $PID ($procline)" >>$LOG_FILE
+		$GDB $BIN $PID -batch --eval-command="bt full" 2>&1 >>$LOG_FILE
+		echo "---end $PID -------------------------------------------------------" >>$LOG_FILE
+	done < <(ps aux | grep '[o]pensips.*-f.*' | grep "$TRAP_FILTER")
```
razvancrainea: This will not work for instances started with the default config (i.e. without the `-f` parameter).

razvancrainea: Personally, I think the previous approach where the `opensipsctl ps` command was used is more flexible than this one. For example, if you have two opensips instances running and you only want to debug one, all you have to do is to set the right `OSIPS_FIFO` file and you're done.

In my opinion, the previous approach had only one limitation: if the FIFO MI engine is stuck, the trap tool cannot fetch the PIDs to execute the gdb dump. And in order to solve this, I would go towards a solution where we can specify the PIDs that we want to make trap for - specifically, if a  `$PID_TIMESTAMP_VECTOR` is specified in the CLI, use that one, otherwise get the PIDs using `opensipsctl ps`.

liviuchircu: That's why I threw the `$TRAP_FILTER` in there. It allows you to do something like: `TRAP_FILTER=opensips-backup.cfg opensipsctl trap`, and only trap instance #2.

razvancrainea: `TRAP_FILTER` still does not solve the issue where there is not `-f` specified at startup.
Again, personally I think using the `OSIPS_FIFO` approach is more elegant, and you only fallback to specifying the `$PID_TIMESTAMP_VECTOR` when you really have to.

bogdan-iancu:
It looks good, let's backport it !

bogdan-iancu:
BTW, what is missing here is an update of the "trap" description (as params) in the help listing

PR #1380, opened 2018-06-08
Bugfix/fix corefile generation

liviuchircu:
A set of patches which:

* improves `SIGKILL`-based shutdown code (do not skip final `cleanup()` on a `SIGKILL` broadcast)
* avoid truncated corefiles on OpenSIPS `2.4+` versions
* fix `systemd` service files, so `restart` commands don't deadlock OpenSIPS

### DIFF
--- a/config.h
+++ b/config.h
@@ -52,6 +52,15 @@
 
 #define CHILD_NO    8		/*!< default number of child processes started */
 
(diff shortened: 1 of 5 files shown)


bogdan-iancu:
@liviuchircu, please do the backport to 2.4 too.

PR #1394, opened 2018-06-15
Improve docs

lemenkov:
These two commits will switch more docs to UTF-8.


bogdan-iancu:
Thank you @lemenkov, this perfectly fits with all the work we do to improve docs as content and format.

PR #1397, opened 2018-06-19
revised timeout value information

spacetourist:
Updated tls_mgm documentation following discovery or a unit error in the settings (s->ms)

Note that Liviu also asks "How do they relate to "tcp_connect_timeout" (core param) and "tcp_send_timeout" (proto_tcp module)" so this information could be added for additional clarity.


liviuchircu:
Excellent work, thanks!

PR #1398, opened 2018-06-20
Number of improvements into the python module:

sobomax:

stuck on exit for 60 seconds.

### DIFF
--- a/modules/python/python_msgobj.c
+++ b/modules/python/python_msgobj.c
@@ -190,6 +190,24 @@ msg_getHeader(msgobject *self, PyObject *args)
     return PyString_FromStringAndSize(hbody->s, hbody->len);
 }
 
+static char *
+pkg_strdup(char *cp, const char *fl, const char *fn, unsigned int ln)
+{
+	char *rval;
+	int len;
+
+	len = strlen(cp) + 1;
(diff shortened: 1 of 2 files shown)


modules/python/python_msgobj.c
```diff
@@ -190,6 +190,24 @@ msg_getHeader(msgobject *self, PyObject *args)
...
+pkg_strdup(char *cp, const char *fl, const char *fn, unsigned int ln)
+{
+	char *rval;
+	int len;
+
+	len = strlen(cp) + 1;
+#if !defined(DBG_MALLOC) || 1
+	rval = pkg_malloc(len);
+#else
+	rval = pkg_malloc_dbg(len, fl, fn, ln);
```
bogdan-iancu: @sobomax , this function is not defined anywhere....Am I missing something ?

sobomax:
This is a separate piece of code that I wanted to post. The idea there is
to allow modules to tap into allocator debug facility, so that they can
provide specific location in case memory gets leaked. I intended to clean
that code a bit and post it as a separate pull request. For the time being
I can either leave it as is or strip down the pkg_malloc_dbg() branch along
with the #if block. Let me know which way you prefer.

Thanks!

-Max

sobomax:
Too old. GC.

PR #1399, opened 2018-06-20
Number of improvements into the python module

sobomax:

stuck on exit for 60 seconds.


sobomax:
Too old. GC.

PR #1401, opened 2018-06-20
Number of improvements into the python code

sobomax:

    stuck on exit for 60 seconds.

    o Merge from 2.3: improve exception handling by reporting particular function that
      has errored.


sobomax:
Too old. GC

PR #1412, opened 2018-06-26
ctl tools: Fix drouting help

willyrgf:


bogdan-iancu:
@willyrgf thanks for your note. Still, the example was wrong, will the listing of the cmd was correct - the correct command in `rmrule`

willyrgf:
I have tested in "version: opensips 2.2.6 (x86_64/freebsd)" and the example command is correct (rmgrule) and the document is wrong (rmrule), please check this.

bogdan-iancu:
Prior to 2.4, the name was indeed `rmgrule` (but was incorrect as it had a typo). This was fixed in 5740fff644f82c5115d0f831e579c62178436eb9 . So 2.4 and master have the correct command name of `rmrule`.
Your PR is against master, where the name of the command is `rmrule`, so I fixed the example. For versions prior to 2.4, the help must be correct indeed to `rmgrule` - I will take care of this.

bogdan-iancu:
Sorry, the fix (from `rmgrule` to `rmrule`) was backported to 2.3 and 2.2 latest versions  too, see https://github.com/OpenSIPS/opensips/blob/2.2/scripts/opensipsctl#L1576

PR #1413, opened 2018-06-27
pv_print_buf_size

NormB:
Make PV_PRINT_BUF_SIZE into a global parameter: pv_print_buf_size.

For example, in opensips.cfg:

pv_print_buf_size = 1024;

### DIFF
--- a/pvar.c
+++ b/pvar.c
@@ -4615,7 +4615,7 @@ void pv_value_destroy(pv_value_t *val)
 	memset(val, 0, sizeof(pv_value_t));
 }
 
-int		pv_print_buf_size = 1024;
+int	pv_print_buf_size = 1024;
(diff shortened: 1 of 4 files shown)


pvar.c
```diff
@@ -4624,14 +4624,14 @@ void pv_value_destroy(pv_value_t *val)
 int pv_printf_s(struct sip_msg* msg, pv_elem_p list, str *s)
 {
 	static int buf_itr = 0;
-	static char buf[PV_PRINT_BUF_NO][PV_PRINT_BUF_SIZE];
```
bogdan-iancu: `buf` *MUST* be static, otherwise you cannot use it after returning from the `pv_printf_s()` function. With your change, the function is totally broken, as the buffer will be lost on the stack after return.

PR #1416, opened 2018-06-28
{csv.value} vulnerability

### LINKED ISSUE #1220: {csv.value} vulnerability
The `{csv.value}` transformation will access invalid memory (beyond its search buffer) when given a well-crafted input as such:

```
$var(test) = "\"foo";
xlog("$(var(test){csv.value,0})\n");
```

```
Nov 23 15:16:05 [8707] ERROR:core:pv_printf: no more space for spec value [6][16272976]
Nov 23 15:16:05 [8707] ERROR:core:pv_printf: buffer overflow -- increase the buffer size from [4096]...
Nov 23 15:16:05 [8707] ERROR:core:xl_print_log: failed to resolve xlog variables!
Nov 23 15:16:05 [8707] ALERT:core:do_action: Cannot print xlog debug message
```

This bug is present in all OpenSIPS 1.8 and above versions. It may or may not also cause a segfault.

fgast:
fixes #1220

### DIFF
--- a/transformations.c
+++ b/transformations.c
@@ -1375,9 +1375,12 @@ static int parse_csv(str *s,csv_t **list)
 			aux = string+1;
 search:
 			/* find coresponding quote */
-			while (*aux != '\"') aux++;
+			while (*aux != '\"' && aux+1 < limit-1) aux++;
 			if ( *(aux+1) != '\"')
 			{
+				if (aux+1 >= limit-1 )
+					 aux  = strchr(string,',')-1;
+
 				/* end of current token, also skip the following comma */
 				len = aux-string+1;
 				if (init_csv(&t,string,len) < 0)
@@ -1396,9 +1399,22 @@ search:
 			else
 			{
 				/* quoted string inside token */
-				aux +=2;
-				/* keep searching for final double quote*/
-				goto search;
+				if (aux+2 < limit) {
+					/* keep searching for final double quote*/
+					aux +=2;
+					goto search;
+				} else {
+					len = aux-string+2;
+					if (init_csv(&t,string,len) < 0)
+					{
+						LM_ERR("no more memory");
+						goto error;
+					}
+
+					/* should be end of string ! */
+					if (last) { last->next = t;} else {*list = t;}
+					return 0;
+				}
 			}
 		}
 		else


liviuchircu:
Thank you for the fix, @fgast! This is a good attempt, but it's still not correct. For example, I can now feed it with:

```
    $var(a) = "\"";
    xlog("XXXXXX: $(var(a){csv.value,0})\n");
```

... and achieve overflow through a newly introduced bug:

```
Jul  6 16:17:33 [1103] ERROR:core:pv_printf: no more space for spec value [8][38448504]
Jul  6 16:17:33 [1103] ERROR:core:pv_printf: buffer overflow -- increase the buffer size from [4096]...
Jul  6 16:17:33 [1103] ERROR:core:xlog_1: global print buffer too small, increase 'xlog_buf_size'
Jul  6 16:17:33 [1103] ERROR:core:do_action: error while printing xlog message
```

Regarding the overall approach behind this fix, I suggest we take a different route, one that's healthier for the codebase.  I have already started a simple CSV parser under `lib/csv.[ch]`.  It does not have full RFC 4180 support, but it is capable of eating a lot of whitespace and is flexible enough to distinguish between and properly free both SHM and PKG buffers using the same API.  The current way to use it is:

`_parse_csv_record(my_str_pointer, CSV_SIMPLE);`

I suggest we do the following:

* wrap up the TODOs in `lib/csv.c`, specifically the RFC 4180 double quoting part. I cannot remember if RFC 4180 allows us to eat leading/trailing whitespace between the CSV values. I would assume the answer to be *no*.
* flip around the `CSV_SIMPLE` flag I added.  It should become the default behavior of `parse_csv_record()`, while a new flag to `_parse_csv_record()` (e.g. `CSV_RFC_4180`) should allow the programmer to enable the standard (RFC-based) parsing logic.
* lastly, we just hook this code in `transformations.c`, both avoiding duplicated code, and fixing the current issues lurking behind the `{csv.value}` transformation.

Cheers,

fgast:
Thanks for the feedback @liviuchircu. 
I will check the code and prepare the changes you suggested.

bogdan-iancu:
@fgast , @liviuchircu , any progress here ?

PR #1421, opened 2018-07-06
* fixed msg payload content to be string rather than binary data

pasanmdev:
According to homer db schema data type should be string for payload msg
other than sip proto

http://lists.opensips.org/pipermail/users/2018-July/039627.html

### DIFF
--- a/modules/sipcapture/sipcapture.c
+++ b/modules/sipcapture/sipcapture.c
@@ -5070,7 +5070,7 @@ static int report_capture(struct sip_msg* msg, str* table, str* cor_id,
 		db_vals[10].val.str_val = capture_node;
 	}
 
-	db_vals[11].type = DB_BLOB;
+	db_vals[11].type = DB_STR;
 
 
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @pasanmdev . Do you have a link to the DB schema for the reports? maybe @adubovikov can confirm this.

adubovikov:
For me it's ok. just worry about SIP-T / SIP -I binary body. Did you test it ?

pasanmdev:
@bogdan-iancu  db schema is in 
https://github.com/sipcapture/homer-api/blob/master/sql/pgsql/schema_data.sql
if you check the msg column for all tables except for sip_capturer table is varchar type means it expect to store str instead of binary data.
@adubovikov This will only effect report_capture function, not the sip_capturer function. So sip messages will continue to store in binary format.

adubovikov:
sorry! In this case for me it's ok.

bogdan-iancu:
Thank you both !!!

PR #1448, opened 2018-08-22
Change topology hiding Contact header to fit RFC 3261

### LINKED ISSUE #1447: Topology hiding without dialog doesn't match RFC 3261
When using topology hiding without dialog module, it encodes all the topology headers in Contact header. It creates header like this for example:
```
Contact: <sip:[0:0:0:0:0:0:0:1]:8061;transport=wss;thinfo="U2h5YhUORg5uUBNZW1dIUAQadlZjRlFWUV0DBHJXRGgWBhZdBwxxSWRaTVBWXxgGcF9pXVNUVg==">
```
https://tools.ietf.org/html/rfc3261#section-25.1
Between LAQUOT and RAQUOT there is addr-spec. And according to RFC it can't contain  quoted-string, or even DQUOTE symbol. And parameter value can't contain EQUAL sign.

vasilevalex:
Fixes #1447 
Please review.

### DIFF
--- a/modules/topology_hiding/topo_hiding_logic.c
+++ b/modules/topology_hiding/topo_hiding_logic.c
@@ -1196,27 +1196,20 @@ static int dlg_th_decode_callid(struct sip_msg *msg)
 	struct lump *del;
 	str new_callid;
 	int i,max_size;
-	char *p;
 
 	if (msg->callid == NULL) {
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thank you @vasilevalex , this is a very good fix and patch, I will apply and port it asap.

PR #1469, opened 2018-09-18
New transformation {s.time,format} + drouting dr_is_gw() 'r' flag to check proto

nikbyte:
1. Converts unix time into readable format using strftime().

Usage example:
$var(time) = 1537277252;
xlog("$(var(time){s.time,%Y-%m-%d %H:%M:%S})\r\n");

Thanks Maxim Bondarenko (@mabondarenko) for transformation code.

.

2. Added 'r' flag to drouting dr_is_gw() function which allows to check protocol.


bogdan-iancu:
@nikbyte, please do one PR per topic - mixing different unrelated topics in the same PR is not a good practice.
The DR change seems a good idea. When generating  the PR, please do rebuild the README (just the xml file) as it is automatically rebuild.
For the time, I'm chewing the idea of creating a new "time" class for transformation, maybe to have also some string to unixtimestamp conversion

PR #1475, opened 2018-09-24
Remove outdated and misleading ChangeLog

lemenkov:
These ChangeLog entries are incomplete and outdated. We shouldn't keep
it. If someone is interested in specific changes then it's better to use
git for that.


lemenkov:
Ping :)

bogdan-iancu:
Hi @lemenkov , indeed, the ChangeLog for master is not updated in time. The file is updated only at release time, so it makes sense to exist only in the "released" branches, but not in master (where the git log does a better job, indeed :) ).

PR #1485, opened 2018-10-05
Drouting module probing socket parameter

### LINKED ISSUE #1484: Drouting module socket parameter for probing gateways
When OpenSIPS has serveral listening sockets, it is impossible to specify which one must be used to probe status of gateways. It will use the first one from configuration file.
But for example, in cluster configuration with Virtual IP moving between servers this causes a lot of errors in logs.

vasilevalex:
Fixes #1484 
I made it exactly like in dispatcher module.
Please review.

### DIFF
--- a/modules/drouting/doc/drouting_admin.xml
+++ b/modules/drouting/doc/drouting_admin.xml
@@ -912,6 +912,35 @@ modparam("drouting", "probing_reply_codes", "501, 403")
 		</example>
 	</section>
 
+	<section id="param_probing_socket" xreflabel="probing_socket">
+		<title><varname>probing_socket</varname> (string)</title>
+		<para>
+		A socket description [proto:]host[:port] of the local socket
+		(which is used by OpenSIPS for SIP traffic) to be used
+		(if multiple) for sending the probing messages from.
+		</para>
+		<para>
+		For probing gateway the highest priority has socket from gateway
+		configuration in dr_gateways table. Then socket from global
+		<varname>probing_socket</varname> parameter and the lowest
+		priority is default behaviour with auto selected socket wich
+		OpenSIPS listens on.
+		</para>
+		<para>
+		<emphasis>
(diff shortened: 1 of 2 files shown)


bogdan-iancu:
@vasilevalex please see my comment on #1484 about the "socket" field in the "dr_gateways" table.

bogdan-iancu:
OK, I see your point. In this case, the per-gateway socket must have override the global socket, as it is more specific. You need to change the condition when pickup the socket for the pinging - if no per-gw socket -> use the global socket ; right now is the other way around. Please be sure this behavior is described in the documentation for the newly added probing_sock parameter.
Also, I suggest the `probing_socket` name versus the shorter `probing_sock` - clarity must come first :).
And please remove the README file from the commit and keep only the `.xml` file - the README files are autogenerated.

vasilevalex:
@bogdan-iancu I made changes. Sorry, didn't know about README files. Is it clear about socket priorities in docs?

bogdan-iancu:
That is perfect, thank you @vasilevalex

bogdan-iancu:
@vasilevalex , thank to @nikbyte , I realized your PR was against 2.4, not master - all the new features must be against master. I "migrated" the patch from 2.4 to master.

vasilevalex:
@bogdan-iancu Ok, next time I'll use master. I didn't know and took the branch which I use...

spacetourist:
Hi @bogdan-iancu - just wondering if/when this might make its way into a point release for 2.4? I have the same situation as #1484 and only my primary node can be enabled to send probes until this feature is available. Many thanks!

bogdan-iancu:
@spacetourist  - as this is a new feature, it will not be backported to 2.4 . 2.4 is a stable release and only fixes are accepted for it.

spacetourist:
@bogdan-iancu OK, thanks for the update. I'll factor that in. Looking forward to 3.0 :+1:

PR #1493, opened 2018-10-10
Added MI function to enable/disable gateways probing in drouting module

### LINKED ISSUE #1484: Drouting module socket parameter for probing gateways

vasilevalex:
Fixes #1484
The command exactly like in nathelper modile to enable/disable gateways probing.
Can be useful in cluster environment. Command name of course can be changed.
Please review. Moved to master branch.


bogdan-iancu:
Thank you @vasilevalex

PR #1496, opened 2018-10-12
Flexible s.substr transformation

nikbyte:
{s.substr,offset,length}

Return substring starting at offset having size of 'length'.
If offset is negative, then it is counted from the end of the value, -1 being the last char.
In case of positive value, 0 is first char.

Length can be positive or negative.
In case length is negative, it is counted from the end of the value, 0 means last character.

Example:
$var(x) = "abcdefghijk";

$(var(x){s.substr,3,4}) = "defg"
$(var(x){s.substr,1,0}) = "bcdefghijk"
$(var(x){s.substr,2,-2}) = "cdbcdefghi"
$(var(x){s.substr,-5,-3}) = "gh"
$(var(x){s.substr,-5,0}) = "ghijk"
$(var(x){s.substr,-100,0}) = "abcdefghijk"
$(var(x){s.substr,2,100}) = "cdefghijk"
$(var(x){s.substr,100,5}) = ""

### DIFF
--- a/transformations.c
+++ b/transformations.c
@@ -478,45 +478,30 @@ int tr_eval_string(struct sip_msg *msg, tr_param_t *tp, int subtype,
 				j = v.ri;
 			}
 			LM_DBG("i=%d j=%d\n", i, j);
-			if(j<0)
+
+			if (i < 0)
 			{
-				LM_ERR("substr negative offset\n");
-				goto error;
+				if (i*-1 >= val->rs.len)
+				{
+					i = 0;
+				} else {
+					i = val->rs.len + i;
+				}
 			}
+			if (j < 1) j = val->rs.len + j - i;
+			if (i+j >= val->rs.len) j = val->rs.len-i; // cut if length > string
+
 			val->flags = PV_VAL_STR;
 			val->ri = 0;
-			if(i>=0)
+
+			if (i >= val->rs.len) // out of range
 			{
-				if(i>=val->rs.len)
-				{
-					LM_ERR("substr out of range\n");
-					goto error;
-				}
-				if(i+j>=val->rs.len) j=0;
-				if(j==0)
-				{ /* to end */
-					val->rs.s += i;
-					val->rs.len -= i;
-					break;
-				}
-				val->rs.s += i;
-				val->rs.len = j;
+				val->rs.s += val->rs.len;
+				val->rs.len = 0;
 				break;
 			}
-			i = -i;
-			if(i>val->rs.len)
-			{
-				LM_ERR("substr out of range\n");
-				goto error;
-			}
-			if(i<j) j=0;
-			if(j==0)
-			{ /* to end */
-				val->rs.s += val->rs.len-i;
-				val->rs.len = i;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
@nikbyte, I made a review + fixes and uploaded your PR. Please review.
As change (in behavior) the negative underflow index is not allowed (this was the old behavior too). Like `$(var(x){s.substr,-100,0})` will translate into `NULL` and not `"abcdefghijk"`. 
Also, if there is nothing to return (due offset), `NULL` is returned instead of empty string.

nikbyte:
Hi Bogdan,

Why not to allow $(var(x){s.substr,-100,0}) ?
It's really useful. Let me provide you use case.

You have a number and you want to leave from this last n characters, but you don't want to care about string length.

Without this option I must check length of string every time, but without this I can just use something like $(var(x){s.substr,-100,0}) and be sure it will return last n characters or whole string if length is less. This behaviour is standard in many programming languages.

Personally me use this when writing CDR to cut long numbers when length of string is more important for me than start of string.

Before these patch I did something like

   if ($(avp(from_did){s.len}) > 30) {
       $acc_extra(from_did) = $(avp(from_did){s.substr,-30,0});
   } else {
       $acc_extra(from_did) = $avp(from_did);
   }

Now I do just:
$acc_extra(from_did) = $(avp(from_did){s.substr,-30,0});

and I doesn't care about my string length, I know the transformation will always return me the rest of string..

razvancrainea:
Hi, Nick!

I agree you have a good use case, but that may also be a limitation. If we allow *out-of-range* negative values, we will be unable to indicate it when somebody may use it by mistake. Moreover, there will also be an inconsistency, since *out-of-range* positive values return `NULL`, not `""`.

Note that the *out-of-range* error has been removed on both cases (positive or negative). Therefere, in your specific case, you can easily do `$acc_extra(from_did) = $(avp(from_did){s.substr,-30,0});` and get the same behavior, since assigning a `NULL` value to an `$acc_extra` will result in a `""` output in DB, and there will be no errors in the logs.

Best regards,
Răzvan

razvancrainea:
Actually it appears I missed your point - if the negative offset is *out-of-range*, we need to return the entire string, not `""`. In this case I agree the request is valid, but if we do it, we should also have the same behavior for positive values too. Or perhaps add an extra flag in the substitution indicating we should ignore *out-of-range* offsets.

bogdan-iancu:
@razvancrainea , it makes sense to ignore the out-of-range only when underflowing (with negative index), as you can "consider" the string between the way-in-the-left index and whatever offset you may have. In case of overflowing (with positive index), you already pass the string (as position), you are way-in-the-right and the string cannot be "considered" between the index and offset.

bogdan-iancu:
just my 2 cents :)

nikbyte:
@razvancrainea, I have nothing agains NULL instead of empty string. It's ok.

My initial patch allowed behaviour when offset points before string and I offer to return this behaviour because it could be useful in cases like I described above.

Thanks!

PR #1521, opened 2018-11-02
fix quotes in clusterer documentation

fgast:
Just a minor fix in clusterer documentation.


bogdan-iancu:
Thank you @fgast for the fix

PR #1539, opened 2018-11-20
add $var() support for timeout to notify_on_event & wait_for_event

fgast:
We have the need to use a variable timeout in notify_on_event like 

```
$var(timeout) = 12;

notify_on_event("E_UL_CONTACT_INSERT","$avp(filter)", "fork_call", "$var(timeout)");

```

If you have any annotations, please let me know.

### DIFF
--- a/modules/event_routing/event_routing.c
+++ b/modules/event_routing/event_routing.c
@@ -165,7 +165,7 @@ static int mod_init(void)
 	/* try binding to TM if needed and if available */
 	memset( &ebr_tmb, 0, sizeof(ebr_tmb) );
 	if ( is_script_func_used("notify_on_event",-1) ) {
-		/* TM may be used passing the transaction context to the 
+		/* TM may be used passing the transaction context to the
 		 * notification routes */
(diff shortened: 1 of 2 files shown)


modules/event_routing/event_routing.c
```diff
@@ -288,6 +288,16 @@ int fixup_wait(void** param, int param_no)
 static int notify_on_event(struct sip_msg *msg, void *ev, void *avp_filter,
 									void *route, void *timeout)
 {
+	int t = 0;
+
+	if (timeout) {
+		gparam_p gp;
```
bogdan-iancu: @fgast , in terms of coding style we do not encourage / favor variable declarations  in the beginning of blocks . Please move it to the beginning of the functions, here and in the wait_for_event(). Otherwise the PR looks good to me.

fgast: @bogdan-iancu I updated the PR.

bogdan-iancu:
Thank you @fgast !

Issue #1548, opened 2018-11-27
opensips 2.4 incorrect parsing IPv6 address from config


vasilevalex:
If there are several domains with matching by IP:port it is impossible to use IPv6 addresses
`modparam("tls_mgm", "server_domain", "domain6=[2a00:1450:400f:80a::200e]:5061")`
So there is error:
```
Nov 26 14:03:20 test81 opensips-m4cfg: Nov 26 14:03:20 [8922] ERROR:tls_mgm:parse_domain_address: [[2a00] is not an ip
Nov 26 14:03:20 test81 opensips-m4cfg: Nov 26 14:03:20 [8922] ERROR:tls_mgm:parse_domain_address: invalid TLS domain address [[2a00:1450:400f:80a::200e]:5061]
```
Because parse_domain_address looking for the first colon, but it must look for the last one.
So for 2.4 branch this patch works:
```
--- modules/tls_mgm/api.h
+++ modules/tls_mgm/api.h
@@ -69,7 +69,7 @@ static int parse_domain_address(char *val, unsigned int len, struct ip_addr **ip

        /* get the IP */
        s.s = p;
-       if ((p = q_memchr(p, ':', len)) == NULL) {
+       if ((p = q_memrchr(p, ':', len)) == NULL) {
                LM_ERR("Domain address has to be in [IP:port] format\n");
                goto parse_err;
        }
```
In master this function was moved to tls_domain.c, but remains unchanged.

bogdan-iancu:
Thanks @vasilevalex for the report and fix too

PR #1549, opened 2018-11-27
Correct parsing IPv6 address

vasilevalex:
Solves #1548


bogdan-iancu:
Thank you @vasilevalex

PR #1554, opened 2018-12-06
add clusterer replication capabilty to cachedb_local

fgast:
We make heavy use of the cachedb in our scripts, but we do not want to add a 'third party' system to distribute the cache in the cluster. 
So i came up with the idea of adding native clusterer support to the cachedb_local module.

If you have any annotations, please let me know.

### DIFF
--- a/modules/cachedb_local/cachedb_local.c
+++ b/modules/cachedb_local/cachedb_local.c
@@ -58,9 +58,9 @@ lcache_col_t* lcache_collection = NULL;
 url_lst_t* url_list=NULL;
 
 str cache_repl_cap = str_init("cachedb-local-repl");
-int replication_cluster = 0;
+int cluster_id = 0;
 enum cachedb_rr_persist rr_persist = RRP_SYNC_FROM_CLUSTER;
-char *restart_persist;
+char *cluster_persist;
 
 static int w_remove_chunk_1(struct sip_msg* msg, char* glob);
 static int w_remove_chunk_2(struct sip_msg* msg, char* collection, char* glob);
@@ -75,8 +75,8 @@ static param_export_t params[]={
 	{ "exec_threshold",     INT_PARAM, &local_exec_threshold },
 	{ "cache_collections",  STR_PARAM|USE_FUNC_PARAM, (void *)parse_collections },
 	{ "cachedb_url",        STR_PARAM|USE_FUNC_PARAM, (void *)store_urls },
-	{ "replication_cluster",INT_PARAM, &replication_cluster },
-	{ "restart_persistency",STR_PARAM, &restart_persist },
+	{ "cluster_id",INT_PARAM, &cluster_id },
+	{ "cluster_persistency",STR_PARAM, &cluster_persist },
 	{0,0,0}
 };
 
@@ -100,7 +100,7 @@ static dep_export_t deps = {
 		{ MOD_TYPE_NULL, NULL, 0},
 	},
 	{ /* modparam dependencies */
-		{"replication_cluster", get_deps_clusterer},
+		{"cluster_id", get_deps_clusterer},
 	},
 };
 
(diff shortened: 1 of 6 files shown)


bogdan-iancu:
@fgast , this looks like a very interesting features, even so the resulting "shared" cache is not a consistent one. Let us review.

modules/cachedb_local/doc/cachedb_local_admin.xml
```diff
@@ -22,7 +22,15 @@
 	<section>
 		<title>&osips; Modules</title>
 		<para>
-		None.
+		The following modules must be loaded before this module:
```
bogdan-iancu: I would suggest adding a small new main chapter describing the clustering related behavior - what are the limitations when acting as a distributed cluster (in terms of consistency and atomic changes), the sync concept (high level). This chapter may cross link with the the new mod params.

modules/cachedb_local/cachedb_local_replication.c
```diff
@@ -0,0 +1,256 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+ *
+ * History:
+ * --------
+ *  2018-12-06  initial version (Fabian Gast)
```
bogdan-iancu: Please drop this kind of history - we are in progress of removing them as the GIT logs are doing a much better job on this :)

modules/cachedb_local/cachedb_local.c
```diff
@@ -69,6 +75,8 @@ static param_export_t params[]={
 	{ "exec_threshold",     INT_PARAM, &local_exec_threshold },
 	{ "cache_collections",  STR_PARAM|USE_FUNC_PARAM, (void *)parse_collections },
 	{ "cachedb_url",        STR_PARAM|USE_FUNC_PARAM, (void *)store_urls },
+	{ "replication_cluster",INT_PARAM, &replication_cluster },
+	{ "restart_persistency",STR_PARAM, &restart_persist },
```
bogdan-iancu: maybe a better naming would be `cluster_id` and `cluster_persistency` - these 2 params are 100% related to clustering support.

modules/cachedb_local/cachedb_local_replication.c
fgast: This is removed now.

modules/cachedb_local/cachedb_local.c
fgast: I changed this. I noticed there is no consistent way for the naming of cluster variables across the different modules - maybe this should be targeted at some point.

modules/cachedb_local/doc/cachedb_local_admin.xml
fgast: I tried - I am that good at writing so any input or help is appreciated.

modules/cachedb_local/cachedb_local.c
bogdan-iancu: yes, you are perfectly right - we are also trying to 'standardize' the name of the params, so it a kind of work in progress :)

modules/cachedb_local/doc/cachedb_local_admin.xml
bogdan-iancu: What I have in mind here is a short explanation on how the data distribution is done when using the clustering support, like - we simply replicate across the cluster all the local operation. Also the warning as there is no guarantee in terms of data consistency across the cluster, as ops from different node over the same key may be applied in different order on different nodes.
Something like this, so people will not have false expectation from this module

fgast: Did you check the updated PR?

bogdan-iancu: oh, missed that. It looks good, thanks

Issue #1560, opened 2018-12-11
[tm] OpenSIPS crash in timer.c


fgast:
Running 2.4.2, we see a crash in tm/timer.c on final_reponse_handler(): 

```
#0  final_response_handler (fr_tl=<optimized out>) at timer.c:391
        my_ctx = 0x7f6ef2674d48
        t = 0x0
#1  timer_routine (ticks=8048328, set=0x0) at timer.c:1074
        __FUNCTION__ = "timer_routine"
#2  0x00000000004793e2 in handle_timer_job () at timer.c:730
        t = 0x7f6ee28bd948
        __FUNCTION__ = "handle_timer_job"
#3  0x0000000000519b35 in handle_io (idx=<optimized out>, event_type=<optimized out>, fm=<optimized out>) at net/net_udp.c:265
        read = 4802401
#4  io_wait_loop_epoll (h=<optimized out>, t=<optimized out>, repeat=<optimized out>) at net/../io_wait_loop.h:284
        ep_event = {events = 4083230720, data = {ptr = 0x61fb560000007f6e, fd = 32622, u32 = 32622, u64 = 7060331398833405806}}
        r = 3
        i = -228111032
#5  udp_start_processes (chd_rank=0x7ffe88dff690, startup_done=0x18) at net/net_udp.c:389
        si = 0x1
        pid = -228111032
        i = 3
        __FUNCTION__ = "udp_start_processes"
```

We have seen this some time ago on 1.11 and build a temporary fix for this issue (see PR #1561) - maybe this is does not fix the original problem but prevents openSIPS from a crash. 

We have a vague idea on how to reproduce the crash, but the szenario and configuration for this is quite complex. Currently the crash happens about 3-4 times a day (without the fix). 

```
version: opensips 2.4.2 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
main.c compiled on  with gcc 4.9.2
```

bogdan-iancu:
[...]

PR #1561, opened 2018-12-11
rels #1560 - prevent crash in tm/timer.c

fgast:


bogdan-iancu:
@fgast , I would try to understand and fix #1560 ...let's see how it goes.

PR #1582, opened 2019-01-10
event_routing: wait_for_event() timeout support

vitalikvoip:
wait_for_event() resumes a transaction after timeout expires


vitalikvoip:
That's interesting. I haven't changed anything in pua_dialoginfo but compilation stopped there.

PR #1583, opened 2019-01-14
Added async for PostgreSQL db_postgres module

ar45:


ar45:
Thanks @bogdan-iancu I see you already implemented it.

PR #1586, opened 2019-01-17
Fixes for db raw_query

ar45:
- db_postgres when executing a SQL query with multiple statements, only the last PGresult was free'd resulting in a memory leak.

- db_postgres: When multiple statements are executed in a single query, return the last PGresult which had result tuples.
    That is desired behaviour when executing somthing like the following
    ```
    BEGIN;
    UPDATE tbl SET col = 'blah';
    SELECT col FROM tbl;
    COMMIT;
    ```
    so that we get the result of the last SELECT instead of the COMMIT statement.

- db_postgres: Do not resubmit a multi statement query if any of the statements succeeded.
    Otherwise, a statement like `UPDATE tbl SET col = col + 1` would be incorectly executed twise.

- db_mysql/db_postgres: When `db_init_async` does not return a async connection (ie, max transferres exceeded.)
    no results were set, and to the user it appeared as if the query failed, because
    `db_mysql_store_result` and `db_postgres_store_result` where not called by async_resume function.
    So instead of executing the query synchronously, return an negative number `ASYNC_CON_UNAVAILABLE`
    and let `ops_async_dbquery()` fallback to sync connection.

    This bug also effected db_virtual async where the same query could theoretically be executed on all nodes
    since the sync query had returned negative result..

### DIFF
--- a/db/db_async.h
+++ b/db/db_async.h
@@ -34,6 +34,12 @@
 
 #include "db_con.h"
 
+/**
+ * Used to indicate when an async trasferrer connection is not available.
+ * ie. max transferres exceeded.
+ */
+#define ASYNC_CON_UNAVAILABLE -128
+
 typedef int (*get_con_fd_f) (void *con);
 
 /**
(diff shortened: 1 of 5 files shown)


ar45:
build error is caused by 4f6dd06 will build once #1587 is merged.

ar45:
@bogdan-iancu can you review?

PR #1599, opened 2019-01-28
Remove double free on bad input

rrb3942:
If bad input is passed to a mi_json command it causes a double free.
This removes the extra free.


bogdan-iancu:
Thank you @rrb3942 for the fix. Unfortunately this is not applicable anymore for trunk as the whole MI part was reworked.
Still, this is a 100% valid fix for 2.4 and 2.2 - so I manually applied your fix there.

PR #1600, opened 2019-01-28
Disable Expect: 100-continue behavior for large posts

rrb3942:
By default on large posts the libcurl library will send the headers with a 'Expect: 100-continue' header, then wait for a 100 response from the server before sending the actual body. This gives the server a chance to reject large requests before receiving the entire body.

This is good client behaviour for a client talking to random web servers, but creates an extra network round-trip.

Since any server that is interacting with db_http should be tested and tightly coupled, this patch removes Expect behavior, and causes the client to fully post the body with the headers immediately.

### DIFF
--- a/modules/db_http/db_http.c
+++ b/modules/db_http/db_http.c
@@ -46,6 +46,7 @@ int cap_id = 0;
 int cap_replace = 0;
 int cap_insert_update = 0;
 int use_ssl = 0 ;
+int disable_expect = 0;
 
 unsigned int db_http_timeout = 30000; /* Default is 30 seconds */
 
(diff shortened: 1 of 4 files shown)


bogdan-iancu:
Thank you @rrb3942  - to be 100% flexible and backward compatible, maybe a modparam should be added to enable/disable the "wait for 100 resp" - and have it set by default to a backward compatible value.

bogdan-iancu:
Again, thank you @rrb3942 . Could you add the param in the module docs please - only in the `.xml` file !

rrb3942:
Knew I forgot something. Just added the dock update.

bogdan-iancu:
Thank you @rrb3942 , just accepted you PR.
BTW, have you got my emails ;) ?

rrb3942:
Hmm, I'm actually not seeing any emails besides these notices from github (which gmail is terrible at filtering).

I just set a public email on github that can be used.

PR #1601, opened 2019-01-28
Don't clip notify socket to older rtpproxies when using unix sockets

rrb3942:
If using an older version of rtpproxy (without NOTIFY_WILD) and a unix socket for timeout, the timeout socket was getting clipped.

This prevents the clipping so that the unix socket is properly sent to rtpproxy.


razvancrainea:
Thanks for the PR! I will also push this down to 2.2.

PR #1613, opened 2019-02-08
 Add the ability to call uac_auth() without incrementing CSEQ

mstocco:
This solves a problem where a UA is known to accept SIP Authorization without requiring the CSEQ being incremented and allowing for relaying without CSEQ adjustments.

### DIFF
--- a/modules/uac/README
+++ b/modules/uac/README
@@ -18,6 +18,7 @@ UAC Module
               1.3.3. rr_from_store_param (string)
               1.3.4. rr_to_store_param (string)
               1.3.5. force_dialog (int)
+              1.3.6. increment_cseq (int)
 
         1.4. Exported Functions
 
@@ -161,6 +162,19 @@ modparam("uac","rr_to_store_param","my_Tparam")
    Example 1.5. Set force_dialog parameter
 ...
 modparam("uac", "force_dialog", yes)
+...
+
+1.3.6. increment_cseq (int)
+
+   Under most circumstances, SIP Authenication can fail if the
+   second INVITE (with the Authorization header) uses the same
+   CSEQ value.  The default behavior is to increment the CSEQ.
+
+   Default value is 1.
+
+   Example 1.6. Set increment_cseq parameter
+...
+modparam("uac", "increment_cseq", 1)
 ...
 
 1.4. Exported Functions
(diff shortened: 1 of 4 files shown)


bogdan-iancu:
Hi @mstocco !
The correct disabling will be not changing and restoring to a cseq value equal to the received one, but skipping the whole cseq changing code and returning [after this line](https://github.com/OpenSIPS/opensips/blob/master/modules/uac/auth.c#L274) .
Even more, to be flexible, I would strongly suggest to have this options per function call (via param), not globally per module.

modules/uac/README
```diff
@@ -163,6 +164,19 @@ modparam("uac","rr_to_store_param","my_Tparam")
 modparam("uac", "force_dialog", yes)
 ...
 
+1.3.6. increment_cseq (int)
+
+   Under most circumstances, SIP Authenication can fail if the
```
wdoekes: *Authenication (+t)

modules/uac/auth.c
```diff
@@ -120,7 +121,9 @@ int apply_cseq_op(struct sip_msg *msg,int val)
 		return -1;
 	}
 
-	cseq_no=cseq_no+val;
+	if (increment_cseq) {
+		cseq_no=cseq_no+val;
```
wdoekes: I'd suggest adding some spaces here in the vicinity while you're touching the code.
```
cseq_no = cseq_no + val;
```
or
```
cseq_no += val;
```
.. is so much more readable.

mstocco:
> Hi @mstocco !
> The correct disabling will be not changing and restoring to a cseq value equal to the received one, but skipping the whole cseq changing code and returning [after this line](https://github.com/OpenSIPS/opensips/blob/master/modules/uac/auth.c#L274) .
> Even more, to be flexible, I would strongly suggest to have this options per function call (via param), not globally per module.

Hi @bogdan-iancu,
As you suggested, the idea of an exported, global parameter was replaced with an optional parameter (string) passed to the uac_auth function.  Specifically, if uac_auth(“CSEQ_SKIP”) is in the route script,  uac_auth returns cleanly where you expressed it should, [at line 274](
https://github.com/OpenSIPS/opensips/blob/master/modules/uac/auth.c#L274).

Anything but “CSEQ_SKIP” (or no param at all) is intended to have uac_auth() continue as it always has.  Thoughts?

PR #1614, opened 2019-02-08
Correct presence list MI flush in OpenSIPS 2.4

wdoekes:
Hi there!

We're still using the soon-to-be-removed mi_datagram.

Apparently it has some bytes-left counting issues which were revealed because mi_list_phtable was flushing mi-nodes irregularly.

This fixes things.

### DIFF
--- a/modules/mi_datagram/mi_datagram_writer.c
+++ b/modules/mi_datagram/mi_datagram_writer.c
@@ -65,7 +65,7 @@ static inline int mi_datagram_write_node(datagram_stream * dtgram,
 
 
 	start = p = dtgram->current;
-	end = dtgram->start + dtgram->len;
+	end = dtgram->current + dtgram->len;
 	LM_DBG("writing the name <%.*s> and value <%.*s> \n",
 			node->name.len, node->name.s, node->value.len, node->value.s);
 	/* write indents */
@@ -200,12 +200,12 @@ int mi_datagram_write_tree(datagram_stream * dtgram, struct mi_root *tree)
 	if (datagram_recur_write_tree(dtgram, tree->node.kids, 0)!=0)
 		return -1;
 
-	if (dtgram->len<=0) {
+	if (dtgram->len <= 1) {
 		LM_ERR("failed to write - EOC does not fit in!!!\n");
 		return -1;
 	}
 
-	*(dtgram->current) = '\n';
+	*(dtgram->current++) = '\n';
 	dtgram->len--;
 	*(dtgram->current) = '\0';
 
@@ -297,8 +297,6 @@ int mi_datagram_flush_tree(datagram_stream * dtgram, struct mi_root *tree)
 		return -1;
 	}
 
-	*(dtgram->current) = '\n';
-	dtgram->len--;
 	*(dtgram->current) = '\0';
 
 	return 0;
--- a/modules/presence/presence.c
+++ b/modules/presence/presence.c
@@ -716,23 +716,26 @@ static struct mi_root* mi_list_phtable(struct mi_root* cmd, void* param)
 	struct mi_root *rpl_tree;
 	struct mi_node* rpl;
 	pres_entry_t* p;
-	unsigned int i;
+	unsigned int i, j;
 
 	rpl_tree = init_mi_tree( 200, MI_OK_S, MI_OK_LEN);
 	if (rpl_tree==NULL) return NULL;
 	rpl = &rpl_tree->node;
 	rpl->flags |= MI_IS_ARRAY;
 
-	for(i= 0; i<phtable_size; i++)
+	for(i = 0, j = 0; i < phtable_size; i++)
 	{
 		lock_get(&pres_htable[i].lock);
 		p = pres_htable[i].entries->next;
 		while(p)
 		{
 			if(mi_print_phtable_record(rpl, p)<0) goto error;
-			p= p->next;;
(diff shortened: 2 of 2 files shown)


modules/mi_datagram/mi_datagram_writer.c
```diff
@@ -65,7 +65,7 @@ static inline int mi_datagram_write_node(datagram_stream * dtgram,
 
 
 	start = p = dtgram->current;
-	end = dtgram->start + dtgram->len;
+	end = dtgram->current + dtgram->len;
```
bogdan-iancu: I'm not sure this is correct - the `dtgram` has a buffer of a fixed `len` (starting with `start`). The `current` is the offset (inside the buffer) with the writing. And we need the `end` as end of the buffer to prevent the overflow.  The end of the buffer is the same all the time (start + end), disregarding where the writing offset is.

bogdan-iancu: ok, sorry, I see, the `len` is the remaining len, not the total len...you are right here :+1:

modules/mi_datagram/mi_datagram_writer.c
```diff
@@ -297,8 +295,6 @@ int mi_datagram_flush_tree(datagram_stream * dtgram, struct mi_root *tree)
 		return -1;
 	}
 
-	*(dtgram->current) = '\n';
-	dtgram->len--;
```
bogdan-iancu: The idea with this extra `\n` is to have it as End OF Message - I know it makes no sense as it is datagram, but the module shares the same grammar as mi_fifo (which is stream oriented). So you need to have the empty line at the end of the whole response text. So, the final writing should keep this `\n`, but not during the flushing.....I guess only the lines 208-209 should be removed and these ones kept.

modules/presence/presence.c
```diff
@@ -897,8 +903,10 @@ static struct mi_root* mi_list_shtable(struct mi_root* cmd, void* param)
 		}
 		lock_release(&subs_htable[i].lock);
 
-		if ((j % 50) == 0)
+		if (j >= 50) {
 			flush_mi_tree(rpl_tree);
+			j = 0;
+		}
```
bogdan-iancu: I think this change does not bring any modification in behavior - correct me if I'm wrong.

wdoekes: Well. This was the direct cause of our problems:

- mi_list_shtable was called
- j = 0
- lock is attained on subs_htable[0]
- no matches are found, j is still 0
- unlock op subs_htable[0]
- j%50 == 0, thus we flush even though mi_print_shtable_record was not added

Alternately, we could go from j=49 to j=52 inside the for, and then we wouldn't flush, even though it is time.

Because the increment of j is done inside the for(), we're not guaranteed to hit this flush-condition exactly at j%50==0. Therefore the change so we flush at around 50 or more.

modules/mi_datagram/mi_datagram_writer.c
wdoekes: > So you need to have the empty line at the end of the whole response text.
> So, the final writing should keep this `\n`, but not during the flushing.....
> I guess only the lines 208-209 should be removed and these ones kept.

Okay, I can agree to that, but then we need to `(dtgram->current++)`, otherwise we'd replace the LF with NUL on (original) line 302.

wdoekes: Force pushed and updated.

modules/presence/presence.c
bogdan-iancu: Hi @wdoekes , if this is the issue, maybe the easiest fix will be to start with `j=1;`

modules/mi_datagram/mi_datagram_writer.c
bogdan-iancu: :+1:

modules/presence/presence.c
wdoekes: That would fix problem 1:
> j%50 == 0, thus we flush even though mi_print_shtable_record was not added

But it would not fix problem 2:
> Alternately, we could go from j=49 to j=52 inside the for, and then we wouldn't flush, even though it is time.

Because there may be several matches inside the for(), causing the `if((j%50)==0)` not to be reached at every increment of j.

wdoekes: The alternate solution was to move the `if()` to inside the `for()`, but that would also put it inside the locked section, which I wanted to avoid.

bogdan-iancu: oh, thanks for your patience here, I see your point. I was a bit confused as I wasn't expecting this. In other modules, like [usrloc](https://github.com/OpenSIPS/opensips/blob/2.4/modules/usrloc/ul_mi.c#L441)  or [dialog](https://github.com/OpenSIPS/opensips/blob/2.4/modules/dialog/dlg_hash.c#L1426), the flush is done under the lock. If acceptable for you, I would prefer to keep the same approach here too (for consistency purposes). Of course, starting with j = 1 :)

wdoekes: No problem. Force pushed. It now looks like this:
https://github.com/OpenSIPS/opensips/pull/1614/commits/fd3a6e4a71674c626c8b7f97f9892d08e48d0f7f#diff-a09dfe59b017a1cfe2ef7b05b4f97203R900

bogdan-iancu:
Thanks a lot @wdoekes

Issue #1590, opened 2019-01-18
Hangs and segfault with concurrent ssl handshakes on debian stretch


hafkensite:
I have been able to crash / hang opensips 2.4.4 on Debian stretch using nothing more than ab (Apache benchmark).

The bug can be reproduced with a clean installation of openssl, the wss module and a snakeoil certificate. Running ab with a concurrency of 100 easily triggers the effect. We have also seen this happening in production. 
Attached is a [zip file](https://github.com/OpenSIPS/opensips/files/2772812/ssl-stretch-hangs.zip) with a `Dockerfile`, `opensips.cfg` and a readme on how to replicate the results. 

When running the test on jessie I get the expected errors about not being a proper websocket connection. On stretch the process hangs after a while with a lot of error messages like:
```
[14] ERROR:proto_wss:tls_accept: New TLS connection from 172.17.0.1:44130 failed to accept
[14] ERROR:proto_wss:tls_print_errstack: TLS errstack: error:1409441B:SSL routines:ssl3_read_bytes:tlsv1 alert decrypt error
[14] ERROR:proto_wss:wss_read_req: cannot fix read connection
```

When using consecutive connections there are no problems, increasing the concurrency the problem is triggered sooner. 

My suspicion is that it has to do with the openssl 1.0 to 1.1 transition, especially the multi-threading. 

When building the binaries from source (no Debian packages) the master branch does not crash, but branch 2.4 (and tag 2.4.4) do crash.

How can we further diagnose and fix this issue?

hafkensite:
Building against the 1.0.x version of libssl does fix the issue, I did that with this patch:

```diff
diff --git a/packaging/debian/control b/packaging/debian/control
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -30,7 +30,7 @@ Build-Depends: bison,
                libsctp-dev [linux-any],
                libsnmp-dev,
                libsqlite3-dev,
-               libssl-dev,
+               libssl1.0-dev | libssl-dev (<< 1.1),
                lsb-release,
                uuid-dev,
                libxml2-dev,```

bogdan-iancu:
@hafkensite , the libssl became really multi-processes unfriendly during the last releases - mainly removing configuration options and assuming all the time multi-threading.
Let me check on Monday with @razvancrainea if this is an issue he has it on the list or it is something new.

razvancrainea:
[...]

PR #1633, opened 2019-03-11
Build debian packages against libssl 1.0 to fix segfault and locking issues

hafkensite:
Fix for issue #1590, in which OpenSIPS 2.4 sefaults on concurrent TLS connections. 
This PR updates the Debian control to ensure libssl 1.0 is used. 

The issue seems to be fixed in master/3.0 by the changes in the tls domain matching.


wdoekes:
Perhaps amend the commit msg with the "why"

volga629:
We see libssl locking issue on fedora server too force downgrade to 1.0 will might introduce unsupported problems or vulnerabilities. I think gnutls is might be optimal. But from another hand right opensips tls implementation are not suitable for production on latest Linux distros with libssl 1.1

bogdan-iancu:
This is a bit of a radical workaround - it may create a bit of a mess in terms of OS installation. And it works only when deploying from packages. If you simply compile and install, you are still in big problems.
We already have on the TODO list to find out a fix at the code level, to try to eliminate the deadlock in the ssl lib by the way we are using it. There are no simple solutions here, still this is the plan

hafkensite:
That would mean that the TLS packages on the official apt repository of OpenSIPS are useless (or worse, they work initially but let you down when concurrency triggers the bug).

I understand that the proper way forward is to have a better handling of the deadlocks in the ssl lib. Are there plans to also include this in the 2.4 branch, or is that just for 3.0? 
Otherwise I would prefer the official 2.4 packages that are provided to be compiled against openssl 1.0.

bogdan-iancu:
This is obsoleted / fixed by 3d9132e722c18c048a66320265977654bccfa341 and 1b624d2d46bf07e9b0b1f198a93267392371a2e8

PR #1634, opened 2019-03-11
New String Transformation: s.reverse

rrb3942:
Adds the s.reverse string transformation which reverses a string.

Example:
$var(forward) = "onetwothree";
$var(reverse) = $(var(forward){s.reverse}) //Contains "eerhtowteno";


bogdan-iancu:
Thank you @rrb3942

PR #1635, opened 2019-03-12
Add Session-Expires header to compact_form_hdrs

rrb3942:
Even though the docs said it was supported, Session-Expires was missing from the list used to generate the mask for checking if a compact form was available. It wasn't actually being made into short form on calls to mc_compact().

This just adds it to the list so it will can be changed to short form by mc_compact().


bogdan-iancu:
:+1:

PR #1639, opened 2019-03-19
Fix build with modern compilers by properly defining inline functions

sobomax:
Fix build with modern compilers by properly defining functions as "static inline" instead of just "inline". For some reason 2.1 is the only branch affected, 1.x and 2.[234] are fine.

This popped up after we moved our travis test base to xenial. It's been pretty painful to keep relevant with ancient gcc 4.8 otherwise. Please see 2.1* stand out below:

https://travis-ci.org/sippy/voiptests/builds/508412647

### DIFF
--- a/parser/parse_sst.c
+++ b/parser/parse_sst.c
@@ -30,17 +30,17 @@
 #include "../mem/mem.h"
 
 
-inline int/*bool*/  is_space( char c ) { return (c == ' ' || c == '\t'); }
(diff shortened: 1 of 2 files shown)


sobomax:
CI is broken, I suspect nobody touched 2.1 in a long time, so it got rusted. :)

bogdan-iancu:
2.1 branch is no longer maintain.  The error seems to be related to some package version (for lua), rather than the code.
Please move your PR for 2.2 or 2.4 branches - these are the only maintained branches at this point.

Thanks!

PR #1643, opened 2019-03-25
db_postgres: Fix errors on bulk insert after delete/update

rrb3942:
Updates or Deletes on a PostgreSQL connection cause 'submit_func_called' to be set.

If bulk inserts are enabled and a delete or update is run before it, db_postgres_insert() may then try to process results that don't actually exist, causing error messages.

'submit_func_called' only seems to be used by the insert function, so I have it reset this value before calls db_do_insert(). This prevents stale state from affecting the insert query.

I tested this against 2.4.


bogdan-iancu:
Thank you @rrb3942 ,  good catch!

PR #1645, opened 2019-03-26
Fixed bug where any addr get positive match in config if cluster_id i…

goharahmed:
Fixed bug where any addr get positive match in config if cluster_id is invalid or Invalid IP addr is given to cluster_check_addr()

Example, start a cluster with cluster_id = 1 but when checking a source address belonging to list of clustered nodes provide different cluster ID. (cluster_id should be loadable into an avp to avoid such mismatches in future) Atleast the function should never return a positive match for erroneous cases.

opensips.cfg 
```
$var(cl_id) = 2;
if (cluster_check_addr("$var(cl_id)", "1.2.3.4")) {
        xlog("L_INFO"," Yes the provided IP is a connected cluster node\n");         
}
```

### DIFF
--- a/modules/clusterer/clusterer.c
+++ b/modules/clusterer/clusterer.c
@@ -910,7 +910,7 @@ static int ip_check(cluster_info_t *cluster, union sockaddr_union *su, str *ip_s
 				return 1;
 		} else {
 			LM_ERR("No address to check\n");
-			return -1;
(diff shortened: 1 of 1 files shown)


liviuchircu:
Excellent catch and fix, @goharahmed !  It definitely shouldn't return `true` for those cases!

Issue #1666, opened 2019-04-16
[BUG] Travis fails


wdoekes:
```
$ CC_EXTRA_OPTS=-Werror FASTER=1 NICER=0 make -j6 exclude_modules="db_oracle osp sngtc cachedb_cassandra cachedb_couchbase cachedb_mongodb" all 2>&1 | grep error: | sort
dlg_replication.c:1198:6: error: ‘ret’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
dlist.c:512:18: error: ‘next_hop_host’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
dlist.c:759:37: error: ‘next_hop_host’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
dlist.c:759:37: error: ‘next_hop_host’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
ldap_mod.c:288:52: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
ul_mi.c:225:25: error: ‘ct_item’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
ul_mi.c:225:25: error: ‘ct_item’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:34:10: error: ‘__builtin_memcpy’: specified size between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
```
Travis chokes on these. See e.g. https://travis-ci.org/OpenSIPS/opensips/jobs/520649398 which ends in:
```
ldap_mod.c: In function ‘w_ldap_result’:
ldap_mod.c:288:52: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
  return ldap_write_result(msg, attr_name, dst_avp, (int)avp_type, subst);
```

wdoekes:
Fixed in 9e2c279a7a8847b29d876c13d60937ffd6dc020d.
```
20:38 < travis-ci> OpenSIPS/opensips#5544 (3.0.0-beta - 67c51ac : Razvan Crainea): The build passed.
20:38 < travis-ci> Change view : https://github.com/OpenSIPS/opensips/compare/3.0.0-beta
20:38 < travis-ci> Build details : https://travis-ci.org/OpenSIPS/opensips/builds/520902718
```
:ok_hand:

PR #1674, opened 2019-04-23
Work around a few poor compiler warnings

wdoekes:
Like:

    dlist.c: In function ‘get_all_ucontacts’:
    dlist.c:759:37: error: ‘next_hop_host’ may be used uninitialized in this
      function [-Werror=maybe-uninitialized]

And:

    In function ‘memcpy’,
      inlined from ‘get_domain_db_ucontacts’ at dlist.c:294:4:
    /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34:10: error:
      ‘__builtin_memcpy’: specified size between 18446744071562067968 and
      18446744073709551615 exceeds maximum object size 9223372036854775807
      [-Werror=stringop-overflow=]

And one probable bug:

    b2b_logic.c: In function ‘mi_b2b_list’:
    b2b_logic.c:1548:10: error: ‘server_item’ may be used uninitialized in this
        function [-Werror=maybe-uninitialized]
    ^- should be client_item

See also #1666

### DIFF
--- a/modules/b2b_logic/b2b_logic.c
+++ b/modules/b2b_logic/b2b_logic.c
@@ -1545,7 +1545,7 @@ static mi_response_t *mi_b2b_list(const mi_params_t *params,
 					if (!clients_item)
 						goto error;
 
-					if (add_mi_number(server_item, MI_SSTR("index"), index) < 0)
+					if (add_mi_number(clients_item, MI_SSTR("index"), index) < 0)
 						goto error;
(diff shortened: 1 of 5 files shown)


liviuchircu:
All fixes look good to me.  Thanks for this!

PS: I tried to reproduce the usrloc compiler warning both with `gcc 4.8` and `gcc 5.5`, and failed miserably.  It seems either Travis is using `gcc 5.4`, which I could not get hold of, or I wasn't controlling the warning flags well enough.

Issue #1669, opened 2019-04-16
[BUG]ERROR:tm:update_cloned_msg_from_msg: failed to clone lumps, not updating


rrb3942:
**OpenSIPS version you are running**

```
version: opensips 2.4.5 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, HP_MALLOC, FAST_LOCK-FUTEX-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 78984b2
main.c compiled on 09:32:38 Apr 15 2019 with clang 4.2.1
```

**Describe the bug**

I sometimes see the following error messages in my logs while processing traffic:
Apr 12 12:49:54 testing /usr/sbin/opensips[22337]: ERROR:tm:update_cloned_msg_from_msg: failed to clone lumps, not updating
Apr 12 12:49:54 testing /usr/sbin/opensips[22337]: ERROR:tm:update_cloned_msg_from_msg: failed to clone body lumps, not updating

**To Reproduce**

Sporadic. Have not been able to link to a certain set of signalling or functions. When I check signalling everything is as expected, I don't see any corrupt messages or unwanted changes.

**Expected behavior**

No error messages.

**Relevant System Logs**

**OS/environment information**
 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

**Additional context**

We use topologyhiding, mc_compac, uac_replace_to, uac_replace_from functions to manipulate signalling and have hp_malloc enabled.

Looking at the update_cloned_msg_from_msg function it looks like shm_malloc would have to fail for these messages to print, but I'm not getting any other error logs, which should happen on a malloc failure.

I updated the error to include the len that should have been malloced and got the following for the l1_len block where c_msg->add_rm ends up being NULL:
ERROR:tm:update_cloned_msg_from_msg: failed to clone lumps, not updating. Did not allocate 200

So it doesn't look like it is trying to allocate some huge amount either. Could their be an issue with hp_shm_malloc? Or a race happening against the shm version of the message?

rrb3942:
[...]

PR #1678, opened 2019-04-23
Only push message updates for the same method

rrb3942:
See https://github.com/OpenSIPS/opensips/issues/1669

After some more testing, I think this is all that is required to fix the race without introducing more locking.

I am currently in the process of testing this, as it is a race issue it can be difficult to trigger at times.

### DIFF
--- a/modules/tm/tm.c
+++ b/modules/tm/tm.c
@@ -712,10 +712,19 @@ static int do_t_cleanup( struct sip_msg *foo, void *bar)
 
 	reset_e2eack_t();
 
-	if ( (t=get_t())!=NULL && t!=T_UNDEFINED &&   /* we have a transaction */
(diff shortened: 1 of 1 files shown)


rrb3942:
This tested good for me. No more errors or crashes. Did not see any negative impact to signalling.

On a negative reply the ACK is considered part of the same transaction. This causes a race between the INVITE and the ACK when the checks and calls to update_cloned_msg_from_msg in do_t_cleanup() are done, whereas both may try to update the copy of the message held inside the transaction. This can cause errors and crashes (segfault). Updating the copy of the INVITE in the transaction from the ACK is most likely wrong, so we only update the transaction copy if our local one is of the same method.

Retransmits do not seem to trigger do_t_cleanup() so no further locking should be needed.

PR #1680, opened 2019-04-24
Fix mantainer-clean -> maintainer-clean typo

wdoekes:


liviuchircu:
And now we've even less mantaining to do!

PR #1681, opened 2019-04-25
Exclude rabbitmq_consumer by default.

sobomax:


bogdan-iancu:
It seams @vladpaiu already did this via 2497e12

PR #1699, opened 2019-05-15
LB will always to one dst when use lb_next

taozhiying:
The current dst is not a larger load. When skipping, the index of the bitmask does not move.


bogdan-iancu:
Thank you @taozhiying , good catch here !!

PR #1735, opened 2019-06-19
Schema extension: registrant.xml

rzerres:
## current status ##
In the given implementation of opensips will handle the syntax
for the aor URI hardcoded to RFC3261 (Registration Mode: sip:<PSTN>@<registar>:<port>

## Updated status ##
This patch will append the schema to introduce a new field `registartion_mode`.
This field holds a valid registration type per row.

### DIFF
--- a/db/schema/registrant.xml
+++ b/db/schema/registrant.xml
@@ -44,6 +44,14 @@
 		<description>URI pointing to the outbond proxy.</description>
 	</column>
 
+	<column id="registrar_mode">
+		<name>registrar_mode</name>
+		<type>string</type>
+		<size>64</size>
+		<default/>
+		<description>Mode used to register the address of record.</description>
+	</column>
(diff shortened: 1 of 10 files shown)


liviuchircu:
Your `&registrar_mode_len;` entity is undefined, so the final result in the SQL files resembles:

```
    ...
    proxy CHAR(255) DEFAULT NULL,
    registrar_mode CHAR DEFAULT '' NOT NULL,
    aor CHAR(255) DEFAULT '' NOT NULL,
    ...
```

... which MySQL (at least) interprets as `CHAR(1)`.  Is that the indended effect?  If not, please define a `&registrar_mode_len;` under `db/schema/entities.xml`, or just hardcode your length directly: I doubt anyone will ever re-use the `&registrar_mode_len;` entity in their schemas.

Issue #1683, opened 2019-04-26
[CRASH] Record-routing after rest ASYNC get t_resume_async  


varghesepaul:
**OpenSIPS version you are running**

```
version: opensips 2.2.7 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
main.c compiled on 03:39:17 May 25 2018 with gcc 4.8.5

```

**Crash Core Dump**

BT Logs : https://www.dropbox.com/s/n2wcr738npkpxja/Opensips-Crash-Report

**Describe the traffic that generated the bug**

UDP Invite call and in the route block , we are calling a ASYNC get request , then got the response and in resume block we are calling the record_route. 

route {
                ..................
                async(
                        rest_get(
                                "$var(AUTH_URL)", "$var(auth_resp)", "$var(ct)", "$var(rcode)"
                        ),
                        auth_resume
                );
                .....................................
}

route[auth_resume] {
        # check auth
        # record routing
        if (!is_method("REGISTER|MESSAGE")) {
                # do the record_route_preset if the request from and to internal servers
                if ($si =~ "^10\..*" ) { 
                        xlog("L_INFO","TRACE[$ci]:route[record_route_relay]:Symmetric record-route for LAN request\n");
                        record_route_preset("$Ri:$Rp");
                } else {
                        xlog("L_INFO","TRACE[$ci]:route[record_route_relay]: Record routing\n");
                        record_route();
                }
        }
}

**To Reproduce**

Just make a call to do record route in async get response block. 

**Relevant System Logs**

2019-04-26 13:58:36.787 /usr/sbin/opensips[30523]:CRITICAL:core:free_lump: called on a not free-able lump:0x7f7dc5ee9780 flags=2

2019-04-26 13:59:40.642 /usr/sbin/opensips[30535]:CRITICAL:core:receive_fd: EOF on 21
2019-04-26 13:59:40.642 /usr/sbin/opensips[30516]:INFO:core:handle_sigs: child process 30523 exited by a signal 6
[...]

PR #1738, opened 2019-06-19
Do not simply link the buffered lumps, but better clone them -> this …

varghesepaul:
…will avoid a mixage of lump types (shm versus pkg) when using async()

This fix is already present for record_route function and missed to add for record_route_preset function. Getting the crash when i tested with 2.4.6 for record_route_preset. This fix is tested and it is working fine. 

Reference : https://github.com/OpenSIPS/opensips/issues/1683


bogdan-iancu:
Thank you @varghesepaul , good catch !!

PR #1742, opened 2019-06-25
uac_replace_xxx - Remove extra spaces and strip display name on "auto" restore

rrb3942:
This pull requests makes to changes to the uac_replace_xxx methods.

1. When stripping the display name it removes everything from the start of the display name up to the start of the uri. This removes trailing spaces and is just a little bit of cleanup on the resulting header.

Previously if you stripped a display name (when was already present) you would end with an extra space in the resulting header. Something like this:
From: <sip:alice@atlanta.example.com>;tag=9fxced76sl**

After the patch:
From:  <sip:alice@atlanta.example.com>;tag=9fxced76sl

2. Currently, changes to the display name are not saved for use with restore_mode "auto". The current behaviour for in-dialog requests in this mode is to replace the URI leaving the display name as-is. This can lead to inconsistent display names and in the worst case information leaking when the display name is changed for privacy reasons (enforcing Privacy: id, etc). This patch makes restore_mode "auto" always strip the display name. I think this is better than inconsistent or leaking display names.


bogdan-iancu:
@rrb3942 , in regards to the second change (on deleting the display name in auto mode) - the same effect is to be achieved (if really wanted) by doing `uac_replace_from("","")` on the sequential requests. This gives you the possibility to decide what to do - to remove it or not. With your change, you will not be able to preserve it (if really wanted) on a sequential request.

rrb3942:
@bogdan-iancu Thanks for pointing that out. I hadn't thought of that scenario. I had gone back and forth on the second change, because even though I think it is probably a good default behaviour it can make certain scenarios harder.

I think stripping the display name on sequential is rarely a bad thing to do and consistent.

The current behaviour I would say is inconsistent (when changing the display name) and at worse leaks information (without additional work on the script writers part), which I would consider a bad thing. I would also consider the name leaking on sequential requests 'surprising' to script writers. Right now, for consistent and unsurprising behaviour (stripping or changing) on display name changes users have to resort to additional calls to the replace functions on sequential requests.

What about either a parameter for auto mode, or a new restore mode (auto_strip, or auto_simple?) that causes the name to be stripped on sequential requests? And a doc update to explain the name replacement behaviour.

That way users could pick auto with stripping if they don't want to worry about inconsistency and leaking, and if they have more advance scenarios they can choose auto and handle it in script on sequential requests.

bogdan-iancu:
@rrb3942 , I see you point on (2) - please make this a separate PR. 

For (1) - what is the reason behind the cleanup ? the spaces at the end of the hdr are legal from RFC perspective ;).

rrb3942:
Removed the name stripping on "auto" restore. Will make a new PR for that.

Is 'It looks nicer' a good enough reason on stripping the extra spaces? :)

RFC wise it is allowed, and I haven't run into any systems that have issue with it. Does nag at me every time I see it in a call trace though. 

A more 'valid' reason may just be we should be trying to match it as if we built the entire from header from scratch ourselves. And I think in that case we would not be including extra spaces.

modules/uac/replace.c
```diff
@@ -270,8 +270,19 @@ int replace_uri( struct sip_msg *msg, str *display, str *uri,
...
 				body->display.len,body->display.s);
+
+			start_offset = body->display.s-msg->buf;
+
+			//If replacing leave trailing spaces
+			if (display->len) {
+				display_len = body->display.len;
+			} else {
+			//Removing, strip all characters up to the URI
+				display_len = ((body->uri.s-msg->buf) - start_offset) - 1;
```
bogdan-iancu: I guess the `-1` is for leaving a single space between the URI and the display, right ? If so, I'm not sure that the case of an enclosed URI is properly handled. If I'm not wrong, the `<` and `>` are not part of the `uri` field.
Also, the patch does not cover the situation of multiple spaces **before** the display, right?

rrb3942: You are correct that `<>` are not part of the uri field.

The `-1` actually stops us from trimming the `<` before the `uri`. I believe a header with a display name, but a non-enclosed `uri` is against spec (https://www.ietf.org/rfc/rfc2822.txt), so it shouldn't be a problem?

We remove everything from the start of the display name up to, but not including the `<`. What is left are the spaces from before the display name and the `<>` enclosed `uri`.

The patch **does not** cover multiple spaces **before** the display, but it could be modified to include those as well.

Example behaviour (verified against a test):

Original:
`From: "Name" <sip:user@domain>;tag=as2ed502bc`
Stripped:
`From: <sip:user@domain>;tag=as2ed502bc`

PR #1743, opened 2019-06-25
Add {ip.isip4} and {ip.isip6} convenience transformations

rrb3942:
Transformations for checking if a string is a valid ipv4 or ipv6 address independently.

Had some instances where I was only concerned if a string was a valid ipv4 address and thought $(var(maybeip){ip.isip4}) was a bit nicer than $(var(maybeip){ip.pton}{ip.family}) == "INET"


bogdan-iancu:
Thank you @rrb3942

PR #1744, opened 2019-06-25
Only report status if in forking daemon mode

rrb3942:
I've been experimenting running opensips with the -F option under systemd.

Everything seemed to run fine, but some errors where being logged on startup ("failed to send 0 status code"). This fixes the errors.

My testing has been under 2.4.x. I know that 3.x changed a bunch of stuff with auto-scaling, but I think this still applies.


bogdan-iancu:
Thanks @rrb3942

PR #1805, opened 2019-08-16
Fix #1804 - mid_registrar_save() should save new contact from registe…

### LINKED ISSUE #1804: [BUG] mid_registrar not forwarding REGISTER after UNREGISTER

```
version: opensips 3.0.0 (x86_64/linux)
git revision: 10d048223
main.c compiled on  with gcc 6.3.0
```

**Describe the bug**

Using usrloc `sql_write_mode` in `write-back` mode with  contact throttling and the default matching_mode `CONTACT ONLY`

When a device UNREGISTERS `mid_registrar_save()` returns `1` that the request is to be forwarded to main registrar. Then, when the same device re-REGISTERS immediately, `mid_registrar_save()` returns `2` that the request is to be absorbed. Leaving the contact unregistered with the main registrar until the next outgoing expire is reached, which could be a very long time.

**To Reproduce**

### With the following module params
```
# restart_persistency to `load-from-sql` so  we default to SQL write back
modparam("usrloc", "restart_persistency", "load-from-sql")
modparam("mid_registrar", "mode", 1)
modparam("mid_registrar", "outgoing_expires", 7200)
```

**Expected behavior**

On the REGISTER request following the unregister `mid_registrar_save()` should return `1`

ar45:
…r following an unregister.

When using `sql_mode` is set to `SQL_WRITE_BACK` and  `rr_persist` is `RRP_LOAD_FROM_SQL` contacts are not
immediately deleted from memory, contacts are marked `expires = UL_EXPIRED_TIME`. So when a `REGISTER` request
is received following an UNREGISTER, do not return the previous contact marked for deletion.

Returning the previous contact record, could have potentially lead to a crash, if opensips would free the contact
record while mid_registrar would retreive that using `get_ucontact()`.

### DIFF
--- a/modules/usrloc/urecord.c
+++ b/modules/usrloc/urecord.c
@@ -864,7 +864,7 @@ int delete_ucontact(urecord_t* _r, struct ucontact* _c, char is_replicated)
 static inline struct ucontact* contact_match( ucontact_t* ptr, str* _c)
 {
 	while(ptr) {
-		if ((_c->len == ptr->c.len) && !memcmp(_c->s, ptr->c.s, _c->len)) {
+		if ((_c->len == ptr->c.len) && !memcmp(_c->s, ptr->c.s, _c->len) && ptr->expires != UL_EXPIRED_TIME ) {
 			return ptr;
 		}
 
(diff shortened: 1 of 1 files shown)


ar45:
Looks like the build failures are unrelated.

ar45:
Also can this be back-ported to 3.0 ?

ar45:
Not stale

liviuchircu:
Hey Aron!

Sorry for the long wait on this one.  I've reviewed the fix and it seems perfectly valid: we should not attempt to renew contacts marked with `UL_EXPIRED_TIME`, rather we just create new ones, if required.

Also, regarding your last comment ("Returning the previous contact record, could have potentially lead to a crash"), I could only spot a single function which had this issue: `overwrite_req_contacts()`, as it was releasing the AoR lock too early - I will fix this in a separate commit.  Do let me know if you found any other racy `get_ucontact()` API calls.

PR #1820, opened 2019-09-09
fix a b2b_logic bug

hance2010:


bogdan-iancu:
Indeed, this patch is not valid for 3.x where the value of the parameters is automatically evaluated (as variables) by the core itself. So the module gets directly the final string value of the params.
See the correct fix here -> 7bb74db

I checked and 2.4 is fine as it is, not need for patching.

PR #1832, opened 2019-09-19
[topology_hiding] Fix encoding $TH_callee_callid and small cleanup

vasilevalex:
Corrected that after PR #1448 pseudo-variable $TH_callee_callid were still using base64 encoding instead of word64.


vasilevalex:
Please review

bogdan-iancu:
Thank you @vasilevalex

PR #1834, opened 2019-09-21
rtpengine: Use standart ALL_ROUTES macro

lemenkov:
ALL_ROUTES macro was introduced in commit
85413735093b55f64390c0284ec4c2f63fad3526. Let's use it instead of
hand-made non-standart ANY_ROUTE.


lemenkov:
This failure seems to be unrelated.

bogdan-iancu:
Thanks @lemenkov

PR #1871, opened 2019-10-25
cdb_add_ct_update creates a SIP URI based hash

Ozzyboshi:
Refering to my last commint 5d1678a37618d697c7c9a6346c0846d01404a142 I noticed  sometimes contacts get duplicated in my mongodb objects.
Investigating, I soon found out that some User Agents clients, like Zoiper softphone, append an "rinstance" field to the contact SIP header by default.
Since the cdb_add_ct_update() function calculates his base64 hash based on the whole contact field (including rinstance) and since rinstance changes on every registration dialog, I end up having multiple entries.
I solved this issue calculating the base64 hash using the SIP URI part of the contact SIP hdr and since this information is not available inside const ucontact_t *ct I used the core function parse_contacts() to get it.
I dont know if there are better ways to do this, if yes please let me know and I will submit another PL.
If you merge this commit into the master branch, please include it also in the 2.4 branch since I am still  following opensips 2.X series in production.

### DIFF
--- a/modules/usrloc/urecord.c
+++ b/modules/usrloc/urecord.c
@@ -39,6 +39,7 @@
 #include "../../ut.h"
 #include "../../hash_func.h"
 #include "../../db/db_insertq.h"
+#include "../../parser/contact/parse_contact.h"
 #include "ul_mod.h"
 #include "utime.h"
 #include "ul_callback.h"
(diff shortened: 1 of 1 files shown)


Ozzyboshi:
fauly implementation please do not merge

PR #1892, opened 2019-11-11
presence/clustering: Keep presence state on all nodes without shared DB

wdoekes:
Add the option to set cluster_federation_mode = 2, which does the
following:

- All PUBLISH/presence state is recorded on the presence node, even when
  no one is subscribed. PUBLISH is now accepted on any node for updated
  state.

This means that you can round-robin your PUBLISH instead of having to
choose a single presence endpoint. And it also means you can switch to
other nodes and get up-to-date state from there, when a node dies.

This is achieved without using a shared DB (which has its own
limitations obviously).

### DIFF
--- a/modules/presence/README
+++ b/modules/presence/README
@@ -228,10 +228,23 @@ modparam("presence", "cluster_id", 2)
 1.4.4. cluster_federation_mode (int)
 
    When enabling the federation mode, the node inside the presence
-   cluster will start partitioning the data (each node will keep
-   its own data set) and exchange the minimimum of needed
-   information via the clustering support. A non-zero value means
-   enabled.
+   cluster will start broadcasting the data to other nodes in the
+   cluster via the clustering support. Each node will keep its own
+   data set. A non-zero value means enabled.
+
+   Possible values for mode:
+     * 0 (default) - no federation/partitioning.
+     * 1 - the minimum needed information is broadcast/kept.
+     * 2 - published state is kept on all presence nodes on all
+       presence nodes, even when there aren't any subscribers.
+
+   If you don't want to use a shared database using fallback2db,
+   but still want a complete data set everywhere, you may choose
+   mode 2.
+
+   This mode allows you to switch PUBLISH endpoints, even for
+   already published Event States, thus allowing you to add and
+   remove presence servers without losing state.
 
    For more on presence clustering see the Section 1.2, “Presence
    clustering” chapter.
--- a/modules/presence/clustering.c
+++ b/modules/presence/clustering.c
@@ -383,7 +383,9 @@ static void handle_replicated_publish(bin_packet_t *packet)
 	delete_cluster_query( &s, ev.parsed, hash_code);
 	lock_release( &pres_htable[hash_code].lock );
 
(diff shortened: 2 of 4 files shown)


wdoekes:
Hi!

As we're using pua/pua_dialoginfo to create state, we'll just want to select a random presence endpoint, and not stick to a fixed one.

Additionally, we want to have the state available on all nodes, so we can resubscribe to a different node and get all state from there.

I think this changeset achieves that. Let me know what you think (and if accepted, perhaps backport to 2.4).

bogdan-iancu:
Hi @wdoekes , thank you for your contribution.
If I understand correctly, the desired effect (in the new clustering mode) is to replicate the entire presentity (published) data over all the node - each node will hold the full set of presentity data. If so, this looks like an alternative to the current DB sharing, but without DB, right ?
Assuming I'm on the right tracks, there is something missing here - when a node comes up, it should do a full sync (fetch the whole data set) with a node from the cluster.

BTW, how does this approach scale in term of memory usage (as all presentities are stored in memory, there is no data partitioning) and traffic (as each publish is replicated to all the other nodes) ? also, for this mode, do you use DB persistence at all ?

wdoekes:
Hi bogdan! Sorry for the late reply. Here goes.

> each node will hold the full set of presentity data

Yes

> If so, this looks like an alternative to the current DB sharing, but without DB, right ?

Correct.

I have very poor experience with asynchronous (mysql) DB replication when it comes to opensips (location, presence) data. And having two modes of replication (mysql AND opensips) running for different parts of the same dataset, makes debugging harder.

So yes. Doing replication through opensips _only_ makes sense to us.

> also, for this mode, do you use DB persistence at all ?

Actually, not really. We assume that the presence data is ephemeral. The only reasons we have a DB is:

- either the config/module mandates it;
- for semi-persistent data over quick restarts;
- for easier access to semi-realtime data for debugging.

> When a node comes up, it should do a full sync (fetch the whole data set) with a node from the cluster.

Yes. That would indeed be beneficial. But right now, we were thinking we could let a new node warm up for a couple of hours before adding it to the dispatcher pool.

We assume that data older than 4 hours is old/expired anyway and that required/wanted data (subscriptions, events) will have been resubmitted by that time.

So yes, on-startup replication might be nice, as long as the donor node doesn't suffer. But as there are no facilities for this yet, this is not a must.

> BTW, how does this approach scale in term of memory usage (as all presentities are stored in memory, there is no data partitioning) and traffic (as each publish is replicated to all the other nodes)

For our purposes, we're not running into memory problems yet. We are concerned about CPU usage and availability. So we hope to increase availability by adding multiple nodes and being able to round robin over them.

If we do run into memory/traffic issues, we would probably solve this by manually sharding the data instead. We have many separate "islands" of presence users, where those islands never need any info from any others.

bogdan-iancu:
Thank you @wdoekes for the reply. The only thing I do not fully agree is about restoring (or lack of restoring) the presence data upon restart. Even if the presentity data is ephemeral, it should be restore - we do the same for registration data, which is also ephemeral ;).
If you do not have the skills, we can take on our side to do the initial SYNC from the cluster.

PR #1902, opened 2019-11-22
Memory leak fix for mongodb module-reply not freed

Ozzyboshi:
mongoc_collection_command_simple() takes &reply as output parameters but probably it needs to be freed.
From the mongodb documentation at http://mongoc.org/libmongoc/1.14.0/mongoc_collection_command_simple.html

```
#include <bson/bson.h>
#include <mongoc/mongoc.h>
#include <stdio.h>

static void
print_collection_stats (mongoc_collection_t *collection)
{
   ...
   bson_t reply;

   ....

   if (mongoc_collection_command_simple (
          collection, cmd, NULL, &reply, &error)) {
          ...
   }
   ...

   bson_destroy (&reply);
   bson_destroy (cmd);
}
```

### DIFF
--- a/modules/cachedb_mongodb/cachedb_mongodb_dbase.c
+++ b/modules/cachedb_mongodb/cachedb_mongodb_dbase.c
@@ -1008,6 +1008,7 @@ int mongo_con_add(cachedb_con *con, str *attr, int val, int expires, int *new_va
 	}
 
 out:
+	bson_destroy(&reply);
 	bson_destroy(cmd);
 	return ret;
 }


liviuchircu:
hey, @Ozzyboshi !  That's definitely a good catch, with a good fix.  But isn't the `mongo_con_raw_query()` function suffering from the same type of leak?  If so, could you please redo the pull request so it fixes both leaks and also get rid of the "stupid commit" in the process?  Many thanks!

Ozzyboshi:
I am going to investigate mongo_con_raw_query today, then i will close this pl and submit another one that addresses the 2 issues

Ozzyboshi:
@liviuchircu I think you were right, mongo_con_raw_query also leaks for the same reason, i fixed it however i was not able to test it because i dont know how to make opensips call cachedb_raw_query in cachedb/cachedb.c.
I see the only way to invoke this function is from do_action() in action action.c but I don't know much about the opensips internals.
I must find a way to pass a CACHE_RAW_QUERY_T action at do_action(), do you know a quick way to do that? do you have any hint?

Ozzyboshi:
i figured out on my own i had just to put something in the opensips.cfg file as documented here
https://opensips.org/html/docs/modules/1.9.x/cachedb_mongodb.html#id293499

Ozzyboshi:
I run raw mongo queries triggering mongo_con_raw_query() as you suggested, I expected some memory leaks but surprisingly there are none according to valgrind.

This is very odd because mongo_con_add() has similar code and according to valgrind it leaks, this is a sample from my valgrind logfile that proves it

```
==1506== 512 bytes in 2 blocks are definitely lost in loss record 1,431 of 1,600
==1506==    at 0x4C2BBAF: malloc (vg_replace_malloc.c:299)
==1506==    by 0xE3649CE: bson_malloc (in /usr/local/lib/libbson-1.0.so.0.0.0)
==1506==    by 0xE357539: bson_copy_to (in /usr/local/lib/libbson-1.0.so.0.0.0)
==1506==    by 0xE0D8C69: mongoc_cluster_run_opmsg (in /usr/local/lib/libmongoc-1.0.so.0.0.0)
==1506==    by 0xE0D8E7B: mongoc_cluster_run_command_monitored (in /usr/local/lib/libmongoc-1.0.so.0.0.0)
==1506==    by 0xE0D6205: _mongoc_client_command_with_opts (in /usr/local/lib/libmongoc-1.0.so.0.0.0) src/libmongoc/src/mong
==1506==    by 0xE0DE5A6: mongoc_collection_command_simple (in /usr/local/lib/libmongoc-1.0.so.0.0.0)
==1506==    by 0xDC94AE5: mongo_con_add (cachedb_mongodb_dbase.c:980)
==1506==    by 0x132C04C7: link_dlg_profile (dlg_profile.c:872)
==1506==    by 0x132C04C7: set_dlg_profile (dlg_profile.c:922)
==1506==    by 0x132FA170: read_dialog_profiles (dlg_db_handler.c:467)
==1506==    by 0x133004CB: load_dialog_info_from_db (dlg_db_handler.c:638)
==1506==    by 0x133004CB: init_dlg_db (dlg_db_handler.c:203)
==1506==    by 0x132F4F47: mod_init (dialog.c:991)
```
At this point I would suggest to leave mongo_con_raw_query unmodified and just insert my bson_destroy(&reply) addition in mongo_con_add() in the main branch and backport it to 2.4.6 series.
BTW I added bson_destroy (&rpl) in mongo_con_raw_query anyway to see what happens: opensips does not crash, actually it behaves exactly like before, it seems that this call is irrelevant.

@liviuchircu did you analyze this issue? what do you think?

Ozzyboshi:
wrong commit deleted

liviuchircu:
Thank you for the fix and investigation, @Ozzyboshi .  I've made a mental note to go over the leak together with #1903.  We're done with this one!

PR #1917, opened 2019-12-17
nat_traversal: default to not saving a keep-alive file

saghul:
Typically OpenSIPS will be installed in /usr/(s)bin/ and the OpenSIPS
process will have no rights to store it there. In addition, this allows
for not having a keep-alive file at all, since the code is ready to
handle that.

### DIFF
--- a/modules/nat_traversal/README
+++ b/modules/nat_traversal/README
@@ -469,7 +469,7 @@ PS\r\nX-MyHeader: some_value\r\n")
    which case it will store it in the OpenSIPS working directory,
    or an absolute path.
 
-   Default value is undefined “keepalive_state”.
+   Default value is NULL, no keep-alive file will be kept.
 
    Example 1.5. Setting the keepalive_state_file parameter
 ...
--- a/modules/nat_traversal/nat_traversal.c
+++ b/modules/nat_traversal/nat_traversal.c
@@ -201,7 +201,7 @@ static Bool keepalive_disabled = False;
 
 static unsigned int keepalive_interval = 60;
 
-static char *keepalive_state_file = "keepalive_state";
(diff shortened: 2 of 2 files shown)


saghul:
@danpascu Can you please take a look? I haven't tested the change but it seems like it does the right thing by looking at the code closely.

Context: I want to disable the keep-alive file functionality since I don't need it and it looks like the code accounts for that, but the default is just "keepalive_file".

danpascu:
Your assessment is incorrect. The state file is not written to the binary's directory but to the runtime directory (/run/opensips) which should be writable unless opensips was not installed correctly.

As for the requested feature (not save the state file), it is already possible, by specifying the state file to be /dev/null

Regarding changing the default behavior, I do not agree with that for multiple reasons:

1. It reduces the functionality for the default setup for no good reason.
2. It breaks backward compatibility
3. It forces every user to go and update the configuration file in order to restore the functionality they had before.
4. It is the first time this feature was requested, so I'll assume the people that don't want to save the state file are in the minority.

saghul:
🤦‍♀️ Using /dev/null is absolutely right, of course!

As for the default value, I had a minimal config and it complained it wasn’t writeable, I’ll check again when I have the chance.

Thanks a lot the the insight!

danpascu:
I also use the minimal config for this and it is written to /run/opensips/keepalive_state

Depending on how your installation is done (systemd/init), /run/opensips might be owned by root and opensips runs as opensips:opensips which makes it unable to write any file in the runtime directory except the pid (which is written before it changes user to opensips).

saghul:
I installed the Debian package and used systemd to start it. Maybe that’s it indeed.

PR #1926, opened 2019-12-26
FIX: opensipsdbctl create schema - oracle11G 

zhouhailin:
you can use command 'opensipsdbctl  create' auto-creating schema

### DIFF
--- a/scripts/opensipsdbctl.oracle
+++ b/scripts/opensipsdbctl.oracle
@@ -12,7 +12,7 @@ DBSYSUSER="sys"
 
 
 # path to the database schemas
-DATA_DIR="/usr/local/opensips/share/opensips/"
+DATA_DIR="/usr/local/share/opensips/"
 if [ -d "$DATA_DIR/oracle" ]; then
 	DB_SCHEMA="$DATA_DIR/oracle"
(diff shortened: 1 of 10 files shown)


bogdan-iancu:
Thank you @zhouhailin for your patch.
Could please explain the nature of these changes:
* why changing the `DATA_DIR`? for all the other backends we are dumping the schema files into `/usr/local/share/opensips`
* why changing the default tablespace ?
* why changing the `grant` format?
* all the changes in the `oracel/*.sql` files ? Also note here that these are auto-generated files via `opensipsctl dbschema`, see https://github.com/OpenSIPS/opensips/blob/2.4/doc/dbschema/xsl/oracle.xsl

zhouhailin:
Thanks @bogdan-iancu 

- DATA_DIR changed, caused by install path is difference
- tablespace changed, caused by oracle 11g not support
- grant changed, caused by oracle 11g not support
- the command of `opensipsctl dbschema` is not found

I found that there are many bugs in the db_oracle module. Is there any plan to fix this module

bogdan-iancu:
Hi @zhouhailin , any reason for closing this PR ?

PR #1941, opened 2020-01-18
Build patches for Arch-Linux

rzerres:
Hey Team,

first: all the best for the new decade 2020! I'm eager to see all that progress to contiue.

# Abstract
Today I got a change to have a look at the master repo.
I did update the packaging stuff for Arch-Linux and attach the needed patches to make a
successfull build using latest Arch packages.

Kindly ask for review and merge to your master.

best regards
Ralf

# v1
* packaging: Arch-Linux updated PKGBUILD.git
* packaging: Arch-Linux update search-path for docbook.xsl
* packaging: Arch-Linux update Makefile.conf.template

# v2
* packaging: Arch-Linux new PKGBUILD
* updating dependency to standard lua  package (>= v5.1)
* updating dependency to bison and flex

### DIFF
--- a/Makefile.conf.template
+++ b/Makefile.conf.template
@@ -61,7 +61,14 @@
 #xmpp= Gateway between OpenSIPS and a jabber server. It enables the exchange of IMs between SIP clients and XMPP(jabber) clients. | parsing/building XML files, typically libexpat1-devel
 #uuid= UUID generator | uuid-dev
 
-exclude_modules?= aaa_radius b2b_logic cachedb_cassandra cachedb_couchbase cachedb_memcached cachedb_mongodb cachedb_redis carrierroute cgrates compression cpl_c db_berkeley db_http db_mysql db_oracle db_perlvdb db_postgres db_sqlite db_unixodbc dialplan emergency event_rabbitmq h350 httpd identity jabber json ldap lua mi_xmlrpc_ng mmgeoip osp perl pi_http presence presence_dialoginfo presence_mwi presence_xml proto_sctp proto_tls proto_wss pua pua_bla pua_dialoginfo pua_mi pua_usrloc pua_xmpp python regex rabbitmq rabbitmq_consumer rest_client rls siprec sngtc snmpstats tls_mgm uuid xcap xcap_client xml xmpp
+#exclude_modules?= aaa_radius b2b_logic cachedb_cassandra cachedb_couchbase cachedb_memcached cachedb_mongodb cachedb_redis carrierroute cgrates compression cpl_c db_berkeley db_http db_mysql db_oracle db_perlvdb db_postgres db_sqlite db_unixodbc dialplan emergency event_rabbitmq h350 httpd identity jabber json ldap lua mi_xmlrpc_ng mmgeoip osp perl pi_http presence presence_dialoginfo presence_mwi presence_xml proto_sctp proto_tls proto_wss pua pua_bla pua_dialoginfo pua_mi pua_usrloc pua_xmpp python regex rabbitmq rest_client rls siprec sngtc snmpstats tls_mgm xcap xcap_client xml xmpp
+###
+# Arch Linux specific
+###
+# Couchbase: needs to be build from aur
+# Oracle: not supported as open-source package
+# Sangoma: can't test Voice Transcoding with D-series Sangoma transcoding cards
+exclude_modules?= cachedb_couchbase db_oracle sngtc
 
 include_modules?=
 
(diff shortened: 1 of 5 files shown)


Makefile.conf.template
```diff
@@ -61,7 +61,14 @@
...
+#exclude_modules?= aaa_radius b2b_logic cachedb_cassandra cachedb_couchbase cachedb_memcached cachedb_mongodb cachedb_redis carrierroute cgrates compression cpl_c db_berkeley db_http db_mysql db_oracle db_perlvdb db_postgres db_sqlite db_unixodbc dialplan emergency event_rabbitmq h350 httpd identity jabber json ldap lua mi_xmlrpc_ng mmgeoip osp perl pi_http presence presence_dialoginfo presence_mwi presence_xml proto_sctp proto_tls proto_wss pua pua_bla pua_dialoginfo pua_mi pua_usrloc pua_xmpp python regex rabbitmq rest_client rls siprec sngtc snmpstats tls_mgm xcap xcap_client xml xmpp
+###
+# Arch Linux specific
+###
+# Couchbase: needs to be build from aur
+# Oracle: not supported as open-source package
+# Sangoma: can't test Voice Transcoding with D-series Sangoma transcoding cards
+exclude_modules?= cachedb_couchbase db_oracle sngtc
```
razvancrainea: I am not sure I understand why the exclude modules are added here; if these should be Arch Linux specific, we should simply overwrite the `exclude_modules` the Arch Linux `PKGBUILD`.

Makefile.conf.template
```diff
@@ -81,20 +88,20 @@ DEFS+= -DDBG_MALLOC #Include additional, debug-enabled allocator flavors
...
 #DEFS+= -DNOSMP #Do not use SMP compliant locking. Faster but won't work on SMP machines
 #DEFS+= -DEXTRA_DEBUG #Compile in some extra debugging code
-DEFS+= -DCC_O0 #Zero compiler optimizations (FAST compile, SLOW code. For devs)
+#DEFS+= -DCC_O0 #Zero compiler optimizations (FAST compile, SLOW code. For devs)
 #DEFS+= -DORACLE_USRLOC #Use Oracle compatible queries for USRLOC
 #DEFS+= -DSHM_EXTRA_STATS #Provide tools to get extra statistics for the shared memory used
 #DEFS+= -DUNIT_TESTS #Include unit testing code into opensips and modules
 
-PREFIX ?= /usr/local/
+PREFIX ?= /usr/
```
razvancrainea: Why change this here? This should be overwritten in the `PKGBUILD`.

razvancrainea: Moreover, the flags are configured of the `master` branch, which should be used **only** for development.

packaging/arch/PKGBUILD
```diff
@@ -0,0 +1,254 @@
...
+	fi
+
+	msg2 " ... examples"
+	DOC_DIR="$pkgdir/usr/share/doc/${pkgbase%%-*}"
+	if [ -d ./examples ]; then
+		mkdir -p "$DOC_DIR"
+		cp --recursive ./examples "$DOC_DIR"
+		chmod --recursive 0644 "$DOC_DIR"
+	fi
+}
```
razvancrainea: I see that there already is an official package for Arch Linux [here](https://www.archlinux.org/packages/community/x86_64/opensips/). Is this one different than that one? Or why do we need a different packaging, for the `git` repo?

packaging/arch/PKGBUILD.git
```diff
@@ -40,35 +44,39 @@ options=('!emptydirs' 'zipman' '!makeflags' 'docs')
...
-		echo "  -> patching: "
-		git am --signoff ../../patches-git/0000-packaging-support-arch-linux.patch
-		git am --signoff ../../patches-git/0001-osp-Fix-module-compilation.patch
-		git am --signoff ../../patches-git/0002-opensips_json_c_helper.h-fix-preprocessor-definition.patch
-		git am --signoff ../../patches-git/0003-Makefile.defs-adapt-default-settings-for-XSL-parser.patch
-		git am --signoff ../../patches-git/0004-doxygen-update-element-list.patch
-		echo "  -> patching [done]"
+		msg2 "patching:"
+		git am --signoff ../../patches-git/0001-packaging-Arch-Linux-update-search-path-for-docbook.patch
+		#git am --signoff ../../patches-git/0002-packaging-Arch-Linux-update-PKGBUILD.git.patch
```
razvancrainea: These will probably not be needed, since we will be pushing the patches upstream.

Makefile.conf.template
rzerres: It is Arch specific, since i can't make the modules compile (cachedb_couchbase), or they are not available (like od_oracle, sngtc).

Makefile.conf.template
rzerres: You are right, PREFIX change can be omitted. 

What flags should be used instead if targeting a production environment??

PR #1965, opened 2020-02-08
Fix issues found while compiling with GCC 10 (-fno-common)

lemenkov:
There is a more strict check about global variables definitions in header flags w/o extern keyword in  GCC 10.

* https://gcc.gnu.org/gcc-10/porting_to.html#common

It let us to catch a few potential issues (missing typedef keyword, globals declated in header files).

### DIFF
--- a/mem/hp_malloc.c
+++ b/mem/hp_malloc.c
@@ -116,9 +116,6 @@ int mem_warming_enabled;
 int mem_warming_percentage = MEM_WARMING_DEFAULT_PERCENTAGE;
 
 #if defined(HP_MALLOC)
-stat_var *rpm_used;
-stat_var *rpm_rused;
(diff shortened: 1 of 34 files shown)


liviuchircu:
Thank you for the initiative, @lemenkov !  Everything was okay, except some tiny `tmb` global structure you forgot to specify, so `dispatcher` wasn't starting due to missing symbol error from the loader.

Allso, I noticed that you were doing some redundant global initializations to `0` or `NULL`, see 1632ff2d516.  As per section 6.7.9.10 in the [C11 standard](http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1570.pdf) (or maybe even an earlier one!), these are always guaranteed to be zeroed, including arrays.

PR #1966, opened 2020-02-09
rtpengine: Remove unused struct fields

lemenkov:
These two fields were copied directly from rtpproxy module but never
used by this module.


liviuchircu:
Good catch!

PR #1975, opened 2020-02-17
rtpengine: Move RTPE_IO_ERROR_CLOSE definition up

lemenkov:
Move RTPE_IO_ERROR_CLOSE macro up to the rest of macros. Also don't
undefine it later - we might consider using it.


lemenkov:
Right now it's a cosmetic change but the functionality I'm going to add later will rely on this macro.

PR #1995, opened 2020-03-08
Remove dead code

l2dy:
`while (hdr)` without `break` ensures that `hdr` will always be `NULL` afterwards, skipping the second `while`.


liviuchircu:
Good catch!  The author most likely was under the impression that they were wrapping up a `do {} while` block ;)

PR #1997, opened 2020-03-09
Fix time comparison

l2dy:
`?:` has lower precedence than `+`.

Fix copied from line 1133.


liviuchircu:
Nice!  What static analysis tool are you using, if I may ask? :)

l2dy:
> Nice! What static analysis tool are you using, if I may ask? :)

I'm using PVS-Studio, they offer [free personal keys](https://www.viva64.com/en/b/0600/) to open source developers with the following limitation:

> The key is individual and can only be used to check open source projects published on GitHub/GitLab/Bitbucket. Free license doesn't extend to projects' mirrors.

PR #1999, opened 2020-03-09
mi_fifo: fix NULL pointer dereference

l2dy:
Trying to print `file` when it's NULL is a null pointer dereference.


liviuchircu:
Interesting report from PVS-Studio, because `printf("%.*s\n", 42, NULL);` is not a crash, it just outputs `(nil)`.  But, let's make it happy! :)

l2dy:
Other compilers may not like it, and it's a better error message anyway. 😄

PR #2001, opened 2020-03-09
usrloc: fix typos

l2dy:


liviuchircu:
Great!  Luckily for me, these typos were harmless, as this code pretty much never fails on startup.  However, luck can only get you so far!

PR #2008, opened 2020-03-12
Fix RFC3261 token check

l2dy:
Excerpt from https://tools.ietf.org/rfc/rfc3261.txt:

```
token       =  1*(alphanum / "-" / "." / "!" / "%" / "*"
               / "_" / "+" / "`" / "'" / "~" )
```


liviuchircu:
In short, this change:

* removes the duplicate `+` check (harmless, very likely to be optimized by the compiler)
* adds support for the following valid token characters, according to RFC 3261: `'`, \` (backtick)  and `!` (apparently harmless)

To prevent any unwanted surprises, I suggest to only merge this to `master` branch, without backporting it to any stable release branch.  I haven't seen any complaints on the matter, so I don't think it's a big issue if we still don't let `2.4` and `3.0` users to have their `INVITE!!` requests properly parsed :)  Do you agree?

l2dy:
> Still, to prevent any unwanted surprises, I suggest to only merge this to `master` branch, without backporting it to any stable release branch. I haven't seen any complaints on the matter, so I don't think it's a big issue if we still don't let `2.4` and `3.0` users to have their `INVITE!!` requests properly parsed :) Do you agree?

Yes, it's a breaking change in behavior, so we should be more careful.

PR #2020, opened 2020-03-12
dialplan: fix insert error condition

l2dy:


liviuchircu:
Luckily, the segfault can never happen with the current state of the code.  Good catch!

PR #2022, opened 2020-03-12
usrloc: fix malloc size

l2dy:
Memory is linear :)


liviuchircu:
Cool bug :) Luckily, `A * B > A + B, if A > 1 and B > 1`, so the logic worked as intended.

PR #2023, opened 2020-03-12
stats: fix loop variable

l2dy:
I hope the sharing of loop variables is not intentional.


liviuchircu:
What a crazy bug!  Although, strangely enough, I think the logic was working fine even before, since fetching the `"all"` statistic special keyword would fully populate the JSON with all statistics, to the point where even if you were to ask for additional ones, e.g. `get_statistics all rcv_requests, drop_requests`, the extra ones wouldn't get included (neither before or after your fix), since JSON object keys are unique!

PR #2027, opened 2020-03-12
msilo: fix excessive check

l2dy:
`msg.len` is checked twice. Should we allow an empty `msg` here?


liviuchircu:
According to [this](https://github.com/OpenSIPS/opensips/blob/master/modules/msilo/msilo.c#L608) code, the answer is "no".

PR #2028, opened 2020-03-12
osp: fix use of uninitialized pointer

l2dy:
See https://github.com/OpenSIPS/opensips/blob/47d9ec972118d3fa9ff0a0e134fb0b013f623a87/modules/osp/orig_transaction.c#L840-L842


l2dy:
Maybe `OSPT_CALL_ID* callids[callidnumber] = {NULL};` is better?

liviuchircu:
This code is completely bogus.  See line +737, it will crash anyway if we set the pointer to `NULL` and not exit on error.  We can merge this, but it really fixes nothing.

liviuchircu:
No, I was wrong, there's a big `else` branch and my assessment was wrong.  Merging.

l2dy:
> This code is completely bogus. See line +737, it will crash anyway if we set the pointer to `NULL` and not exit on error. We can merge this, but it really fixes nothing.

Line 737 is in the `else {}` block, so it would not be executed if ospGetCallId() failed, correct?

PR #2033, opened 2020-03-14
auth: return -1 on error in auth_get_ha1()

l2dy:
Revert #2024, which ignored "Username not found" cases.

### DIFF
--- a/modules/auth/auth_mod.c
+++ b/modules/auth/auth_mod.c
@@ -442,7 +442,7 @@ static inline int pv_authorize(struct sip_msg* msg, str *domain,
 	cred = (auth_body_t*)h->parsed;
 
 	res = auth_get_ha1(msg, &cred->digest.username, domain, ha1);
-	if (res != 0) {
+	if (res < 0) {
 		/* Error */
 		if (sigb.reply(msg, 500, &auth_500_err, NULL) == -1) {
 			LM_ERR("failed to send 500 reply\n");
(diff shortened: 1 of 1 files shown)


liviuchircu:
@l2dy we kinda bungled this on the initial merge, as I completely missed the 2nd `if` statement after calling `auth_get_ha1()`!  There were still some adjustments to be made even after your commits, see my comments in the merge commit.  Cheers!

l2dy:
> @l2dy we kinda bungled this on the initial merge, as I completely missed the 2nd `if` statement after calling `auth_get_ha1()`! There were still some adjustments to be made even after your commits, see my comments in the merge commit. Cheers!

@liviuchircu Thanks. But the error conditions in `auth_get_ha1()` in your merge commit is different. Does `pv_get_spec_value()` return 0 on error or "not found"?

liviuchircu:
The main idea is that we want to return "not found" (retcode `1`) if:

* `auth_get_ha1()` returns non-zero (PV is not set)
* `auth_get_ha1()` return zero, but the value does not match the `;username=value` from the `WWW-Authenticate` header field

l2dy:
Thanks for the explanation! These functions really need some comments to explain their return values.

liviuchircu:
Agreed.  There is still some weird logic in there / room for perfection.  For example:  why does `pv_authorize()` return a "username not found" error if the script writer forgot to set the `$var(password)` variable?  It should be "password not found"...

PR #2037, opened 2020-03-16
tracer: fix null pointer dereference

l2dy:
Not sure if this is the correct fix, maybe line 3273 also got it wrong.

https://github.com/OpenSIPS/opensips/blob/590a19f11a7afa3fb4d60792be6886fd59183558/modules/tracer/tracer.c#L3273

### DIFF
--- a/modules/tracer/tracer.c
+++ b/modules/tracer/tracer.c
@@ -3127,7 +3127,7 @@ trace_dest get_next_trace_dest(trace_dest last_dest, int hash)
 
 	for (instance = info->instances; instance; instance = instance->next) {
 		for (it=instance->trace_list; it && it->hash == hash; it=it->next) {
-			if (it->type == TYPE_HEP && (it->traceable || !(*it->traceable))) {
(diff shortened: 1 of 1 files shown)


l2dy:
@liviuchircu Can you help review the remaining PRs?

bogdan-iancu:
Hey @l2dy , looking at a place where the negative condition is user (see https://github.com/OpenSIPS/opensips/blob/master/modules/tracer/tracer.c#L3268), I would say your fix is the right one.
The line 3273 has the negative test is done IMHO, so it should be fine (it wants to see if  NOT traceable , so the pointer must be valid and hold a negative value).

If you agree with me, I can apply the PR.

l2dy:
Does it make sense that if `traceable` is NULL `it` is considered traceable? If so, I think this PR is good to merge.

bogdan-iancu:
Based on the logic from the rest of the code, I would assume yes. Anyhow, checking the code, a NULL `traceble` is impossible, as it is allocated together with the `trace_instance`.

PR #2047, opened 2020-03-17
dialog: fix possible overflow

l2dy:


l2dy:
Overflow is probably not possible because `val < 8*sizeof(unsigned int)`. But it simplifies the expression anyway.

PR #2065, opened 2020-04-03
[Performance] Reduce Context Switching

rrb3942:
These are changes that I am forward porting from our internal 2.4 branch that greatly reduces context switching under Linux. It uses EPOLLSXCLUSIVE when available and makes some changes to how events are handled before a process goes back to waiting.

Changes are mostly around UDP with some changes around internal pipes (IPS, timer) as they get the largest benefit. TCP is left untouched. Some of these changes could also be applied to the TCP listening socket, but I don't know how much of a benefit that would make (we only run UDP).

Not fully tested under master other than it compiles and OpenSIPS seems to start fine.

Under similar loads these changes on 2.4 brought context switching (as reported by vmstat) down to 25k/sec from around 45k/sec. CPU Usage as reported by our hypervisor dropped by 5-10% under light load and by nearly 20% under heavier loads.

It takes some care to avoid the pitfalls of using EPOLLEXCLUSIVE, and we have found that it works well in production.

Mostly just wanted to get this out so that some of the ideas might be incorporated before the next LTS.

Let me know what you think.

### DIFF
--- a/io_wait.h
+++ b/io_wait.h
@@ -347,7 +347,8 @@ inline static int io_watch_add(	io_wait_h* h, // lgtm [cpp/use-of-goto]
 								void* data,
 								int prio,
 								unsigned int timeout,
-								int flags)
+								int flags,
+								int exclusive)
 {
 
 	/* helper macros */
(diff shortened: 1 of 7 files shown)


wdoekes:
This looks promising.

bogdan-iancu:
I does, I was looking to get more time to read in details on the topic and to try to put together some basic testing, as validate the change

PR #2073, opened 2020-04-12
rtpengine: Mark rtpengine as busy when sending-receiving data

lemenkov:
This doesn't change anything in synchronous mode but acts as a guard in
async mode preventing us from reading others' replies.

It doesn't use any lock_get/lock_release just because it shouldn't
block or wait, quite the contrary - check quickly and try the next one
if busy.

### DIFF
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -45,6 +45,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdbool.h>
 #include <unistd.h>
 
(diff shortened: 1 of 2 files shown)


bogdan-iancu:
Thanks @lemenkov , it looks good to me, @razvancrainea could you double check it?

razvancrainea:
I'd argue this is not correct, for example if you're getting a reply, we should use the same RTPEngine as was used for the request. However, if that engine is busy, a different node would be chosen, leading to an inconsistent state.
Perhaps a better approach would be to use a different connection.

PR #2074, opened 2020-04-13
Add support for upcoming json-c 0.14.0.

besser82:
TRUE/FALSE are not defined anymore.  1 and 0 are used instead.


bogdan-iancu:
Thank you @besser82  for your contribution. I also backported this patch to 2.4 and 3.0 versions.

PR #2077, opened 2020-04-14
Add sctp multihoming support

russagit:
This is a follow up for https://opensips.org/pipermail/users/2019-March/040715.html

### DIFF
--- a/cfg.lex
+++ b/cfg.lex
@@ -212,6 +212,7 @@ LOGSTDERROR	log_stderror
 LOGFACILITY	log_facility
 LOGNAME		log_name
 LISTEN		listen
+LISTEN_SCTP_SEC		listen_sctp_sec
 MEMGROUP	mem-group
 ALIAS		alias
 AUTO_ALIASES	auto_aliases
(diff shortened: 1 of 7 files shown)


bogdan-iancu:
Hi @russagit , I'm trying to understand the idea behind the 'Add sctp multihoming support ' commit. If I get it right, via the `LISTEN_SCTP_SEC` you define only one additional listener for sctp, right ? (btw, multiple definition of that parameter will leak some sockets, only the last being kept and used).

Further, this single extra SCTP socket will be used for **ALL** the defined SCTP listner, right ?

russagit:
Hi @bogdan-iancu,
Yes, you are right in everything. Since this is my first try with opensips it is clumsy and limited but covers our needs in production at the moment.
I would be glad if someone could modify this to be more universal.

bogdan-iancu:
@russagit , I understand your point, but in this current format, this PR cannot be accepted - are too many things which are not ok in this approach.
To be a valid PR, it should have the extra socket definition as part of the listener definition "listen=sctp:x.y.z.w:1234 extra a.b.c.d:1234"; in this way, the additional sockets are per listener, not global.

russagit:
Sorry, I have to add that sctp_bindx does not create extra SCTP socket, it just adds extra IP (secondary IP address without port) to existing socket.

bogdan-iancu:
Yes, I understand that.
And let's keep this open, maybe someone will find it a good starting point to work out this feature.

russagit:
Hi @bogdan-iancu,
Could you take a look at another try: https://github.com/SIfoxDevTeam/opensips/commit/3b4899ae25aff172bc358ad31d123146244a3432
New token sctp_sec_addr added, example of defining sctp with secondary ip address:
listen=sctp:192.168.1.1:5060 sctp_sec_addr 192.168.2.1

russagit:
Hi @bogdan-iancu,
Could you take a look at https://github.com/SIfoxDevTeam/opensips/commit/3b4899ae25aff172bc358ad31d123146244a3432

kgalinurov:
Any progress here?

bogdan-iancu:
Last attempt here to revive this ticket

Issue #2083, opened 2020-04-22
[QUESTION] force_send_socket socket selection


attermann:
**OpenSIPS version you are running**
```
version: opensips 2.4.7 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 495b980
main.c compiled on 22:17:41 Apr  7 2020 with gcc 5.4.0
Running on Ubuntu 16.04
```

Not certain if this is a bug per-se, a limitation, or something else entirely.  My proxy/registrar is configured to listen on multiple ports of the same interface, and I use these ports in my config to detect whether a request or response originates internally or externally, and whether or not transcoding should be configured.

To be specific, I use port 5060 to communicate with external UAs and 6050 to communicate with internal UAs *without* transocding, and port 7080 to communicate with external UAs and 8070 to communicate with internal UAs *with* transocding.  I am using force_send_socket() to choose which port a request or response goes out on.  Most traffic is TCP.

Recently I have discovered that sometimes when a particular server registers clients with opensips, two separate applications running on that server (both registering to opensips) can use the same TCP port when they are connecting to different ports on opensips (since these are client connections which can re-use a port as long the destination ip:port is different).

So for example, assuming my opensips is listening on tcp:1.1.1.1:5060 and tcp:1.1.1.1:7080, the server might register client A from tcp:2.2.2.2:1234 to opensips at tcp:1.1.1.1:5060, and client B from tcp:2.2.2.2:1234 (same source protocol:ip:port) to opensips at tcp:1.1.1.1:7080 (same destination protocol:ip but different port).

This is all fine and good so far, but when it comes time to send a call to client B my script would use force_send_socket(tcp:1.1.1.1:7080) to send from port 7080 to the server at tcp:2.2.2.2:1234, but the application that registered client A would actually receive the call instead.  I don't know if force_send_socket() is not selecting the correct socket or if the issue is something different but I wanted to post my question here in case anybody that is familiar with the internal workings of force_send_socket() may know what's happening here.

Thanks in advance,

Chad.

[...]

PR #2090, opened 2020-04-25
Fix for TCP socket selection

attermann:
Issue #2083:

Fix for TCP socket selection in tcp_conn_get() when client connections to opensips share the same ephemeral port on the client side.  This can happen when opensips is listening on more than one port, and a given client host/ip connects to that opensips on more than one of its listening ports.  Under these circumstances the client TCP stack will allow client connections to share a local port as long as the destination ip:port are different.

In opensips, when searching for a TCP socket to send over, the general search algorithm in tcp_conn_get only considers the destination ip:port and not the source ip:port.  What can happen as a result is, for example, if the same client host/ip registers two UAs on different opensips listening ports, when it comes time to send a call to one of the UAs, opensips will randomly send to the UA using the first TCP connection that it finds based on the destination ip:port search which will be the wrong connection half the time.

This patch causes "send_sock" to be passed to tcp_conn_get() from all of the modules that call it which contains the desired local ip:port to use.  This is used in the search algorithm in tcp_conn_get to prefer connections that match both source and destination ip:port first, and falls back to just destination ip:port matching as before if a more specific match can not be made.
Note that this search algorithm was implemented using an additional pass for the primary source and destination ip:port match.  This could be optimized by a more sophisticated check for both cases in a single pass but that would involve much care with locks and reference counters by a person more familiar with their implementation.

Below is an excerpt of extra debug logs after this patch showing correct socket selection in a scenario like that described above.  Prior to patch the first socket on the wrong port would have been selected, but after the patch that socket is skipped and the second (correct c=0x7f98e14342a0) socket is selected instead.

```
DBG:core:tcp_conn_get: 0  port 5555
DBG:core:print_ip: tcpconn_find: ip 2.2.2.2
DBG:core:tcp_conn_get: p1 a=0x7f98e1433d90, c=0x7f98e1433cc0, c->id=1848113304, alias port= 5555 src_port=5555 dst_port=5060
DBG:core:print_ip: src_ip=2.2.2.2
DBG:core:print_ip: dst_ip=1.1.1.1
DBG:core:tcp_conn_get: p1 a=0x7f98e1434370, c=0x7f98e14342a0, c->id=1848113305, alias port= 5555 src_port=5555 dst_port=7080
DBG:core:print_ip: src_ip=2.2.2.2
DBG:core:print_ip: dst_ip=1.1.1.1
DBG:core:tcp_conn_get: con found in state 0
DBG:core:tcp_conn_get: tcp connection found (0x7f98e14342a0), acquiring fd
```

### DIFF
--- a/modules/proto_bin/proto_bin.c
+++ b/modules/proto_bin/proto_bin.c
@@ -522,9 +522,9 @@ static int proto_bin_send(struct socket_info* send_sock,
 	if (to){
 		su2ip_addr(&ip, to);
 		port=su_getport(to);
-		n = tcp_conn_get(id, &ip, port, PROTO_BIN, &c, &fd);
+		n = tcp_conn_get(id, &ip, port, PROTO_BIN, &c, &fd, send_sock);
 	}else if (id){
-		n = tcp_conn_get(id, 0, 0, PROTO_NONE, &c, &fd);
+		n = tcp_conn_get(id, 0, 0, PROTO_NONE, &c, &fd, send_sock);
 	}else{
 		LM_CRIT("tcp_send called with null id & to\n");
 		return -1;
--- a/modules/proto_hep/proto_hep.c
+++ b/modules/proto_hep/proto_hep.c
@@ -693,9 +693,9 @@ static int hep_tcp_send (struct socket_info* send_sock,
 	if (to) {
 		su2ip_addr(&ip, to);
 		port=su_getport(to);
-		n = tcp_conn_get(id,&ip, port, PROTO_HEP_TCP, &c, &fd);
+		n = tcp_conn_get(id,&ip, port, PROTO_HEP_TCP, &c, &fd, send_sock);
 	} else if (id) {
-		n = tcp_conn_get(id, 0, 0, PROTO_NONE, &c, &fd);
+		n = tcp_conn_get(id, 0, 0, PROTO_NONE, &c, &fd, send_sock);
 	} else {
 		LM_CRIT("tcp_send called with null id & to\n");
 		return -1;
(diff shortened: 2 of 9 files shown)


attermann:
Also note, I chose not to pass socket_info to tcp_conn_get() in modules/tls_mgm/tls_select.c:get_cur_connection() because I was was unsure about potential unintended side-effects.

```tcp_conn_get(msg->rcv.proto_reserved1, 0, 0, PROTO_NONE, &c, NULL/*fd*/, NULL/*send_sock*/);```

TODO: Evaluate if msg force_send_socket or another socket_info should be passed from here.

attermann:
Anybody acting on pull requests still for this project?  While my use case of different ports talking to the same remote service might not be a common one, it still exposes and corrects something pretty fundamental in TCP socket selection. Hoping that somebody can review this soon and get it merged into 2.4 and later so that we don't have to keep merging it into future releases ourselves.

bogdan-iancu:
@attermann , this topic is a bit delicate, grant me a bit of time to review it.

attermann:
@bogdan-iancu No worries I completely understand. I just wanted to be sure that it wasn't overlooked.

BTW, I added another commit that changes the search slightly. As you'll see in the description of the new commit, I removed the "fallback" search of destination-only because I found that it cna return the wrong connection after the correct connection has closed.

attermann:
Hi @bogdan-iancu.  Just checking-in to see if this might be included in the upcoming 2.4.9 release. Thanks.

bogdan-iancu:
This patch was obsoleted by #2513 , thank you @attermann !

PR #2109, opened 2020-05-11
proto_hep: Fix HEPv3 IPv6 destination contents (send DST not SRC addr)

wdoekes:
Found by the nice people at Devhouse Spindle. If you could cherry-pick this one down to 2.4 too, that'd be great :)


liviuchircu:
Nice catch, all!  Cheers!

Issue #2127, opened 2020-06-01
[BUG] Missed column in database schema for PostgreSQL (maybe something else broken too?)


Shkiperon:
```
version: opensips 3.1.0-beta (x86_64/linux)
flags: STATS: On, SHM_EXTRA_STATS, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, CC_O0, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 93798a002
main.c compiled on 13:00:32 May 28 2020 with gcc 8
```
Cannot start opensips-3.1 with callcenter module and postgresql backend - fail to start. Reason - bad sql-configuration file for deploy database, because there are no column `wrapup_time`, but callcenter need it.

I'm not sure that this is all, but maybe you can find something else that missed in DB schema or in code of module

**To Reproduce**
Example:
1. Configure opensips.cfg to work with callcenter module and PostgreSQL backend
2. Try to start OpenSIPS

**Expected behavior**

A clear OpenSIPS start with ready for action callcenter module.

**Relevant System Logs**
```
Jun 01 11:37:09 os1 /usr/local/sbin/opensips[3590]: INFO:db_postgres:mod_init: initializing...
Jun 01 11:37:09 os1 /usr/local/sbin/opensips[3590]: INFO:call_center:mod_init: Call Center module - initializing
Jun 01 11:37:09 os1 /usr/local/sbin/opensips[3590]: INFO:core:evi_publish_event: Registered event <E_CALLCENTER_AGENT_REPORT(4)>
Jun 01 11:37:09 os1 /usr/local/sbin/opensips[3590]: ERROR:db_postgres:db_postgres_submit_query: 0x7fbdfc10c020 PQsendQuery Error: ERROR:  column "wrapup_time" does not exist
                                                    LINE 1: ... agentid,location,skills,logstate,wrapup_end_time,wrapup_tim...
                                                                                                                 ^
                                                     Query: select agentid,location,skills,logstate,wrapup_end_time,wrapup_time from cc_agents
[...]

PR #2130, opened 2020-06-02
Fix db schema for cc_agents table

Shkiperon:
Fix for issue #2127


bogdan-iancu:
@Shkiperon , many thanks for the fix. I had to change it as the default value must be 0, to allow the per-agent wrapup time to be overwritten by the global wraptime value.

PR #2131, opened 2020-06-03
Fix missing column in subscriber table

Shkiperon:
I found another problem in DB schema configuration. Now it is "subscriber" table.
In current version if you try register SIP-account based on auth_db module you will catch error (if PostgreSQL backend):
```
Jun  3 16:26:07 os1 /usr/local/sbin/opensips[16659]: ERROR:db_postgres:db_postgres_submit_query: 0x7f4b4a61ffe0 PQsendQuery Error: ERROR:  column "uri_user" does not exist#012LINE 1: ...rname='101' AND domain='example.local' AND uri_user='...#012
^#012 Query: select username from subscriber where username='101' AND domain='example.local' AND uri_user='101'
Jun  3 16:26:07 os1 /usr/local/sbin/opensips[16659]: ERROR:core:db_do_query: error while submitting query - [select username from subscriber where username='101' AND domain='example.local' AND uri_user='101']
Jun  3 16:26:07 os1 /usr/local/sbin/opensips[16659]: ERROR:auth_db:check_username: Error while querying database
```

### DIFF
--- a/db/schema/subscriber.xml
+++ b/db/schema/subscriber.xml
@@ -77,6 +77,14 @@
     </column>
 
     <column>
+        <name>uri_user</name>
+        <type>string</type>
+        <size>&user_len;</size>
+        <default/>
+        <description>URI user</description>
+    </column>
+
+    <column>
         <name>rpid</name>
         <type>string</type>
         <size>&domain_len;</size>
--- a/scripts/db_berkeley/opensips/subscriber
+++ b/scripts/db_berkeley/opensips/subscriber
@@ -1,5 +1,5 @@
 METADATA_COLUMNS
-id(int) username(str) domain(str) password(str) email_address(str) ha1(str) ha1b(str) rpid(str)
+id(int) username(str) domain(str) password(str) email_address(str) ha1(str) ha1b(str) uri_user(str) rpid(str)
 METADATA_KEY
 1 2 
 METADATA_READONLY
@@ -7,4 +7,4 @@ METADATA_READONLY
 METADATA_LOGFLAGS
 0
 METADATA_DEFAULTS
-NIL|''|''|''|''|''|''|NULL
+NIL|''|''|''|''|''|''|''|NULL
--- a/scripts/dbtext/opensips/subscriber
+++ b/scripts/dbtext/opensips/subscriber
@@ -1,1 +1,1 @@
-id(int,auto) username(string) domain(string) password(string) email_address(string) ha1(string) ha1b(string) rpid(string,null) 
+id(int,auto) username(string) domain(string) password(string) email_address(string) ha1(string) ha1b(string) uri_user(string) rpid(string,null) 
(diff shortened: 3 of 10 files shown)


bogdan-iancu:
@Shkiperon , the table is *not* supposed to have that column. And I suppose we are talking about 3.1 here, right?

What is the script function you are using, which triggers the above error ?

Shkiperon:
Yes, it is about 3.1.
I try to use subscriber table, as I was do earlier. Here is the script:
```
        if ( !(is_method("REGISTER")  ) ) {
                if (is_from_local()) {
                        if (!www_authorize("", "subscriber")) {
                                www_challenge("", "auth");
                                exit;
                        }
                        if (!db_is_from_authorized("subscriber")) {
                                send_reply(403,"Forbidden auth ID");
                                exit;
                        }
                        consume_credentials();
                } else {
                        if (!is_uri_host_local()) {
                                send_reply(403,"Relay Forbidden");
                                exit;
                        }
                }
        }
```

Shkiperon:
"Earlier" - I mean in 2.4

bogdan-iancu:
Let me check, the problem may be present starting 3.0 when the 'uri' module was removed and its functionality moved into either `auth_db`, either `sipmsgops` (see https://www.opensips.org/Documentation/Migration-2-4-0-to-3-0-0#toc27)

Shkiperon:
In ./modules/auth_db/authdb_mod.c I have found this code:
```
/*
 * Module parameter variables
 */
static str db_url           = {NULL,0};
str user_column             = {USER_COL, USER_COL_LEN};
str domain_column           = {DOMAIN_COL, DOMAIN_COL_LEN};
str pass_column             = {PASS_COL, PASS_COL_LEN};
str pass_column_2           = {PASS_COL_2, PASS_COL_2_LEN};

str uri_user_column         = str_init("username");
str uri_domain_column       = str_init("domain");
str uri_uriuser_column      = str_init("uri_user");

int calc_ha1                = 0;
int use_domain              = 0; /* Use also domain when looking up in table */
```
This is the reason why I think about missing of column in DB schema

bogdan-iancu:
investigating.....

bogdan-iancu:
@Shkiperon , do you use in 2.4 the `uri` table  ?

Shkiperon:
No, I'm use `subscriber` table in 2.4.

bogdan-iancu:
@Shkiperon you did a wrong migration from 2.4 to 3.0.

1) if you do not use the `uri` table in 2.4, as per [migration docs](https://www.opensips.org/Documentation/Migration-2-4-0-to-3-0-0#toc27), the `db_check_from()` should be replaced with `$au==$fU `.

2) starting 3.0, the [`db_is_to|from_authorized()`](https://opensips.org/html/docs/modules/3.0.x/auth_db.html#func_db_is_to_authorized) functions take as parameter the name of an `uri` like table, not of a `subscriber` like tables, hence your error (querying a wrong table)

PR #2150, opened 2020-06-23
dispatcher: added database column type checks on reload

pateldp:
Added some database column checks while reloading dispatcher data from database to OpenSIPS cache. It strictly checks database values and user has to follow the table version strictly otherwise dispatcher module will not load and return error.

### DIFF
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -1025,19 +1025,13 @@ static ds_data_t* ds_load_data(ds_partition_t *partition, int use_state_col)
 		}
 
 		weight = 1;
-
-		/* weight */
-		if (VAL_TYPE(values+3) == DB_INT || VAL_TYPE(values+3) == DB_BIGINT) {
-			weight = VAL_INT(values+3);
-			memset(&weight_st, 0, sizeof weight_st);
-		} else {
-			/* dynamic weight, given as a communication socket string */
-			get_str_from_dbval("WEIGHT", values+3,
-			                   0/*not_null*/, 0/*not_empty*/, weight_st, error2);
-			if (!is_fs_url(&weight_st)) {
-				str2int(&weight_st, (unsigned int *)&weight);
-				memset(&weight_st, 0, sizeof weight_st);
-			}
+                /* weight */
+		/* dynamic weight, given as a communication socket string or a static weight*/
+		get_str_from_dbval("WEIGHT", values+3,
+		    0/*not_null*/, 0/*not_empty*/, weight_st, error2);
+		if (!is_fs_url(&weight_st)) {
+		    str2int(&weight_st, (unsigned int *)&weight);
+		    memset(&weight_st, 0, sizeof weight_st);
 		}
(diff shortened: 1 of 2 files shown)


pateldp:
Hello, can anyone look into this. I know it is not something critical. but it creates confusion if you are using mysql view for dispatcher table. E.g. your setid has char datatype and you have some random number as setid and that is the pointer you stored as string and similar happen with priority.

modules/dispatcher/dispatch.c
```diff
@@ -1030,7 +1027,7 @@ static ds_data_t* ds_load_data(ds_partition_t *partition, int use_state_col)
 		weight = 1;
 
 		/* weight */
-		if (values[3].type == DB_INT) {
+		if (VAL_TYPE(values+3) == DB_INT || VAL_TYPE(values+3) == DB_BIGINT) {
```
bogdan-iancu, line 1030: Why not using `get_int_from_dbval` here also?

pateldp: Hello Bogdan,

I think, the weight column should have integer or char (**freeswitch esl string for dynamic weight calculation**) datatype. so I just provide the flexibility there. But as per our mysql table version the weight column has char datatype. If we want to use only char datatype there, then we can even remove even integer support for the weight column. what you thing about that?

pateldp:
Hello @bogdan-iancu 

I recommended that weight column have char type otherwise we are not going to accept the weight and fail the opensips due wrong datatype of dispatcher table.

modules/dispatcher/dispatch.c
```diff
@@ -1025,19 +1025,13 @@ static ds_data_t* ds_load_data(ds_partition_t *partition, int use_state_col)
...
-			get_str_from_dbval("WEIGHT", values+3,
-			                   0/*not_null*/, 0/*not_empty*/, weight_st, error2);
-			if (!is_fs_url(&weight_st)) {
-				str2int(&weight_st, (unsigned int *)&weight);
-				memset(&weight_st, 0, sizeof weight_st);
-			}
+                /* weight */
+		/* dynamic weight, given as a communication socket string or a static weight*/
+		get_str_from_dbval("WEIGHT", values+3,
+		    0/*not_null*/, 0/*not_empty*/, weight_st, error2);
```
liviuchircu, line 1031: I wholeheartedly agree with your simplification -- there is no need to tolerate `INT` for the `weight` column, if the official DB schema says it's a `CHAR(64)`.

liviuchircu (changes requested):
Hey, @pateldp!

What are the core problems that you are trying to solve with this PR, @pateldp?  If you are trying to make the C code compatible around some `dispatcher` SQL view that produces `BIGINT` instead of `INT`, then I suggest fixing this on the SQL side instead, with a function that casts `BIGINT` data down to `INT`:

```
CREATE FUNCTION `BigToInt`(n BIGINT) RETURNS int(11)
    DETERMINISTIC
RETURN n;
```

Cheers,

PR #2188, opened 2020-07-29
Fix command in documentation for clusterer_list_cap

fgast:
Change command in example to the right one.


liviuchircu:
Sharp eyes, @fgast!  Nice!

PR #2193, opened 2020-08-03
Add TLS support to module/httpd

fgast:
As our policies require encrypted connections only, i enabled tls support within the httpd module,
so opensips-cli can be used from a remote instance over an encrypted channel. 

Due the limitations of the used libmicrohttpd, enabling https will disable http connections, but in our case this is fine & wanted.

A PR for opensips-cli handling with self-signed certificates will follow.

### DIFF
--- a/modules/httpd/doc/httpd_admin.xml
+++ b/modules/httpd/doc/httpd_admin.xml
@@ -1,20 +1,29 @@
 <!-- Module User's Guide -->
 
 <chapter>
-	
+
 	<title>&adminguide;</title>
-	
+
 	<section id="overview" xreflabel="Overview">
 	<title>Overview</title>
 	<para>
 		This module provides an HTTP transport layer for &osips;.
 	</para>
 	<para>
-		Implementation of httpd module's http server is based on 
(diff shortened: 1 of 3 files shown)


razvancrainea:
Hi, @fgast ! Thank you for your contribution, really appreciated!
The only thing that I would point out is that other modules that are using tls (such as [rest_client](https://opensips.org/docs/modules/3.1.x/rest_client.html#func_rest_init_client_tls), [db_mysql](https://opensips.org/docs/modules/3.1.x/db_mysql#param_use_tls)) use the [tls_mgm domains](https://opensips.org/docs/modules/3.1.x/tls_mgm.html#idp2426272) to define TLS configurations. While you (currently) only need some of those fields, some others might be required in the future. Therefore it would be nice if we could have `httpd` use them as well. On the other hand this might be a bit misleading, as TLS domains might be provisioned through DB, and reloaded during runtime - this won't be visible in the httpd module, since it can only be configured during startup. But I think it's worth exploring the `tls_mgm` domains in the future anyway.

PR #2217, opened 2020-08-19
Fix db_insert_update (upsert) in sqlite

egreenmachine:
In the `db_sqlite` module the function `db_insert_update` uses sql syntax that is not supported by sqlite. It looks like it was copied from the `db_mysql` module. The mysql command ends up being a `INSERT - ON DUPLICATE KEY UPDATE`. The `ON DUPLICATE KEY` is not supported by sqlite. https://stackoverflow.com/questions/2717590/sqlite-insert-on-duplicate-key-update-upsert

I took the approach of changing this to be `INSERT OR REPLACE` for sqlite to allow wider compatibility. Based on the syntax Actual [upsert support in sqlite](https://www.sqlite.org/lang_UPSERT.html) was added in version 3.24 which is not available on some distros.

### DIFF
--- a/modules/db_sqlite/dbase.c
+++ b/modules/db_sqlite/dbase.c
@@ -684,7 +684,7 @@ int db_last_inserted_id(const db_con_t* _h)
 
 	CON_SET_CURR_PS(_h, &ps);
 #endif
-	ret = snprintf(sql_buf, SQL_BUF_LEN, "insert into %.*s (",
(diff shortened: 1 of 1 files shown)


liviuchircu (approved):

liviuchircu:
Let's get this one merged, @egreenmachine -- I will take care of the backports.  Cheers!

PR #2222, opened 2020-08-22
Fix tls_method typo in cfg templates

l2dy:
It was `TLSv1` before 61fca80e536e5e2f1d7434a5bff017b235ed1f46.


liviuchircu:
Nice catch, @l2dy! <3

PR #2228, opened 2020-08-27
emergency: fix memory leak on error

l2dy:


wdoekes (changes requested):
Might as well return before the malloc() instead.

And I wonder what that `curl_global_cleanup` is doing there...

l2dy:
@liviuchircu Can you please take a look?

liviuchircu (approved):
Apologies for the long wait, @l2dy -- I've taken a look and everything looks good.

@wdoekes: I agree that the `curl_global_cleanup()` is completely misplaced there and must be removed -- let me do that on a separate commit.  As far as refactoring goes... I suggest we take a step back for now --  if we see heavy usage from the `emergency` module, we can go back and refactor it.

Issue #1710, opened 2019-05-27
[BUG] nathelper next_hop off by one for usrloc path


hafkensite:
**OpenSIPS version you are running**

```
version: opensips 2.4.5 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
main.c compiled on  with gcc 6.3.0
```
Thats with one aditional commit: b7570319ef0e150053a124275bc74cd87ff7bf03

**Describe the bug**
The nathelper module gets the wrong host for registrations that have a path set in the `usrloc` table. It seems to be an off by one, because of changes in [get_domain_db_ucontacts](https://github.com/OpenSIPS/opensips/commit/b7570319ef0e150053a124275bc74cd87ff7bf03#diff-39ffc0c353f96afad98ccfe9489eb091R111). 
Without b7570319ef0e150053a124275bc74cd87ff7bf03 this bug does not trigger (but we do experience segfaults).

**To Reproduce**
Build opensips 2.4.5 with commit b7570319ef0e150053a124275bc74cd87ff7bf03 added to it, then register a device (with a Path header) and have the nathelper do its thing.
I don't have a a clean configuration I can share, but don't think its needed.

**Expected behavior**
`DBG:nathelper:nh_timer: resolving next hop: ':172.20.0.25'` should have been `DBG:nathelper:nh_timer: resolving next hop: '172.20.0.254'`.

**Relevant System Logs**
```
DBG:core:db_new_result: allocate 48 bytes for result set at 0x7fe104ebebf8
DBG:db_mysql:db_mysql_get_columns: 6 columns returned from the query
DBG:core:db_allocate_columns: allocate 168 bytes for result columns at 0x7fe104ebed28
DBG:db_mysql:db_mysql_get_columns: RES_NAMES(0x7fe104ebed58)[0]=[received]
DBG:db_mysql:db_mysql_get_columns: use DB_STRING result type
DBG:db_mysql:db_mysql_get_columns: RES_NAMES(0x7fe104ebed68)[1]=[contact]
[...]

Issue #1652, opened 2019-03-29
[CRASH] OpenSIPS crashes since of child that serves rtpproxy


zenichev:
**OpenSIPS version you are running**

```
version: opensips 2.4.5 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: d025b4f61
main.c compiled on 17:14:18 Mar 19 2019 with gcc 7
```

**Crash Core Dump**

**Describe the traffic that generated the bug**

I have a setup with three SBCs working simultaneously, and traffic is balanced to them using DNS.
The problem is that I migrated from 2.1 branch to 2.4.5 (that is latest LTS) and have started to experience some weird issues with crash of child that serves rtpproxy service.

Why I think it's rtpproxy, is because the PID with which rtpproxy starts working, matches the PID that is set as dead child in the last CRITICAL row. So like this:
```
Mar 29 09:55:09 <some_name> opensips[24108]: INFO:rtpproxy:rtpp_test: rtp proxy </var/run/rtpproxy/rtpproxy.sock> found, support for it enabled
Mar 29 10:02:48 <some_name> opensips[24108]: CRITICAL:core:sig_usr: segfault in process pid: 24108, id: 35
Mar 29 10:02:48 <some_name> opensips[24136]: CRITICAL:core:handle_worker: dead child 35 (EOF received), pid 24108
Mar 29 10:02:48 <some_name> opensips[24064]: INFO:core:handle_sigs: child process 24108 exited by a signal 11
```
And this is a constant thing, so each time opensips crashes, child that crashed, has PID of rtpproxy service process (not sure how to describe this better, but I hope it's descriptive for you).

**To Reproduce**

1. Start OpenSIPS 2.4.5 on 18.04.2 LTS
2. Start rtpproxy with version 2.2.alpha.2af612e (the one taken from master branch of the rtpproxy git project).
3. Start obtaining a traffic, especially traffic that time-outs sessions by absence of sound for a while (no audio and BYE generated with help of rtpp_notify_socket).
4. Get a crash

[...]

PR #2233, opened 2020-08-31
Fix next_hop crash (seen in nathelper nh_timer) due to reading reused mem

wdoekes:
Problem:

- get_domain_db_ucontacts (through get_domain_ucontacts) was handing out
  (next_hop) pointers to memory that was unused. This resulted in a
  crash when this memory was reused before the invalid pointer was
  reused.

Relevant issues and commits:

- #1652 [OpenSIPS crashes since of child that serves rtpproxy]
- e162f5f10 [fix 1652: usrloc: make next_hop point within the shared buffer]
- #1710 [nathelper next_hop off by one for usrloc path]
- 0300eb1d5 [fix 1710 / revert 1652: usrloc: fix next hop compute for ...]

That is: e162f5f10 fixes this exact problem in get_domain_db_ucontacts, in
get_domain_mem_ucontacts and in get_domain_cdb_ucontacts (cdb_pack_ping_data).

But in 0300eb1d5 it is reverted for only get_domain_db_ucontacts.

This fix:

- Rewrites the fix for get_domain_db_ucontacts and
  get_domain_cdb_ucontacts, making it less fragile/bug-prone.
- Adds comments about fragility to get_domain_mem_ucontacts
- Fixed unaligned memcpy that might affect non-intel CPUs:
  `((struct proxy_l *)cp)->name.s = next_hop_host`

Bug reported and fix tested by Jasper Hafkenscheid @hafkensite (VoIPGRID).

### DIFF
--- a/modules/usrloc/dlist.c
+++ b/modules/usrloc/dlist.c
@@ -122,7 +122,7 @@ static int get_domain_db_ucontacts(udomain_t *d, void *buf, int *len,
 	db_res_t *res = NULL;
 	db_row_t *row;
 	db_val_t *val;
-	str uri, host, flag_list;
+	str flag_list;
 	int i, no_rows = 10;
 	time_t now;
 	char *p, *p1;
@@ -131,6 +131,7 @@ static int get_domain_db_ucontacts(udomain_t *d, void *buf, int *len,
 	int needed;
 	int shortage = 0;
 	uint64_t contact_id;
+	void *record_start;
 
 	/* Reserve space for terminating 0000 */
 	if (zero_end)
@@ -254,13 +255,36 @@ static int get_domain_db_ucontacts(udomain_t *d, void *buf, int *len,
 				continue;
 			}
 
+			/* We cannot add items that point to p or p1 here,
+			 * unless they point into the buffer we're keeping.
+			 * So: we write to buf, but keep a pointer to the
+			 * start, so we can revert this record if needed. */
+			record_start = buf;
+
+			/* write received/contact */
+			memcpy(buf, &p_len, sizeof p_len);
+			buf += sizeof p_len;
+			memcpy(buf, p, p_len);
+			p = buf; /* point to to-be-kept copy of p */
+			buf += p_len;
+
+			/* write path */
+			memcpy(buf, &p1_len, sizeof p1_len);
+			buf += sizeof p1_len;
+			memcpy(buf, p1, (unsigned)p1_len);
+			p1 = buf; /* point to to-be-kept copy of p1 */
+			buf += p1_len;
+
 			/* determine and parse the URI of this contact's next hop */
 			if (p1_len > 0) {
 				/* send to first URI in path */
+				str uri, host;
 				host.s   = p1;
 				host.len = p1_len;
 				if (get_path_dst_uri(&host, &uri) < 0) {
 					LM_ERR("failed to get dst_uri for Path\n");
+					/* revert writing this record, continue with next */
+					buf = record_start;
 					continue;
 				}
 				if (parse_uri(uri.s, uri.len, &puri) < 0) {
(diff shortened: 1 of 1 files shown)


wdoekes:
Note that this is an issue in 2.4 too. So a backport would be nice :)

liviuchircu:
Thank you for taking the time to document all of this and prepare a fix, @wdoekes !  First of all, to confirm we're on the same page, please confirm/infirm each of my views of the current situation:

1. the initial fix, e162f5f, was pretty strong, as it fixed `get_domain_mem_ucontacts()` and `get_domain_cdb_ucontacts()` for good.  However, it introduced a small bug in `get_domain_db_ucontacts()`.
2. This bug had to do with the computation of the `path` offset, wisely spotted by @hafkensite in #1710 and addressed in 0300eb1
3. this time, 0300eb1 was an unfortunate quick-fix, as the original bug behind #1652 had resurfaced (pointing to freed PKG memory within the `next_hop.name` struct).  This bug/crash is also the reason why you started investigating this matter in the first place

Quick question:  just to confirm I understood your fix correctly, wouldn't the following "cheap fix" remediate the bug discussed in point (3) and basically equivalate any behavioral fixes that this PR brings?

1.  revent commit 0300eb1 (i.e. "revert the revert")
2.  apply the following _cheap fix_:

```
diff --git a/modules/usrloc/dlist.c b/modules/usrloc/dlist.c
index e1ba0de5b..815fa1467 100644
--- a/modules/usrloc/dlist.c
+++ b/modules/usrloc/dlist.c
@@ -271,7 +271,7 @@ static int get_domain_db_ucontacts(udomain_t *d, void *buf, int *len,
                                        return -1;
                                }
                                /* next_hop_offset skips the 'sip:[...@]' */
-                               next_hop_offset = puri.host.s - uri.s;
+                               next_hop_offset = puri.host.s - p1;
                        } else {
                                if (parse_uri(p, p_len, &puri) < 0) {
                                        LM_ERR("failed to parse contact of next hop: '%.*s'\n",
```

Let me know your thoughts.  By the way, I'm a big fan of your idea of refactoring to reduce fragility and to make future code less error-prone!

wdoekes:
Hi there @liviuchircu,

you're correct on all 5 counts.

So yes. The "cheap fix" in terms of minimal changes would indeed be 2.1 + 2.2.

However, I would argue that my version is more readable *while it is exactly as performant.* The fact that 0300eb1 was incorrectly chosen as a valid fix, substantiates my argument that the code readability needed improvement.

The only drawback I see with my changeset is that I didn't touch the get_domain_mem_ucontacts() code, leaving the harder to read (and thus harder to validate) code/computations there. (But replacing that code with the same fix would be awkward, as the values were already parsed there.)

modules/usrloc/dlist.c
```diff
@@ -254,13 +255,36 @@ static int get_domain_db_ucontacts(udomain_t *d, void *buf, int *len,
...
+			memcpy(buf, &p1_len, sizeof p1_len);
+			buf += sizeof p1_len;
+			memcpy(buf, p1, (unsigned)p1_len);
+			p1 = buf; /* point to to-be-kept copy of p1 */
+			buf += p1_len;
+
 			/* determine and parse the URI of this contact's next hop */
 			if (p1_len > 0) {
 				/* send to first URI in path */
+				str uri, host;
```
liviuchircu, line 281: While I'm a big fan of block-scope variable declarations, what do you think is the best approach given that using them opens the doors to problems such as the one in 3ccea1a32?  Talk to Nick to drop those types of builds altogether or take it upon ourselves to refrain from assuming a C99+ standard for a couple more years?

liviuchircu (approved):

liviuchircu:
I forgot to sum up here:  everything looks great, @wdoekes.  Also, you know what?  Let's merge it as it is -- if the build breaks, we can re-open the "variable scoping" discussion then.  Slowly, we have to move forward to C99+ C standard anyway!

modules/usrloc/dlist.c
wdoekes: This is already legal in old C.

The only problem in 3ccea1a is `for (int i` that's standard C++ and C99+, but not ansi/C89.

See this example:
```c
int func() {
    int x;

    {
#ifndef OLD_C
        for (int i = 0; i < 10; ++i) {
#else
        int i;
        for (i = 0; i < 10; ++i) {
#endif
            float y;
            y += i;
            x = y;
        }
    }

    return x;
}
```
```console
$ gcc -Wall -ansi -c example.c
example.c: In function ‘func’:
example.c:6:9: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
         for (int i = 0; i < 10; ++i) {
         ^~~
example.c:6:9: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code

$ gcc -DOLD_C -Wall -ansi -c example.c
(success)
```

liviuchircu: Brilliant!  Thank you! <3

PR #2237, opened 2020-09-02
usrloc: Update doxygen comment about packing get_all_ucontacts

wdoekes:
Only later I noticed the comments had been out-of-sync for some time. This should fix things.

### DIFF
--- a/modules/usrloc/dlist.c
+++ b/modules/usrloc/dlist.c
@@ -798,24 +798,25 @@ get_domain_mem_ucontacts(udomain_t *d,void *buf, int *len, unsigned int flags,
 
 /*! \brief
  * Return list of all contacts for all currently registered
- * users in all domains. Caller must provide buffer of
- * sufficient length for fitting all those contacts. In the
- * case when buffer was exhausted, the function returns
- * estimated amount of additional space needed, in this
- * case the caller is expected to repeat the call using
- * this value as the hint.
+ * users in all domains. Caller must provide a buffer of
+ * sufficient length to fit all those contacts. If the buffer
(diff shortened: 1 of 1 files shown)


modules/usrloc/dlist.c
```diff
@@ -798,24 +798,25 @@ get_domain_mem_ucontacts(udomain_t *d,void *buf, int *len, unsigned int flags,
...
- * |contact1.len|contact1.s|sock1|flags1|path1|contact_coords1 |
- * +------------+----------+-----+------+-----+----------------+
- * |contact2.len|contact2.s|sock2|flags2|path1|contact_coords2 |
- * +------------+----------+-----+------+-----+----------------+
- * |..........................................|................|
- * +------------+----------+-----+------+-----+----------------+
- * |contactN.len|contactN.s|sockN|flagsN|pathN|contact_coordsN |
- * +------------+----------+-----+------+-----+----------------+
+ * +------------+----------+---------+-------+------------+--------+--------+---------------+
+ * |int         |char[]    |int      |char[] |socket_info*|unsigned|str     |uint64         |
```
liviuchircu, line 810: the `next_hop` seems to be a `struct proxy_l`, not an `str`, right?

liviuchircu (changes requested):

modules/usrloc/dlist.c
wdoekes: Ah, drat. You're absolutely right.

wdoekes: Updated.

liviuchircu (approved):

PR #2241, opened 2020-09-07
[tracer] Enable dynamic tracing even if global trace_on is disabled

### LINKED ISSUE #2240: [BUG] Dynamic sip tracing not enabled on creation
```
version: opensips 3.1.0 (x86_64/linux)
git revision: unknown
main.c compiled on 11:12:50 Sep  4 2020 with gcc 8
```

**Describe the bug**
Hi team,

If I have in the script:
```
#### HEP module
loadmodule "proto_hep.so"
modparam("proto_hep", "hep_id", "[hep_dst] <ip>:<port>; transport=udp; version=3")
#### SIP Tracing module
loadmodule "tracer.so"
modparam("tracer", "trace_on", 0)
modparam("tracer", "trace_id", "[test]uri=hep:hep_dst")
```
and in several places of routing script:
`trace("test", "t");`

Tracing is disabled by default and it's very seldom when I want to enable it to get all the SIP traffic to HEP server or to DB. But of course, I can enable it with 'opensips-cli -x mi trace mode=on'
And I have opensips-cli and it's relatively often, when I want to trace some IP-address or user using console without enabling global tracing. I can create dynamic tracing, but OpenSIPS does not send anything to dynamically created destinations, as global tracing is disabled.
So how to use these short-living dynamic tracings without enabling all the statically configured destinations?

**To Reproduce**
1. Use module `tracer`, but set `trace_on=0`. Several static tracing destinations (to HEP, DB), which are used very seldom.
2. Start `opensips-cli` and use command `trace` to create dynamic tracing ID.
3. Nothing comes from OpenSIPS

**Expected behavior**
Dynamic tracing ID should be enabled by default without globally enabling all possible destinations.

 - Operating System: CentOS Linux release 8.2.2004 (Core)
 - OpenSIPS installation: `git`
 - other relevant information:

vasilevalex:
Fixes #2240

### DIFF
--- a/modules/tracer/doc/tracer_admin.xml
+++ b/modules/tracer/doc/tracer_admin.xml
@@ -32,17 +32,17 @@
 	The tracing tracing can be turned on/off using fifo command.
 	</para>
 	<para>
-	opensips-cli -x mi trace on
-	opensips-cli -x mi trace [some_trace_id] on
+	opensips-cli -x mi trace mode=on
+	opensips-cli -x mi trace id=[some_trace_id] mode=on
 	</para>
 	<para>
-	opensips-cli -x mi trace off
-	opensips-cli -x mi trace [some_trace_id] off
+	opensips-cli -x mi trace mode=off
+	opensips-cli -x mi trace id=[some_trace_id] mode=off
 	</para>
 
 	<para>
 		Starting with &osips; 3.0 you can use the <emphasis>trace_start</emphasis> to
-		create dynamic dynamic tracing destinations based on some custom filters.
+		create dynamic tracing destinations based on some custom filters.
 	</para>
 	</section>
 	<section id="dependencies" xreflabel="Dependencies">
--- a/modules/tracer/tracer.c
+++ b/modules/tracer/tracer.c
@@ -1058,7 +1058,7 @@ static int save_siptrace(struct sip_msg *msg, db_key_t *keys, db_val_t *vals,
 		return -1;
 	}
 
-	if (!(*trace_on_flag)) {
+	if (!(*trace_on_flag)&&(!dyn_trace_list)) {
 		LM_DBG("trace is off!\n");
 		return 0;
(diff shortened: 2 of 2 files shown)


vasilevalex:
Please review, may be it's not the way it intended to work.

vasilevalex:
waiting for review

razvancrainea:
More or less, this is exactly how it is intended to work - if the tracing is disabled, dynamic tracing is also disabled.
As I said in ticket #2240 , I suspect that this commit enables tracing for all destinations, even the non-dynamic ones. Did you get a chance to check this?

vasilevalex:
@razvancrainea I've rechecked the changes. When going through trace_list (static entries) it always checks (it->traceable && !(*it->traceable)) and during init stage this flag is set to the value of global `trace_on`. And I also made tests and I have traces in opensips-cli, but have nothing send over hep. So with this PR only dynamic tracing is enabled. static entries are still disabled.
But without this PR there is no way at all to enable dynamic tracing except global setting. For static entries it is possible to enable specific entry with `opensips-cli -x mi trace id=[some_trace_id] mode=on` but this `id` can belong to static list, dynamic completely omitted.

vasilevalex:
up

razvancrainea:
I've finally got some chance to run a few tests, and confirmed that the patch was not complete. As I previously said, `dyn_trace_list` always has a value, otherwise the module wouldn't load - therefore the check you did was useless, basically it was always false.
The problem appears in the following scenario:

1. start opensips with tracing off
2. enable tracing
3. enable specific tracing id
4. turn of tracing
5. start dynamic tracing (using cli)

At this point, both dynamic and static/script tracing is done, even though the global tracing is off, only specific id is on. This is not correct.
Therefore I will close this PR, I will soon push a fix on master. Will update #2240 when this is completed.
Kindly ask you to open a new PR for fixing the documentation though.

PR #2253, opened 2020-09-18
Fix a rather obscure copy-n-paste bug in the digest parser

sobomax:
Fix a rather obscure copy-n-paste bug in the digest parser causing  "Digestn" to be accepted just as well as "Digest" would, however if the header happens to overflow to the next line right after that, i.e.:

```
Authorization: Digest[LF]
 username="opensipit2020"...
```

The header would fail to be parsed at LF. Use is_ws() macro here (and in two more places), don't reinvent a wheel.

### DIFF
--- a/trim.h
+++ b/trim.h
@@ -25,7 +25,14 @@
 #include "str.h"
 
 /* whitespace */
-#define is_ws(c) ((c) == ' ' || (c) == '\r' || (c) == '\n' || (c) == '\t')
+static inline int
+is_ws(unsigned char ch)
+{
+    const unsigned int mask = (1 << (' ' - 1)) | (1 << ('\r' - 1)) |
+        (1 << ('\n' - 1)) | (1 << ('\t' - 1));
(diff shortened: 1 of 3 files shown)


liviuchircu (approved):

trim.h
```diff
@@ -25,7 +25,14 @@
 #include "str.h"
 
 /* whitespace */
-#define is_ws(c) ((c) == ' ' || (c) == '\r' || (c) == '\n' || (c) == '\t')
+static inline int
+is_ws(unsigned char ch)
+{
+    const unsigned int mask = (1 << (' ' - 1)) | (1 << ('\r' - 1)) |
```
liviuchircu, line 31: Notice you have diverged from the overall coding style of the file on two levels:

* Windows-style function type/name alignment (2 lines), instead of Linux-style (1 line)
* 4 spaces instead of tab for alignment (98%+ of OpenSIPS code uses tabs)

PR #2262, opened 2020-09-28
Few FreeBSD fixes into test system.

sobomax:
o Don't assume bash lives in /bin;

o fix arguments ordering while calling head to work correctly on non-GNU systems.

### DIFF
--- a/test/1.sh
+++ b/test/1.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # load a minimal config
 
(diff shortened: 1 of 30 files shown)


liviuchircu (approved):

liviuchircu:
As a general guideline, please make sure that each commit has a title from now on.  Otherwise, the resulting emails have a strange title:

```
o Don't assume bash lives in /bin;
```

What is "o Don't"?  Would be an eye-brow raise from each reader.

PR #2268, opened 2020-09-29
Re-use alg_t from "digest/digest_parser.h"

sobomax:
Re-use alg_t from "digest/digest_parser.h" and get rid of the duplicate AUTHENTICATE_MD5 and AUTHENTICATE_MD5SESS definitions.

### DIFF
--- a/modules/uac_auth/auth.c
+++ b/modules/uac_auth/auth.c
@@ -285,7 +285,7 @@ void uac_calc_HA1( struct uac_credential *crd,
 	MD5Update(&Md5Ctx, crd->passwd.s, crd->passwd.len);
 	MD5Final(HA1, &Md5Ctx);
 
-	if ( auth->flags& AUTHENTICATE_MD5SESS )
+	if ( auth->algorithm == ALG_MD5SESS )
 	{
 		MD5Init(&Md5Ctx);
 		MD5Update(&Md5Ctx, HA1, HASHLEN);
--- a/parser/parse_authenticate.c
+++ b/parser/parse_authenticate.c
@@ -33,10 +33,6 @@
 #include "parse_authenticate.h"
 
 
-#define AUTHENTICATE_MD5         (1<<0)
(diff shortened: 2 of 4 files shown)


modules/uac_auth/auth.c
```diff
@@ -285,7 +285,7 @@ void uac_calc_HA1( struct uac_credential *crd,
 	MD5Update(&Md5Ctx, crd->passwd.s, crd->passwd.len);
 	MD5Final(HA1, &Md5Ctx);
 
-	if ( auth->flags& AUTHENTICATE_MD5SESS )
+	if ( auth->algorithm == ALG_MD5SESS )
```
liviuchircu, line 288: Please squash this fix-commit into the main commit of this PR.  You can use `git rebase -i` and merge them, then force-push to your branch in order to obtain this effect.  Alternatively, you can `git reset HEAD~`, then `git add modules && git commit --amend`, then force-push to your branch.

liviuchircu (changes requested):
Besides the commit squashing change, everything else looks good.  Of course, ideally, we would use the [50/72 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule) for git commit title & body messages, so that horrible 120-character title doesn't wrap around or get cut by various email tools and web GUIs, but let's leave that for the future.

I fully agree with your cleanup idea.  Those macros were 100% code duplication and it's better to simply re-use the ones from `digest/digest_parser.c` (for parsing of the `*Authorization` responses) in order to parse the `*Authenticate` challenges as well.

modules/uac_auth/auth.c
sobomax: Tried git rebase -i as suggest, not sure if it helped a lot, I see 4 commits now. :(

liviuchircu:
Indeed, it got worse now.  It's okay, I have done this operation for you and merged the resulting commit.  If you are interested on the `git rebase -i` commit squashing technique, check it out here:

Here is how

PR #2270, opened 2020-09-30
Cleanup digest_parser code a bit.

sobomax:
o replace another implementation of is_ws() with a is_ws call;

 o use str.len as the primary and the only way to see if string
   has zero lengh. Depending on how algorithm works, str.s might
   be non-NULL even if str.len == 0.


liviuchircu (approved):
LGTM

PR #2272, opened 2020-10-04
Fix in the residential script. Does not make sense to send a plus cha…

flaviogoncalves:
In the residential file it does not make sense to send a number to drouting starting with + as drouting does not support + as a prefix.


bogdan-iancu:
Thank you @flaviogoncalves

PR #2276, opened 2020-10-06
Clean up UAS auth code a bit before I dive into RFC8760 rework.

sobomax:
Changes:

o Get rid of the most XYZ_LEN constants;

o normalize parameters to not pass char * + len, pass str * instead;

o add const where appropriate;

o GC stale _PRINT_MD5 section, it's going to be replaced with the
  new code really soon;

o avoid calling strlen() when the string lenth is well known.

No functional changes (I hope).

### DIFF
--- a/modules/auth/challenge.c
+++ b/modules/auth/challenge.c
@@ -42,8 +42,6 @@
 #include "index.h"
 #include "api.h"
 
-static str auth_400_err = str_init(MESSAGE_400);
-
 
 /*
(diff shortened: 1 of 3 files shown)


sobomax:
P.S. The change has been tested here: https://travis-ci.com/github/sippy/voiptests/builds/188289197

sobomax:
OOPS, one issue has been noticed. Let me clean and re-open.

PR #2277, opened 2020-10-06
Clean up UAS auth code a bit before I dive into RFC8760 rework.

sobomax:
o Extend STR_NULL and str_init() to set a type explicitly. This allows those macros to be used elsewhere in the function body,
not only in variable declaration(s);

o Get rid of the XYZ_LEN constants;

o normalize parameters to not pass char * + len, pass str * instead;

o add const where appropriate;

o GC stale _PRINT_MD5 section, it's going to be replaced with the
  new code really soon;

o avoid calling strlen() when the string lenth is well known.

No functional changes (I hope). Tested here: https://travis-ci.com/github/sippy/voiptests/builds/188456499

### DIFF
--- a/modules/auth/challenge.c
+++ b/modules/auth/challenge.c
@@ -42,8 +42,6 @@
 #include "index.h"
 #include "api.h"
 
-static str auth_400_err = str_init(MESSAGE_400);
-
 
 /*
  * proxy_challenge function sends this reply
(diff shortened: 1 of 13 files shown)


liviuchircu (approved):
Everything looks great, @sobomax!  The `(str)` cast extension is _especially_ powerful, I remember doing quite a few workarounds because `str_init()` couldn't be assigned within a function, hehe.

However, before merging, could you please squash the last two commits into a single one?  Position yourself in your branch, on the last commit (4th of this PR), and do:

    git reset HEAD~
    git add .
    git commit --amend
    git push --force

sobomax:
Thanks Liviu for a feedback! :) Changes have been made as requested, please check.

PR #2279, opened 2020-10-07
Improve auth module examples

sobomax:
o Use $avp(xyz) consistently;

o fix www_challenge / proxy_challenge second argument to match
  reality.


modules/auth/doc/auth_admin.xml
```diff
@@ -205,7 +205,7 @@ modparam("auth", "rpid_avp", "$avp(13)")
 		<example>
 		<title><varname>username_spec</varname> parameter usage</title>
 		<programlisting format="linespecific">
-modparam("auth", "username_spec", "$var(username)")
+modparam("auth", "username_spec", "$avp(username)")
```
liviuchircu, line 208: As a script development best-practice, I suggest normalizing them to `$var` instead, as their scope is much more limited (to the processing of the current message, instead of transaction), as well as the fact that they cannot hold multiple values (which also dispels some AVP-related confusions such as: _"How will the code work if I provide 2 usernames / passwords into each AVP?  Does each pair get checked?"_).

On a performance-related note, vars are stored in PKG memory instead of SHM, so they do not involve any synchronization between processes.  Moreover, they do not impose any extra cleanup overhead at the end of the current message processing, as AVPs do.

liviuchircu (changes requested):

modules/auth/doc/auth_admin.xml
sobomax: Done. Please check. In general, I'd suggest renaming "avp" into something that better captures its purpose. For example "sharedvar" ("globalvar?").

liviuchircu (approved):

modules/auth/doc/auth_admin.xml
liviuchircu: Good point.  There are plans in this direction, but they didn't make it to `3.1`.  For example, I remember I was suggesting something along the lines of: `$var`, `$t_var` and `$dlg_var`.

PR #2286, opened 2020-10-14
Add ability to build various static supporting libraries to contain code that is shared between several modules.

sobomax:
As discussed on opensips-developers mailing list, here is the part of the RFC8760 work to add support for building static libraries for code to be shared between multiple modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.o
 *.d
 *.so
+*.a
 *.sw?
 tags
 cscope*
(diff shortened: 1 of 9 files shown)


liviuchircu (approved):

liviuchircu:
Looks good, @sobomax.  Let's note, one last time, that this PR impacts commands such as `make modules module="dialog usrloc"`, which now have the side-effect of additionally building all miscellaneous libraries.  But we can optimize these Makefile dependencies on a later commit.

Cheers!

PR #2287, opened 2020-10-14
Fix Memory leak in Freeswitch ESL module.

zooptwopointone:
Fix memory leak in freeswitch ESL module.

### DIFF
--- a/modules/freeswitch/esl/src/esl.c
+++ b/modules/freeswitch/esl/src/esl.c
@@ -1406,6 +1406,7 @@ ESL_DECLARE(esl_status_t) esl_recv_event(esl_handle_t *handle, int check_q, esl_
 		if (!esl_safe_strcasecmp(hval, "text/disconnect-notice") && revent->body) {
 			const char *dval = esl_event_get_header(revent, "content-disposition");
 			if (esl_strlen_zero(dval) || strcasecmp(dval, "linger")) {
+				free(revent->body);
 				goto fail;
 			}
 		}


modules/freeswitch/esl/src/esl.c
```diff
@@ -1406,6 +1406,8 @@ ESL_DECLARE(esl_status_t) esl_recv_event(esl_handle_t *handle, int check_q, esl_
 		if (!esl_safe_strcasecmp(hval, "text/disconnect-notice") && revent->body) {
 			const char *dval = esl_event_get_header(revent, "content-disposition");
 			if (esl_strlen_zero(dval) || strcasecmp(dval, "linger")) {
+				char *body = strdup(revent->body);
+				free(body);
```
liviuchircu, line 1410: This is an excellent catch (a real memleak!), however the fix is poorly executed:

1) you are allocating a new block, `body`, then freeing it immediately afterwards?!  A no-operation, basically.
2) if `revent->body` is `NULL`, your patch will now cause a segfault
3) the leak is still there

To fix all above issues, just change to:

    free(revent->body);

liviuchircu (changes requested):

modules/freeswitch/esl/src/esl.c
zooptwopointone: Thanks.  Yea not a very good C programmer myself.    There is one other bug I haven't figured out how to fix with this code.   It is how connects/reconnects are done.  It appears memory is leaking there.  There is comments in the code about basically a FIXME as there was not a great way to handle it.  If someone can provide me some assistance I am happy to work on it/test it.  But I don't understand it well enough to work it out myself.  Anyway this request I have done as you reqeuested.

liviuchircu (changes requested):
Please merge both commits into a single one:

    git reset HEAD~
    git add .
    git commit --amend
    git push --force

liviuchircu (approved):

PR #2289, opened 2020-10-16
Expand Travis setup, add 10 more build variants. Fix many issues.

sobomax:
Switch to bionic (aka Ubuntu 18) as main test vehicle and add bunch of extra jobs:

    o Build with GCC 7 @ Ubuntu 18;
    o Build with GCC 9 @ Ubuntu 18;
    o Build with GCC 10 @ Ubuntu 18;
    o Build with Clang 7 @ Ubuntu 18;
    o Build with Clang 9 @ Ubuntu 18;
    o Build with Clang 10 @ Ubuntu 18;
    o Build with GCC 7 @ Ubuntu 18 (ARM);
    o Build with Clang 7 @ Ubuntu 18 (ARM);
    o Build Core with GCC 7 @ Ubuntu 18 (i386);
    o Build Core with Clang 7 @ Ubuntu 18 (i386).


atomic.h
```diff
@@ -1,203 +0,0 @@
-/*
- * Copyright (C) 2006 kernel.org
- *
- * This file is part of opensips, a free SIP server.
```
liviuchircu, line 4: Interesting stuff, @sobomax!  This header was originally added in 2006, since there was no libc support for these operations.  It is refreshing to find out that they've been available since 2011 and we can get away with maintaining less code :)

liviuchircu (approved):

PR #2300, opened 2020-11-06
Split out gigantic main.c into 3 more TUs: globals.c, shutdown.c and signals.c

sobomax:
This change is a precursor of making it possible to link most of the OpenSIPS code into various unit tests, future direction is something like this:

https://github.com/sippy/opensips/commit/41a154bff641e75d5c556da13fe1ff7e5705483a

https://travis-ci.com/github/sippy/opensips/jobs/428452011#L864

Thanks!


bogdan-iancu:
Hey @sobomax , could you share some details about the needs of this split? I understand it is for making unit tests possible, but can you describe what exactly are the issues with the monolithic approach (the big main.c) and how exactly the split solves the problems?
Also you mentioned a "first step" and "future directions", so could you share the roadmap - this will help (at least me) to evaluate this PR.
Many thanks for your work!

sobomax:
Hey @bogdan-iancu thanks for asking. One problem @liviuchircu is having with integrating his unit tests into the build framework. On one hand he wants to test the exact code that goes into the final binary, on the other hand he then has to have it as an option that is incompatible with anything else (so that we can bypass the configuration file and normal module loading process.

This change would allow us to build libopensipcore.a that includes 99.9% of opensips code and then discretionally link it with either normal "main.o" or lighweight "main.o" that only performs specific unit tests or benchmarks.

See my prototype here for example:

https://github.com/sippy/opensips/commit/41a154bff641e75d5c556da13fe1ff7e5705483a

Building and running unit tests in MIPS64 emulation:

https://travis-ci.com/github/sippy/opensips/jobs/428452011#L864

Hope it explains.

liviuchircu (approved):
Summing up this PR, it basically takes a bunch of functions and global variables from "main.c" and re-distributes them across 5 new files:

* globals.c
* shutdown.[ch]
* signals.[ch]

Apart from the comment I made regarding the need to import "str.h" from "globals.h", I think that long-term, it's better to move to a stand-alone binary for unit-testing (e.g. `opensips_test`), rather than continuing to use the optional `-T` command-line option which was added in the initial version of the unit testing framework.

PR #2301, opened 2020-11-06
Fix MIPS64, add MIPS64 cross-build for travis, merge updates into fastlocks.h

sobomax:
o Merge in relevant improvements and fixes from OpenSER commits by
  andrei@iptel.orgi (2006-2007):

  - mips inline asm gcc 3.x warnings fixed
  - mips2 NOSMP mode (skip sync)
  - minor x86 & mips optimizations
  - fastlock: minor fixes
  - x86/x86_64 lock optimizations:  spinning on a lock should be friendlier now
    for the other cpus caches (at the extra cost of a cmp mem + jump) ; tried to
    arrange a little better the instructions to allow for some parallel
    execution.
  - x86 unlocks with xchg by default (since some x86s reorder stores, so a
    simple mov is unsafe)
  - ppc fixes (s/stw/stwx/, s/lwz/lwzx)
  - missing early clobbers added for x86, sparc*, armv6, ppc*, alpha
  - small arm/sparc optimizations.
  - moved sched_yield() wrapper into sched_yield.h at Miklos's request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,10 +90,10 @@ jobs:
         - GCOV_CMD="llvm-cov-10 gcov"
       before_install:
         - sudo apt-get -y install clang-10 llvm-10-dev libc++-10-dev libc++abi-10-dev
-    - name: "Build with GCC 7 @ Ubuntu 18 (ARM)"
+    - name: "Build with GCC 7 @ Ubuntu 18 (ARM64)"
       compiler: gcc
       arch: arm64
-    - name: "Build with Clang 7 @ Ubuntu 18 (ARM)"
+    - name: "Build with Clang 7 @ Ubuntu 18 (ARM64)"
       compiler: clang
       arch: arm64
     - name: "Build Core with GCC 7 @ Ubuntu 18 (i386)"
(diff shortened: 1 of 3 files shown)


liviuchircu (changes requested):
So, basically, this PR is grabbing the following 4 commits:

* https://github.com/kamailio/kamailio/commit/613e86abfd08
* https://github.com/kamailio/kamailio/commit/a2fd48d9643a
* https://github.com/kamailio/kamailio/commit/ec495407830b
* https://github.com/kamailio/kamailio/commit/a83e261254cb

... correct?  Maybe a 2nd opinion is needed here, from someone who lived those times and knows how the codebase evolved, as I cannot tell whether those are OpenSER commits or SER commits.  OTOH, it's all GPLv2 code, so it can be imported as long as the proper credits are given.  So my suggestion would be to import this work as it was originally done:

* 4 individual commits instead of 1 merge-commit
* the author of these commits must be `"Andrei Pelinescu-Onciul <andrei@iptel.org>"`, instead of `"Maksym Sobolyev <sobomax@sippysoft.com>"`

sobomax:
Thanks Liviu for a review. It's more like a 6 commits some of them trimmed to only keep relevant parts. From my previous PRs I thought you wanted everything flat, so I compressed it down. However here is original unmodified history of my work here:

https://github.com/sippy/opensips/commits/travis4mips

Could you please take a peek and see if that how you want it to look? Thanks!

liviuchircu:
> From my previous PRs I thought you wanted everything flat, so I compressed it down
The reason why I asked you to flatten some commits was because one of them was correcting a bug in a previous one.  Given that we were in a PR context, we still had the option to produce a coherent history with a couple of `git` commands.  It's easier to follow 1 commit which produces a correct feature than 1 commit which produces a bugged feature + 1 bugfix commit, in the same PR.

The situation in this PR is different:  each commit has its own purpose, be it a MIPS bugfix or an optimization.  Merging them would cause confusion, as multi-purpose commits are harder to review.

From a code importing standpoint, the branch you linked looks very good, so please force-push that version.  Unfortunately, MIPS is out of my league for now, so I cannot offer useful advice on that code.  However, since I understand it's currently broken, I can only see this PR doing good things! 
 (Travis also confirms this)

liviuchircu (approved):

sobomax:
Longer version of the merge history is posted. Thanks!

PR #2309, opened 2020-11-12
Add TCP and TCPv6 communication modes for RTPProxy control sockets. 

sobomax:
Those modes are supported on RTPProxy end since 2.0 and are getting popular recently due to Kubrenetes and Docker's networking shenanigans and peculiarities.

### DIFF
--- a/modules/rtpproxy/rtpproxy.c
+++ b/modules/rtpproxy/rtpproxy.c
@@ -174,6 +174,7 @@
 #include "../../parser/parse_body.h"
 #include "../../msg_callbacks.h"
 #include "../../evi/evi_modules.h"
+#include "../../lib/dassert.h"
 
 #include "../dialog/dlg_load.h"
 #include "../tm/tm_load.h"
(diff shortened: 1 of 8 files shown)


modules/rtpproxy/rtpproxy.c
```diff
@@ -2120,9 +2148,16 @@ send_rtpp_command(struct rtpp_node *node, struct iovec *v, int vcnt)
...
 			}
 		}
-		v[0].iov_base = gencookie();
-		v[0].iov_len = strlen(v[0].iov_base);
-		for (i = 0; i < rtpproxy_retr; i++) {
+		if (!CM_STREAM(node)) {
+			v[0].iov_base = gencookie();
+			v[0].iov_len = strlen(v[0].iov_base);
+		} else {
+			memmove(v, v + 1, (vcnt - 1) * sizeof(*v));
```
razvancrainea, line 2155: I am pretty sure that modifying `v` is not correct, as if the current node fails, the function above (`force_rtp_proxy_body`) might failover to a new node, that runs over UDP, so it will need to store the cookie in `v[0]`.
It would probably be easier if we always add the last `\n`, and if the node is stream based, use `v+1`, otherwise `v`.

razvancrainea (commented):
Looks good, there's only one thing I spotted regarding nodes failover.

modules/rtpproxy/rtpproxy.c
sobomax: Good catch, Razvan, thanks for review. I'll work on a fix. I did not like that memmove(), anyhow. 👍 :)

PR #2318, opened 2020-11-22
shm_free(NULL) has no effect 

lemenkov:
shm_free(NULL) has no effect so we can simplify `if(foo) shm_free(foo)` to `shm_free(foo)`.

### DIFF
--- a/modules/acc/acc_logic.c
+++ b/modules/acc/acc_logic.c
@@ -177,8 +177,7 @@ void free_extra_array(extra_value_t* array, int array_len)
 	int i;
 
 	for (i=0; i < array_len; i++) {
-		if (array[i].value.s)
-			shm_free(array[i].value.s);
+		shm_free(array[i].value.s);
 	}
 	shm_free(array);
 }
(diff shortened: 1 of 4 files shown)


liviuchircu:
Hi, Peter!

Some useful tips while you're working on this topic, as I've ran into this problem already and have put some thought into it.

* in places where a pointer is _often_ going to be `NULL`, the code will be faster, on average, if you **include** an `if` before the `shm_free()`, because you avoid two function calls: `shm_free()`, followed by `fm_free()`, for example.
* in places where a pointer is _rarely_ going to be `NULL`, the code will be slower, on average, if you **remove** the `if` before the `shm_free()` (the object of this PR)

Ultimately, I didn't proceed to do these micro-optimizations right away, as I think we could use a better technique, based on _compiler branch prediction_ hints (see the [__builtin_expect()](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) gcc macro).  By integrating `__builtin_expect()` into some kind of `LIKELY()` and `UNLIKELY()` macros, we can properly perform this Pull Request as follows:

* if the pointer is likely to be `NULL`, then we should use `if (UNLIKELY(ptr)) shm_free(ptr)`
* if the pointer is likely to not be `NULL`, then you should remove the `if`, exactly as you propose

lemenkov:
@liviuchircu thanks for the tip! I was thinking about LIKELY/UNLIKELY macros already and I will implement it.

lemenkov:
Hold on for a second. It's the end of the year so I have plenty of things to finish before this almost cosmetic one :)

sobomax:
@lemenkov you can use "Draft" instead of [DO NOT MERGE]. But I totally agree that shm_free(NULL)==nop() is a bad idea from code quality prospective.

PR #2319, opened 2020-11-23
tm/async: Handle ASYNC_DONE_CLOSE_FD when switching to sync

lemenkov:
One more place where we have to close file descriptor according to the
specification while doing async packet processing.


lemenkov:
Waiting for the comments (just to remove "stale" tag).

bogdan-iancu:
Good catch, thank you !!

PR #2320, opened 2020-11-23
tm: Use shm_malloc(_bulk) consistently with free.

lemenkov:
Let's be consistent here. If we use shm_malloc(_bulk) then let's call
shm_free(_bulk) consistently. Even it's the same fuctions which differs
only in locking.

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -268,8 +268,8 @@ inline static int update_totag_set(struct cell *t, struct sip_msg *ok)
 
 	if (!s || !n) {
 		LM_ERR("no more share memory \n");
-		if (n) shm_free(n);
-		if (s) shm_free(s);
+		if (n) shm_free_bulk(n);
+		if (s) shm_free_bulk(s);
 		return 0;
 	}
 	memset(n, 0, sizeof(struct totag_elem));


modules/tm/t_reply.c
```diff
@@ -268,8 +268,8 @@ inline static int update_totag_set(struct cell *t, struct sip_msg *ok)
 
 	if (!s || !n) {
 		LM_ERR("no more share memory \n");
-		if (n) shm_free(n);
-		if (s) shm_free(s);
+		if (n) shm_free_bulk(n);
+		if (s) shm_free_bulk(s);
```
liviuchircu, line 272: Hi, Peter!

This is not OK.  The purpose of `shm_free_bulk()` is to be called within a `shm_lock() / shm_unlock()` block.  It is incorrect to call this function in any other context, as the code will break if you use either of the `F_MALLOC` or `Q_MALLOC` allocators.

Regarding that specific `if` block: it is correct as it is -- no need for any change there.

liviuchircu (changes requested):

modules/tm/t_reply.c
lemenkov: Ah ok, I can see your point. Yes this is a mistake.

Issue #2329, opened 2020-12-03
[BUG] OpenSIPS 3.1 doesn't send actual log_level from xlog to heplify-server


Shkiperon:
**OpenSIPS version you are running**

```
version: opensips 3.1.0 (x86_64/linux)
flags: STATS: On, SHM_EXTRA_STATS, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: e85b96028
main.c compiled on 22:54:43 Nov  9 2020 with gcc 8
```

**Describe the bug**

When using the combo of modules proto_hep, tracer and proto_tcp (trace_on parameter) for sending HEPv3 packets to heplify-server via UDP (with TCP transport heplify-server have issue - https://github.com/sipcapture/heplify-server/issues/447) the "LOG" packets generated when call xlog function with command like `xlog("L_NOTICE","new branch at $ru\n");` it generates "Event" with "ERROR" severity instead of "NOTICE"

**To Reproduce**

1. Configure OpenSIPS 3.1 as HEP-client for heplify-server;
2. Configure script with call xlog fuction with any custom log level instead of "L_ERROR";
3. Start tcpdump and try to make call;
4. Check dump with plugin for wireshark - https://github.com/sipcapture/hep-wireshark/blob/master/lua/hep.lua
5. See that "Event" in "LOG" packet is "ERROR"

**Expected behavior**

Expected that Event type in HEP-packet should be equal the log level the define in call of xlog function in OpenSIPS script.

**Relevant System Logs**

![изображение](https://user-images.githubusercontent.com/38580109/101051650-8f9a5980-35a7-11eb-9ef6-b2c829b3a05d.png)

**OS/environment information**
 - Operating System: Debian 10 
 - OpenSIPS installation: build from git sources 
 - other relevant information:

**Additional context**

Shkiperon:
[...]

PR #2331, opened 2020-12-06
Add updating xlog_print_level in set_local_xlog_level

Shkiperon:
Looks like this update of variable was missed when xlog.c was updated.
Linked issue: https://github.com/OpenSIPS/opensips/issues/2329


liviuchircu (changes requested):
Hey, @Shkiperon!  First of all, let's clear up what problem you are trying to solve.  In `3.1`, a new global was added, [`xlog_level`](https://www.opensips.org/Documentation/Script-CoreParameters-3-2#xlog_level), which decouples engine logging from scripting logging.

The [`xlog_print_level`](https://www.opensips.org/Documentation/Script-CoreParameters-3-2#xlog_print_level) global controls the default logging level of [`xlog()`](https://www.opensips.org/Documentation/Script-CoreFunctions-3-2#toc55) statements which do not include the "log_level" parameter.

I do not see why setting the `$xlog_level` variable for a SIP worker would also have to force a different _default xlog() logging level_ for that worker -- the two settings do not seem to have much in common.

What do you think?

liviuchircu:
Update: just noticed issue #2329, where you describe an actual bug, related to `xlog()` and shipping the log to Homer.  Let me come back to you once I understand it.

Shkiperon:
As I already wrote in my issue - I found a problem that is expressed in the fact that in the HEP-package from OpenSIPS, events with the LOG type go towards the heplify-server with the default event level (ERROR) even if I change it in the script. My knowledge of the C language was only enough to understand that the problem is somewhere in xlog.c. When testing on OpenSIPS 2.4.8, this problem does not occur.

liviuchircu:
Could you clean up your changes, then apply the following small patch after placing yourself in the `opensips/` source code directory and see if it works?  As I see it, the line changed by the patch is all we need to fix here.

```
git apply <(base64 -d <<EOF | gzip -dc
H4sIAM3U0F8AA02NwW6DMBBEz/grVjlBiGko0ChFlVK1OVTKIYo49GZtbENWJQRhK43Un68hiFZ7
mPXMPo+isgTOK7KAD7f6UkUSjuPCqFH6BkmcpEmWaRlFyTpLyjTFDOLl8ilNGed84lgYhn/sZgM8
Xj0uVhDexRnGoiUJ1NTUaLheSAEqJXpCKLTo2w6lFmdtDFYaRl0Ml3NoscNzwMAbvzG2A0Ot6PWl
fwlqyPoz582CnIEb7plvsvIE/tDRdtRYUeurrgP4YeGUzod4CsDzJBoNO/G62x6K597w+oLhIjKu
7n1/z0Sx/SxyOHYav/J/4Nvhw3G/2TRJn14BAAA=
EOF
)
```

Shkiperon:
Yes, I'll try to do it tomorrow. I will give feedback on the results.

Shkiperon:
@liviuchircu your patch has fixed my problem. Thank you

liviuchircu:
Just merged the fix and closed both issues. Thank you for your help!

PR #2339, opened 2020-12-09
Re-introduce using stdatomics for the purposes of core stats. 

sobomax:
This change puts back stdatomic to be used by default when compiler supports it. In case support has not been detected, the compilation fails back to the "poor-man's atomics in atomic.h".

### DIFF
--- a/mem/module_info.c
+++ b/mem/module_info.c
@@ -25,7 +25,11 @@
 #ifdef SHM_EXTRA_STATS
 
 #include <dlfcn.h>
+#ifdef HAVE_STDATOMIC
 #include <stdatomic.h>
+#else
+#include "../atomic.h"
+#endif
 #include <string.h>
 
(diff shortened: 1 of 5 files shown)


mem/module_info.c
```diff
@@ -25,7 +25,11 @@
 #ifdef SHM_EXTRA_STATS
 
 #include <dlfcn.h>
+#ifdef HAVE_STDATOMIC
 #include <stdatomic.h>
+#else
+#include "../atomic.h"
+#endif
```
liviuchircu: On a first thought, maybe we could put this cookie-cutter "#ifdef" code into some common header and just include that one.

However, the more I think about it, the more I like your way better, because it **screams** that it needs simplification to a single line in some distant future, when we will just drop support for gcc 4.8 and below.

liviuchircu (approved):

PR #2343, opened 2020-12-15
WIP: Add whitesource github workflow

dannydiazvonage:
Add whitesource github workflow


dannydiazvonage:
Should only be for h2o repos

PR #2358, opened 2020-12-31
dialog: move "No matched dialogs" msg from LM_ERR to LM_DBG

ccppprogrammer:
Get rid of spam messages "No matched dialogs" in LM_ERR. We only need these messages in LM_DBG.


bogdan-iancu:
Thanks @ccppprogrammer , it makes all the sense.

PR #2363, opened 2021-01-06
Add optional ds state reload

lmendes86:
We manage dispatcher files from an external application (SaltStack), including states, but these don't change in loaded dispatcher sets when running the ds_reload command, only after restarting the daemon.
We noticed this behavior is by design; because of that, we added a new flag to avoid the overwrite of the new states gathered from the DB with the old values loaded in memory. If the flag is not set, the behavior stays the same.

### DIFF
--- a/modules/dispatcher/README
+++ b/modules/dispatcher/README
@@ -44,6 +44,7 @@ dispatcher Module
               1.3.29. socket_col (string)
               1.3.30. fetch_freeswitch_stats (integer)
               1.3.31. max_freeswitch_weight (integer)
+              1.3.32. reload_state (int)
 
         1.4. Exported Functions
 
@@ -134,6 +135,7 @@ dispatcher Module
    1.35. ds_count usage
    1.36. ds_is_in_list usage
    1.37. OpenSIPS config script - sample dispatcher usage
+   1.38. Set the reload_state parameter
 
 Chapter 1. Admin Guide
 
@@ -700,6 +702,19 @@ modparam("dispatcher", "fetch_freeswitch_stats", 1)
    Example 1.33. Set the max_freeswitch_weight parameter
 ...
 modparam("dispatcher", "max_freeswitch_weight", 1000)
+...
+
+1.3.32. reload_state (int)
+
+   Specifies whether the state column should be reloaded on ds_reload
+   MI command execution.
+
+   Default value is “0” (disable).
+
+   Example 1.38. Set the reload_state parameter
+...
+# enable reload state from DB when ds_reload is executed
+modparam("dispatcher", "reload_state", 1)
 ...
 
 1.4. Exported Functions
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -1001,9 +1001,12 @@ int ds_reload_db(ds_partition_t *partition)
 
 	/* destroy old data */
 	if (old_data) {
-		/* copy the state of the destinations from the old set
-		 * (for the matching ids) */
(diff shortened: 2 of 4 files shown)


bogdan-iancu:
Thank you @lmendes86. I'm trying to see the reason behind this feature. I would say that during a reload it is important to preserve the in-memory status (for the destination) as this reflects the reality. May I ask why would you want to reload from DB and override the in-memory status ?
BTW, how is this new option interfering with the [persistent_state](https://opensips.org/html/docs/modules/2.4.x/dispatcher.html#param_persistent_state) option?

lmendes86:
Hello Bogdan, thanks for the answer.
I will try to give more context; we are currently using the dispatcher as a white list in some instances using `ds_is_in_list` function. SaltStack manages this configuration by creating/modifying a dispatcher DB file and executing the `ds_relaod` MI command.
With this being said, sometimes we need to have some destinations in a set to which we don't want to send any traffic, but that we still want to receive traffic from, our first approach was to use `weight` for that by setting `weight=0` to the destinations we didn't want to send traffic to, unfortunately, this only works while you have other active/working/reachable destinations on the same set with a higher weight (set on the same priority of course); after this, we decided to use the `state` attribute which works as expected. 
The only problem with this is that currently the state is only pick up from the DB on the daemon start but not on `ds_reload` execution, that is why we introduce the flag to allow that. In our case, we don't use the state for anything else than this, so we don't really care what was before the reload action.
About the persist_state option, when it's deactivated then the state will not be queried from the DB at all (not even on a restart) and set on active by default, then this feature will overwrite the current states from memory to these new values (all active).
I though on entangle them, like extending the `persistent_state` adding a new possible value to it to allow the reload on this new case, but for me, they were clearly different features, I don't know what you think.

bogdan-iancu:
OK, so you have some destination in DS. Some may be "disabled" (as state, via DB or MI) to prevent sending traffic to (but still accepting from). So far so good.
Now, the question is : why do you want the DB state to be "picked up from the DB during runtime" ? The intended way to change the state of the destination (at runtime) is by using the [ds_set_state](https://opensips.org/html/docs/modules/3.1.x/dispatcher.html#mi_ds_set_state) MI function.

PS: the `persistent_state` has nothing to do with loading the state from DB upon startup. What is does is to write back the in-memory state into DB, so during an OpenSIPS restart the in-memory state will survive.

lmendes86:
Hello Bogdan, thanks for the answer again.
Yes, the intended way to change the states at runtime is to use the `ds_set_state` MI function indeed, but for our use case is a lot easier to overwrite the DB, which is a text file managed by SaltStack, and just trigger `ds_reload`. Of course, is up to you to decide if this feature could be useful for others with a similar setup.
About `persistent_state`, yes, totally agree with you, but I forgot to clarify that in our case, because we are using a text file as the dispatcher DB, OpenSIPS can't write back the in-memory states so then, of course, these states are lost.

bogdan-iancu:
@lmendes86 , we are walking a thin line, between trying to have a lot of flexibility when comes to features, on one side, and trying not to make the usage too confusing, hard to understand, on the other side.
Now, you say you fire the `ds_reload` via SaltStack ? why not triggering the `ds_set_state` instead. In this case, you will update both status, in the text file (as so far) and in-memory, so nothing more is needed.

lmendes86:
Hello @bogdan-iancu, yes indeed we can use the `ds_set_state` and I think that is what we are gonna do in the end to avoid needing this patch. I will close the pull request. Thanks for the discussion!

PR #2371, opened 2021-01-15
b2b_logic: Fix parsing of uri in header

nikbyte:
I found an issue in b2b when process REFER.

I use next rule in xml:
```
<destination>
    <value type="header">Refer-To</value>
</destination>
```

When b2b_logic processes Refer-To header in REFER and the header has content like `"display name" <sip:name@domain.com>`, the b2b_logic module can't parse url there. This patch fixes this issue.

### DIFF
--- a/modules/b2b_logic_xml/logic.c
+++ b/modules/b2b_logic_xml/logic.c
@@ -590,6 +590,7 @@ static b2bl_entity_id_t* b2bl_new_client(str* to_uri, str* from_uri,
 	ci.client_headers= hdrs;
 	ci.body          = (tuple->sdp.s?&tuple->sdp:NULL);
 	ci.from_tag      = NULL;
+	ci.dst_uri       = msg->dst_uri;
 	ci.send_sock     = msg?(msg->force_send_socket?msg->force_send_socket:msg->rcv.bind_address):NULL;
 	if (ci.send_sock) get_local_contact(ci.send_sock, NULL, &ci.local_contact);
 	else ci.local_contact = server_address;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thanks @nikbyte , indeed, this is an issue - so far it was not a big problem, as the devices are usually not pushing a display name into the REFER header, nevertheless, a totally valid options.
I will check the patch and update here.

nikbyte:
However softphones do this. For example, zoiper.

bogdan-iancu:
@nikbyte , could you try my alternative patch, it gets rid of some memory leaks ;). NOTE that the hdr here may be anything , like the "Foo" hdr. And OpenSIPS will have no idea how to free the parsed-body of such unknown hdr - so, adding the "to_body" to the "parsed" holder will result in a leak, as (again) OpenSIPS will not attempt to free it.

bogdan-iancu:
please give a try to my alternative fix and let me know if it does the job ;). Thanks !!

nikbyte:
Bogdan, thank you. Your patch is perfect and works well.

bogdan-iancu:
thank you @nikbyte for the testing , still [this line](https://github.com/OpenSIPS/opensips/pull/2371/files#diff-a109e5901c39ef82abfdc8353930f5a3c7504f3658e8b4af42e7021d6f75ec05R593) did not make it into my patch. 
I see it is very similar to #2374 - what is the plan here ?

bogdan-iancu:
Ok, the pending line from the PR is a NO-GO, thanks @nikbyte for the confirmation.

PR #2389, opened 2021-01-27
[Draft] Libcouchbase3 for merge

lemenkov:
[Libcouchbase](https://github.com/couchbase/libcouchbase)  3.x introduced a different API incompatible with one from 2.x version. I've tried to keep both but I've ended up with unreadable mess of ifdefs. Failed to keep both API inplace I have to say that this is just a draft intended for anyone who is trying to use it with recent libcouchbase library. 

I can maintain it out-of-tree until libcouchbase 2.x will be dropped (replaced with 3.x) from main distributions.

### DIFF
--- a/modules/cachedb_couchbase/cachedb_couchbase_dbase.c
+++ b/modules/cachedb_couchbase/cachedb_couchbase_dbase.c
@@ -42,51 +42,64 @@ volatile lcb_STATUS op_error  = LCB_SUCCESS;
 
 static void couchbase_get_cb(lcb_INSTANCE* instance,
 		const void *cookie, lcb_STATUS error,
-		const lcb_get_resp_t *item)
+		const lcb_RESPGET *item)
 {
 	op_error = error;
 
+	const char *key, *value;
+	size_t nkey, nvalue;
+	lcb_respget_key(item, &key, &nkey);
+
 	if (error != LCB_SUCCESS) {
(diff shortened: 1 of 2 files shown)


sobomax:
@lemenkov is there any interest to add some support and create community-driven semi-official out-of-the tree modules repository? I've floated this idea with @liviuchircu few months ago, he was not terribly excited but not dismissive either. FreeSwitch for example has one AFAIK.

lemenkov:
> @lemenkov is there any interest to add some support and create community-driven semi-official out-of-the tree modules repository? I've floated this idea with @liviuchircu few months ago, he was not terribly excited but not dismissive either. FreeSwitch for example has one AFAIK.

Honestly I am a bit worried about separating unpopular modules to the different repository. I'm afraid they will bitrot there slowly but steadily. I would rather merge it into main repository.

sobomax:
@lemenkov you got me bit wrong here. What I am talking about is to update
official module to v3 and pull v2 into an external repo. Where you can
maintain it (or anyone else) until it becomes irrelevant. We have
dispatcher_simple here for example in our private repo, one before they
switched it to db configuration few years ago. Which is in our use case a
bit redundant.

What makes it more interesting, perhaps we can see if this repo can be
compiled with multiple OpenSIPS versions, in which case that might be
additional compelling reason for people to use "extra modules" instead.

-Max

PR #2392, opened 2021-02-01
b2b_logic: support for rfc3515 NOTIFY and rollback after failed transfer

nikbyte:
Adds possibility to generate rfc3515 NOTIFY and not hangup call in case of failed transfer.

Instead of lifetime parameter there is flags parameter in b2b_bridge() function which can have next values:
`t` - set lifetime
`n` - enable NOTIFY
`f` - in case of negative reply during transfer, rollback the transfer attempt

### DIFF
--- a/modules/b2b_logic/logic.c
+++ b/modules/b2b_logic/logic.c
@@ -84,6 +84,7 @@ static str method_invite= {INVITE, INVITE_LEN};
 static str method_ack   = {ACK, ACK_LEN};
 static str method_bye   = {BYE, BYE_LEN};
 static str method_cancel= {CANCEL, CANCEL_LEN};
+static str method_notify= {NOTIFY, NOTIFY_LEN};
 
 static str ok = str_init("OK");
 static str notAcceptable = str_init("Not Acceptable");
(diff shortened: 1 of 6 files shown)


rvlad-patrascu (commented):
Hi @nikbyte ,

This generally looks ok but I have to point out that it is not correct to assume that the entity that sent the REFER (and has to be later notified), is always `tuple->servers[0]`. It might as well be a client entity. I think you'll have to save this entity separately in the tuple structure.

modules/b2b_logic/logic.c
```diff
@@ -1123,15 +1180,26 @@ int _b2b_handle_reply(struct sip_msg *msg, b2bl_tuple_t *tuple,
...
+		if(method_value == METHOD_NOTIFY) goto done1; /* Silently ignore reply on NOTIFY */
+
 		if(method_value != METHOD_INVITE)
 		{
 			LM_ERR("Wrong scenario state [B2B_BRIDGING_STATE] for this"
 				" reply(for method %d)\n", method_value);
 			goto error;
 		}
+
+		if(statuscode >= 200 && entity == tuple->bridge_entities[1]) /* Reply from new bridge entity */
```
rvlad-patrascu, line 1196: You will have to also check the `B2BL_BR_FLAG_NOTIFY` flag here. When we receive a reply in the bridging state, it is not always in a refer scenario and we might not need to send the notify.

modules/b2b_logic/logic.c
```diff
@@ -1123,15 +1180,26 @@ int _b2b_handle_reply(struct sip_msg *msg, b2bl_tuple_t *tuple,
...
+		{
+			process_bridge_notify(tuple->servers[0], cur_route_ctx.hash_index, msg);
+			if(statuscode == 200 || !(tuple->bridge_flags & B2BL_BR_FLAG_RETURN_AFTER_FAILURE))
+				b2bl_delete_entity(tuple->servers[0], tuple, tuple->hash_index, 1);
```
rvlad-patrascu, line 1200: Following the same logic as above, I think we should not delete this entity automatically, if we are not in a refer scenario.

modules/b2b_logic/logic.c
```diff
@@ -473,9 +474,25 @@ int process_bridge_dialog_end(b2bl_tuple_t* tuple, unsigned int hash_index,
...
-			b2b_end_dialog(tuple->bridge_entities[0], tuple, hash_index);
-			b2b_mark_todel(tuple);
+			if(tuple->bridge_flags & B2BL_BR_FLAG_RETURN_AFTER_FAILURE)
+			{
+				/* Delete failed entity */
+				b2bl_delete_entity(bentity, tuple, hash_index, 1);
+
+				/* Restore initial bridge */
+				tuple->bridge_entities[1] = tuple->bridge_entities[0];
+				tuple->bridge_entities[0] = tuple->servers[0];
```
rvlad-patrascu, line 484: Same idea as with the refer sending entity, we cannot assume that this is always a server entity.

nikbyte:
Hi Vlad. Thank you for reviewing. I've just pushed new commit which should fix these cases.

rvlad-patrascu (approved):

PR #2393, opened 2021-02-03
Refactor structure(s) packing code to do not put C structures at random unaligned offsets

sobomax:
One things that people often don't realize about heap management routines (malloc() etc), is that memory returned is ALWAYS aligned to provide memory that is suitable for placing C structures onto it. As such, one should not just allocate chunk of data and place structure at random offset within, at least if the data is then to be accessed without memcpy to a local variable. On some architectures such access is not allowed at all, on others just penalized performance-wise. Moreover, if memory is shared across cores, placing critical data, such as pointers, across cache line boundaries could also lead to broken value to be obtained by the core that reads the data shortly after another core has updated it.

### DIFF
--- a/modules/event_route/route_send.c
+++ b/modules/event_route/route_send.c
@@ -36,13 +36,16 @@
 int route_build_buffer(str *event_name, evi_reply_sock *sock,
 		evi_params_t *params, route_send_t **msg)
 {
-	route_send_t *buf;
+	struct {
+		route_send_t rt;
+		evi_param_t eps[0];
+	} *buf;
 	evi_param_p param, buf_param;
 	int len, params_len=0;
 	unsigned int param_no = 0;
 	char *s;
 
-	len = sizeof(route_send_t) + event_name->len;
+	len = sizeof(*buf) + event_name->len;
 	if (params) {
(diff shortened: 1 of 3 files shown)


liviuchircu (commented):
LGTM, @sobomax.  For me, particularly, it was an interesting cross-platform coding lesson, as I had no idea casting structures out of a contiguous buffer is problematic on 32-bit to begin with!  So, thank you for that!

sobomax:
> LGTM, @sobomax. For me, particularly, it was an interesting cross-platform coding lesson, as I had no idea casting structures out of a contiguous buffer is problematic on 32-bit to begin with! So, thank you for that!

No worries, @liviuchircu I'd not see the issue with that code myself few years ago. But this is not only affecting 32-bit arches, I should warn you. On the most popular platform x64 that could also lead to a nasty bugs when compiler generates fast vectorized code that *expects* data to be aligned. The fact that it has not happened to OpenSIPS is just a sheer pure luck.

https://pzemtsov.github.io/2016/11/06/bug-story-alignment-on-x86.html

Accessing unaligned data in C standard is regarded as UB (undefined behavior) and so it's generally a bad code to have.

PR #2396, opened 2021-02-03
tm: Be more verbose when async detected in non-request route

lemenkov:
If async operation is detected in non-request route then tm tries to
switch to sync mode emmitting just a debug message. Debug messages are
relatively rarely being turned on, so this switch to sync can be left
unnoticed.  Let's raise the severity to WARN so this situation can be
spotted much easier.


liviuchircu:
Solid improvement, that warning will be very helpful!

PR #2403, opened 2021-02-15
Add support for UUID versions 3 and 5

john08burke:
Adds UUIDv3 and UUIDv5 functionality to the uuid module.

The new method signature is:
`uuid(out_var, [version], [namespace], [name])`

There also seems to be a bit of an oddity with the return values. The existing docs mention it: 

> "If UUID version 1 is used, the function will return the value 2 if the UUID was generated in an unsafe manner."

IMO the return value should be some negative integer for unsafe UUID generation where possibility of collision exists. I did kept the existing behavior in place, but can add to the PR is the OpenSIPS team agrees.


rvlad-patrascu (approved):

rvlad-patrascu:
Hi @john08burke ,

Thanks for the contribution! Regarding the return value, I think that as long as the function does return an UUID, the (slim) collision possibility is not an error per se.

john08burke:
That sounds good to me @rvlad-patrascu! Let me know if you need anything further.

PR #2407, opened 2021-02-18
Use DBG level for per-connection nominal log messages

danlshields:
AWS health checks open a TCP connection very frequently to check on the status of the server. The per-connection INFO logs get spammed. This commit changes these messages to debug messages since that log level may be more appropriate for them anyway.


liviuchircu (approved):
LGTM, @danlshields!  The texts look like debug logs to begin with (not that important to an administrator, as they **expect** TCP keepalive to be on, per the global setting they have done).

PR #2409, opened 2021-02-22
Trace support for b2b_logic

nikbyte:
This allows to use trace() with b2b_logic.

Current limitations:
- Can't catch ACK on negative responses
- Can't catch incoming 100 Trying


bogdan-iancu:
Thank you @nikbyte  for the patch - before reviewing, could you share what is the intended usage of `trace()` in the context of the b2b module ? When and where is the tracing function supposed to be called ?

nikbyte:
It's intended to be called before b2b_init_request(). For example `trace("hep", "b", "sip");`.
When called at once, this is working until the tuple is alive, including all new entities.

modules/b2b_entities/dlg.c
```diff
@@ -783,12 +783,22 @@ int b2b_prescript_f(struct sip_msg *msg, void *uparam)
 	if(method_value == METHOD_PRACK)
 	{
 		LM_DBG("Received a PRACK - send 200 reply\n");
+
+		// Tracer callback: incoming PRACK and its transaction
+		if (dlg && dlg->tracer_param) // request
+			dlg->tracer_cback(msg, NULL, dlg->tracer_param, B2B_NONE);
+
```
bogdan-iancu, line 790: @nikbyte , this is wrong here as the `dlg` is not yet matched, is it still NULL. The matching against the `b2b_dlg_t` is done later, after the `search_dialog` label.
Without the `dlg` , I see no way to trace this PRACK.

modules/b2b_entities/dlg.c
```diff
@@ -964,8 +983,16 @@ int b2b_prescript_f(struct sip_msg *msg, void *uparam)
...
 
 				if(method_value == METHOD_BYE)
+				{
 					tmb.t_reply(msg, 200, &ok);
-				else
+					tm_tran = tmb.t_gett();
+					if(tm_tran && tm_tran!=T_UNDEFINED)
+					{
+						// Tracer callback: our reply on BYE
+						if (dlg && dlg->tracer_param)
```
bogdan-iancu, line 992: Here the tracing is also broken, as dlg is NULL all the time, we are on a `if(dlg== NULL)` branch

nikbyte:
@bogdan-iancu Thank you for reviewing and applying.

Issue #2413, opened 2021-02-23
[BUG] Dispatcher pvar_algo_pattern


john08burke:
**OpenSIPS version you are running**

```
devel, 3.1
```

**Describe the bug**
Starting in 3.1, using dispatcher module with algorithm 9 (pvar_algo) appears to be broken. Looking through the logs I see reference to algo 10 even though 9 is being used: `ERROR:dispatcher:ds_select_dst: No hash_route param provided`.

It looks like when algo 10 was added, `break` statement was missed for algo 9.

Snippet from 3.1:
https://github.com/OpenSIPS/opensips/blob/a98b3b5e009865cb1bd8e86d4a8e42d44a8a84cc/modules/dispatcher/dispatch.c#L1734-L1760

liviuchircu:
Good catch, @john08burke, your suggested fix was on the spot!

PR #2415, opened 2021-02-23
Add support for set id substitution for dispatcher algorithm 9

john08burke:
This PR adds support for replacing `%i` with the selected dispatcher set ID when using algorithm 9.  Currently, only support for URI substitution exists via `%u` string. However, if using the same URI in multiple sets then weighting can lead to unexpected behaviors as the same `$stat(...)` will be referenced/updated by multiple sets. By allowing use of `%i`, multiple sets with same URI(s) can be weighted independently.

Example usage:
```
modparam("dispatcher", "pvar_algo_pattern", "$stat(weight_%i_%u)")
...
...
$var(setid) = 1;
ds_select_domain($var(setid),9);

#prefer local dispatch
if ($(rd{ip.matches,MY_PUBLIC_IP/24}))
    update_stat("weight_$var(setid)_sip:$rd:$rp",1);
else
    update_stat("weight_$var(setid)_sip:$rd:$rp"",10);
```

The same type of weighting logic can be achieved by using algorithm 10, however, for large dispatcher sets algo 9 has the advantage of only calling weighting logic 1x per request as opposed to Nx per request with algo 10.

### DIFF
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -76,6 +76,11 @@ struct fs_binds fs_api;
 #define dst_is_active(_dst) \
 	(!((_dst).flags&(DS_INACTIVE_DST|DS_PROBING_DST)))
 
+enum ds_pattern_type {
+	DS_PATTERN_NONE=0,
+	DS_PATTERN_ID,
+	DS_PATTERN_URI,
+};
 
 int init_ds_data(ds_partition_t *partition)
 {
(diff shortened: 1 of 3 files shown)


john08burke:
I failed to mention, but I split out a relating bug fix into #2413.  I didn't include the fix in this PR, as I thought it would be easier to backport to 3.1... but you will need it to test :smile:

bogdan-iancu:
Thank you @john08burke

PR #2421, opened 2021-02-26
add  rl_dump_pipe MI command.

swk:
Dump underlying buckets for SBT rate limits.


bogdan-iancu:
Hi @swk , as I understand, this new `rl_dump_pipe` is just printing the details of the pipe, which, for non-SBT pipes, is the same as done by `rl_list` right? (the counter value only). If so, maybe we can find a way to merge these 2 functions ?

swk:
The only way to merge them would be to add another parameter to rl_list  as for SBT Pipes theres the  bucket values and the aggregate value. rl_list shows the aggregate values for the SBT pipes where rl_dump_pipe shows the underlying data.

bogdan-iancu:
@swk , ok, it makes sense, let's keep them separately for now

PR #2423, opened 2021-02-27
db _postgres fix wrong variable type when escape characters

nikbyte:
Hello,

I've found a bug which causes a crash of db_postgres module.

Instead of using VAL_STR() macros for DB_STR type there mistakenly was used VAL_STRING().

Here is the fix.

### DIFF
--- a/modules/db_postgres/val.c
+++ b/modules/db_postgres/val.c
@@ -263,21 +263,21 @@ int db_postgres_val2str(const db_con_t* _con, const db_val_t* _v,
 
 	case DB_STR:
 		l = VAL_STR(_v).len;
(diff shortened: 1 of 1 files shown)


nikbyte:
@rvlad-patrascu btw, this breaks b2b_logic when writing to postgresql.

nikbyte:
@rvlad-patrascu could you please provide more details? for me it's fixed crashes when use b2b+pgsql.

razvancrainea:
The same bug has been fixed by PR #2591 which does the same thing (it's based on this PR), but I think the code there is cleaner, that's why I went in favor of that one.

PR #2426, opened 2021-02-28
b2b: properly release tuples

nikbyte:
Release b2b tuples properly which prevents corrupted data usage.

### DIFF
--- a/modules/b2b_logic/entity_storage.c
+++ b/modules/b2b_logic/entity_storage.c
@@ -412,9 +412,12 @@ static void receive_entity_create(enum b2b_entity_type entity_type,
 
 	return;
 error:
-	lock_release(&b2bl_htable[hash_index].lock);
 	if (tuple && !old_tuple)
+	{
 		b2bl_delete(tuple, hash_index, 0, 0);
+		tuple = 0;
+	}
+	lock_release(&b2bl_htable[hash_index].lock);
 	if (entity) {
 		if (entity->dlginfo)
 			shm_free(entity->dlginfo);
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Hi @nikbyte , could you please provide some comments explaining the issues you found and the fixes you did - this will help me understand your PR and validate it is faster. Thanks for the great job!

nikbyte:
I expected random crashes. All of them were related to existing tuple or dlg pointers which were pointed to not allocated or random memory. This occurs usually on BYE or final ACK. For example, try to write garbage to database in b2b_entities and b2b_logic module.

nikbyte:
After these two patches there are no any crashes by this reason.

modules/b2b_logic/entity_storage.c
```diff
@@ -412,9 +412,12 @@ static void receive_entity_create(enum b2b_entity_type entity_type,
 
 	return;
 error:
-	lock_release(&b2bl_htable[hash_index].lock);
```
rvlad-patrascu, line 415: Moving this lock release after the `b2bl_delete()` function is the only change from these commits that has any effect. The rest of the changes are essentially NOPs.

rvlad-patrascu (changes requested):
Can you simplify this commits by only leaving that relevant change?

modules/b2b_logic/entity_storage.c
nikbyte: I tested only this change and it doesn't work on high load. What I offered works only for me. I debugged the code and crashes are because of structure not nulled.

rvlad-patrascu: Maybe when you tested you had also other changes? Becuase setting `tuple = 0;` at the points in code where you did, only writes a local variable that is no longer read after that.

nikbyte: I can't comment now because I investigated this for a long time ago,
but I tried to remove this few times and it definitely helped.
Now it's working with this patch in production and doesn't crash.

It only starts to affect on high CPS, let's say more than 400.
I had many crashes and this fixed for me.

If you think it's useless, I trust you. We'll see later how it works on latest b2b.

rvlad-patrascu: I was just letting you know in case you wanted to update the PR so I could merge it. But I'm also ok with closing this PR and pushing the fix myself.

nikbyte: Feel free to close it. 🙂

PR #2432, opened 2021-03-09
[freeswitch] Fix memory leak with Sockets not being cleaned up.

zooptwopointone:


modules/freeswitch/fs_api.c
```diff
@@ -146,6 +146,9 @@ void evs_free(fs_evs *sock)
 	shm_free(sock->host.s);
 	shm_free(sock->user.s);
 	shm_free(sock->pass.s);
+	if (sock->handle) {
+	  pkg_free(sock->handle);
+	}
```
liviuchircu, line 151: Could you just do `pkg_free(sock->handle);`, please?  The function has internal NULL-checking anyway, there is no need for the `if` :)

liviuchircu (changes requested):

liviuchircu:
Never mind my comment above, @zooptwopointone, just fixed it for you.  Thank you for the catch!

PR #2434, opened 2021-03-11
Make opensips behave consistently in respect of nsswitch regardless of DNS caching.

bambyster:
The current integration with dns_cache module has an issue. When dns_cache module is disabled the OpenSIPS uses gethostbyname(), gethostbyname2(), gethostbyaddr() functions from libc which respect the nsswitch.conf settings and for example automatically perform a lookup in /etc/hosts. However when dns_cache is enabled the OpenSIPS starts to use res_search() instead, which only performs DNS lookups and as a result it becomes impossible to query other name resolution services configured in the nsswitch.conf.

I propose to use gethostby*() functions consistently regardless of caching settings.

### DIFF
--- a/resolve.c
+++ b/resolve.c
@@ -80,14 +80,8 @@ static struct bl_head *failover_bl=0;
 #define DNS_REVOLVER_BL_NAME  "dns"
 #define DNS_BL_EXPIRE         4*60
 
-#define MAX_BUFF_SIZE	8192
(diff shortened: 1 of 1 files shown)


sobomax:
The problem with this patch is that we don't have any TTL information when using gethostbyXXX() functions. :( As such it breaks invalidation of the cache.

Issue #2118, opened 2020-05-21
[CRASH] usrloc module


adigeo:
**OpenSIPS version you are running**

```
version: opensips 3.1.0-dev (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 7f16a1cdb
main.c compiled on  with gcc 9
```

**Crash Core Dump**

Core was generated by `/usr/sbin/opensips -w /run/opensips -P opensips.pid -a F_MALLOC_DBG -m 128'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc0ed15f376 in wb_timer (_r=_r@entry=0x7fc0eeb5a870, ins_list=0x7fc0eeb4c168) at urecord.c:345
345					VAL_BIGINT(cid_vals+cid_len) = t->contact_id;
(gdb) bt
#0  0x00007fc0ed15f376 in wb_timer (_r=_r@entry=0x7fc0eeb5a870, ins_list=0x7fc0eeb4c168) at urecord.c:345
#1  0x00007fc0ed15f920 in timer_urecord (_r=_r@entry=0x7fc0eeb5a870, ins_list=ins_list@entry=0x7fc0eeb4c168) at urecord.c:439
#2  0x00007fc0ed14d232 in mem_timer_udomain (_d=<optimized out>) at udomain.c:1166
#3  0x00007fc0ed167489 in _synchronize_all_udomains () at dlist.c:1068
#4  0x00007fc0ed16d76c in destroy () at ul_mod.c:517
#5  0x0000557873eff15c in destroy_modules () at sr_module.c:544
#6  0x0000557873ef8126 in cleanup (show_status=1) at main.c:363
#7  0x0000557873ef88fc in shutdown_opensips (status=<optimized out>) at main.c:527
#8  0x0000557873ef904b in handle_sigs () at main.c:610
#9  0x0000557873e64ae8 in main_loop () at main.c:867
#10 main (argc=<optimized out>, argv=<optimized out>) at main.c:1488

[...]

PR #2439, opened 2021-03-18
Restore registrar 'f' flag behaviour for "sql-only" db_mode

### LINKED ISSUE #2372: [BUG] REGISTER problem when using save() method with "f" flag
### OpenSIPS version you are running

```
version: opensips 3.1.0 (x86_64/linux)
git revision: 867fb320a
main.c compiled on 10:49:32 Nov  6 2020 with gcc 8
```

### Describe the bug

When handling a **REGISTER** request with the `save()` method and the `fc1` flags:

```tcl
# Save contact internally, the argument 'location' is the name of the
# database table that stores the contacts. The 'f' (force registration) flag
# is used to force the registration of NEW contacts even if the maximum
# number of contacts is reached. In such a case, older contacts will be
# removed to make space to the new ones, without exceeding the maximum
# allowed number.

if (! save("location", "fc1")) {
    # Send back an error reply describing the nature of the last internal
    # error.
    sl_reply_error();
    exit;
}
```

I get more than **1** contact per **AOR** in the `location` table of the database, for example, when registering two softphones with the same username and domain. For what I have understood, it should store only one contact, the last one that have been sent with a **REGISTER** request and delete any other existing contact on the database for that **AOR**, as per explained on the [save() documentation and the f flag](https://opensips.org/docs/modules/3.1.x/registrar.html#func_save). Even more, if I remove the **f** flag to get `save("location", "c1")` the latest **REGISTER** won't succeed, since `c1` flag restricts to only one contact per `AOR` (note this one is an expected behavior). However, the `f` flag is not working as documented, it enables new registrations for an existing **AOR** but do not deletes the previous ones.

### To Reproduce

1. Start OpenSIPS
2. Register a first Softphone with a sip uri like: username@my.domain.com
3. Register a second Softphone with the same sip uri: username@my.domain.com
4. Review the `location` table of the database.

### Expected behavior

The call to `save("location", "fc1")` should accept a new `REGISTER` having a new **contact** for an existing **AOR** and delete previously existing contacts for that **AOR**, leaving only one on the `location` table of the database.

### Relevant System Logs

At next, you can see an example of the `location` table after receiving two `REGISTER` from different softphones and the same `AOR`:

```sh
MariaDB [opensips]> select contact_id, username, domain, contact, expires, callid, user_agent from location;
+------------+----------+-----------------+-----------------------------------------------------------------------+------------+--------------------------+-----------------------------------+
| contact_id | username | domain          | contact                                                               | expires    | callid                   | user_agent                        |
+------------+----------+-----------------+-----------------------------------------------------------------------+------------+--------------------------+-----------------------------------+
|        162 | 1000     | my.domain.com   | sip:1000@181.99.71.249:38463;transport=UDP;rinstance=c0170f0c5f811628 | 1610761869 | w5dxiuaeR4lVDzlTVaFxbw.. | Z 5.4.9 rv2.10.11.7               |
|        163 | 1000     | my.domain.com   | sip:1000@181.99.71.249:5060;transport=udp                             | 1610761949 | EK2n94GVCx               | Linphone/3.12.0 (belle-sip/1.6.3) |
+------------+----------+-----------------+-----------------------------------------------------------------------+------------+--------------------------+-----------------------------------+
```

My expectations were to only have the record with `contact id=163`, since it was part of the last **REGISTER** request received.

### OS/environment information
 - Operating System: CentOS Linux release 8.2.2004 (Core) 
 - OpenSIPS installation: git clone https://github.com/OpenSIPS/opensips.git -b 3.1 opensips_3_1

### Additional context

The **REGISTRAR MODULE** configuration is:

```tcl
loadmodule "registrar.so"
modparam("registrar", "max_contacts", 1)
modparam("registrar", "received_avp", "$avp(received_nh)")
modparam("registrar", "tcp_persistent_flag", "TCP_FLAG")
```

The **USRLOC MODULE** configuration is:

```tcl
loadmodule "usrloc.so"
modparam("usrloc", "nat_bflag", "NAT")
modparam("usrloc", "working_mode_preset", "sql-only")
modparam("usrloc", "db_url", "mysql://some-database-url")
```

I'm also using the **NATHELPER** module and methods `fix_nated_contact()` and  `fix_nated_register()` when a device is detected behind a NAT, just in case.

If you need further information, just ask. Also, I'm interested in any workaround that can be used from the routing script in order to get it work at expected.

Thanks in advance!

wdoekes:
On 2020-05-25 fix 3a1f9c1e2 for #2118 was introduced into opensips 3.1.
This change also modified things so cid_keys and cid_vals would *not*
be allocated when the database was in "sql-only" mode.

When "sql-only" mode is used, all usrloc records are saved to the DB
immediately, and are loaded from DB immediately.

The relevant settings:

    modparam("usrloc", "working_mode_preset", "sql-only")

Which equals the deprecated db_mode setting 3:

    modparam("usrloc", "db_mode", 3)  # 3 = "sql-only", DEPRECATED

Or, as three separate settings:

    modparam("usrloc", "cluster_mode", "sql-only")     # CM_SQL_ONLY
    modparam("usrloc", "restart_persistency", "none")  # RRP_NONE
    modparam("usrloc", "sql_write_mode", "none")       # SQL_NO_WRITE
    # This last option is slightly awkwardly named, but it refers
    # to *additional* SQL writing which is not performed. Because,
    # the is only *direct* SQL reading/writing.

In any case: the changeset removed the access to temporary storage of
cid_keys/cid_vals for the CM_SQL_ONLY case, under the assumption that
it was not needed:
```patch
       if (rr_persist == RRP_LOAD_FROM_SQL) {
          if (!(sync_lock = lock_init_rw())) {
             LM_ERR("cannot init rw lock\n");
             return -1;
          }
-      }
-   }

-   if (cluster_mode != CM_NONE || rr_persist == RRP_LOAD_FROM_SQL) {
-      cid_keys = pkg_malloc(max_contact_delete *
-            (sizeof(db_key_t) + sizeof(db_val_t)));
-      if (!cid_keys) {
-         LM_ERR("oom\n");
-         return -1;
-      }
+         /* initialize the "merged contact deletes" array */
+         cid_keys = pkg_malloc(max_contact_delete *
+               (sizeof(db_key_t) + sizeof(db_val_t)));
+         if (!cid_keys) {
+            LM_ERR("oom\n");
+            return -1;
+         }
```
However, those "merged contact deletes" were still needed when the
registrar save() function tries to limit the number of Contacts (using
the 'f' force overwrite flag).

registrar:add_contacts() calls usrloc:release_urecord() with the
contacts that should be removed (from the database). This calls
usrloc:db_only_timer() (and wb_timer) and finally
usrloc:db_multiple_ucontact_delete().

The wb_timer was supposed to populate the cid_keys/cid_vals with the
database IDs passed along from the registrar module. But because that
temporary cid_vals storage was not available, that bit would be skipped:
```c
            /* Should we remove the contact from the database ? */
            if (cid_vals && st_expired_ucontact(t) == 1
```
The result: usrloc:db_multiple_ucontact_delete() had nothing to
delete; the max contacts save settings would be ignored.

This changeset restores the `cluster_mode != CM_NONE` alternative that
initialises cid_keys/cid_vals, restoring the force/max_contacts
behaviour.

(Closes #2372. @liviuchircu: care to take a look?)

### DIFF
--- a/modules/usrloc/ul_mod.c
+++ b/modules/usrloc/ul_mod.c
@@ -908,7 +908,7 @@ int ul_check_db(void)
 			return -1;
 		}
 
-		if (rr_persist == RRP_LOAD_FROM_SQL) {
+		if (cluster_mode != CM_NONE || rr_persist == RRP_LOAD_FROM_SQL) {
 			if (!(sync_lock = lock_init_rw())) {
 				LM_ERR("cannot init rw lock\n");
 				return -1;


liviuchircu (approved):
Excellent catch and fix, @wdoekes!  This one definitely slipped through the cracks, I had `CM_SQL_ONLY` in mind when I edited [this line](https://github.com/OpenSIPS/opensips/blob/3.1.1/modules/usrloc/ul_mod.c#L890) in the same commit!  However, it must have eluded me that `rr_persist` is zeroed in that mode.

Also, I'd argue that my bug also broke the contact deletion of the `remove()` function, as well as the `ul_rm_contact` MI command when using the `sql-only` preset.

Many thanks!

PR #2441, opened 2021-03-21
Fix memory cleak from freeswitch events.

zooptwopointone:
Fix memory leak for Freeswitch module


liviuchircu (commented):
Looks good, good catch!

PR #2442, opened 2021-03-24
Implement RFC8760 digest authentication. 

sobomax:
This PR adds SHA-256 and SHA-512/256 as a supported authentication methods as well as unifies UAC/UAS digest calculation code to remove duplication. Few bonuses are included:

- "-sess" variants of authentication has been fixed to work properly in both UAC and UAS scenarios. Both have been broken in various places since pretty much forever;

- bin-to-hex generation routine has been improved to give some 10x speedup on modern (intel/arm) 64-bit CPUs using vector instructions (AVX/whatever). Considering that we do few rounds of those as part of digest generation/verification of a single response this should at least offset cost of generating multiple challenge fields and verifying loner and longer digests somewhat;

- digest verification routine has been reworked to operate on digest binary comparing it to provided hex representation using vectorizable code, instead of converting binary to hex first (using old slow method), saving it to memory and then comparing two hex strings. 

- algorithm to hide timestamp inside nonce (preferred method to prevent replay) has been reworked to use AES instead of just sending timestamp/index in plain and do a rather weak "signature" using md5 hash+server side secret. Few more bits have been added into concealed space to also prevent "downgrade" attack on a SHA-256 / SHA-512/256 challenge and QOP type. High-res system timer is now used to generate timestamps, instead of number of seconds since epoch to give resulting AES-encrypted nonce 20-more somewhat random bits (very hard to guess to be precise);

- "ha1b" hack from 2002-2003 has been removed for good. As discussed with @liviuchircu and @bogdan-iancu;

- since security of the scheme relies on good random nonces, the seeding of the RNG has been moved into child_init() of the UAS module, so each opensips instance has its own independent random sequence;

- nonce encoding scheme has been changed from hex to base64, which provides better density of bits (6 per character, vs. 4) and reduces the overhead of including multiple challenge headers and/or 2x longer nonce on 40x challenge response size.


liviuchircu:
For any other brave developers who intend to review this PR and have a hard time working with the 3600-line diff, here are the steps to pull all of the underlying 273 commits which have been squashed into a single, giant one:

```
# checkout the proper RFC8760 branch
git checkout -b sippy-RFC8760 master
git pull https://github.com/sippy/opensips.git RFC8760

# build the review-TODO file, the "====" line works as a marker (commits above the marker are considered "reviewed")
echo "=======" >/var/tmp/RFC8760-code-review
git log origin/HEAD..HEAD --oneline | tac | grep -v "Merge remote-tracking branch" >>/var/tmp/RFC8760-code-review

# resume code reviewing at any time (after your "====" marker in the file, which you can manually move)
exec 3<&0; while read line; do git show $(echo $line | awk '{print $1}'); read -p'Continue? ' _ <&3 ; done < <(grep -v "===" /var/tmp/RFC8760-code-review | tail +$(grep -n '===' /var/tmp/RFC8760-code-review | awk -F: '{print $1}')); exec 3<&-
```

@sobomax: I cannot say how long the review will take, but it's definitely going to be at least a couple of days.

sobomax:
Thanks, @liviuchircu . I should also mention that few scenarios have been created to test various bits and pieces, including UAS (simple), UAC, UAS (db) and tested against Sippy B2BUA code using all 6 variants (MD5, SHA-256, SHA-512-256, MD5-sess, SHA-256-sess and SHA-512-256-sess) here:

https://github.com/sippy/voiptests/actions/runs/622511700

The new vectorized code has also been verified to work properly on big-endian (mips64) and 32-bit arches as part of the my ongoing effort to add unit tests rounds into CI.

https://github.com/sippy/opensips/actions/runs/532478940

liviuchircu (approved):
Managed to wrap up the review, here are a few comments:

31c8847fe - I still maintain my original opinion that fixing the `LOWER1B()` macro to be 100% correct isn't really worth the trade-off of introducing extra `if` assembly code in all digest parameter comparisons.  So what if we keep the old behavior of matching a false positive such as `"SHAM256M512"` or `"SHAM256Msess"`? (because `'-' | 0x20 = 'M'`)  It's just an easter-egg: not a critical bug, not an RFC non-compliance.  Also note that this macro is only used for the authentication headers, not elsewhere in code.
22d0b2c81 - Do you consider this reverse-search to be optimal for the upcoming decade?  I would assume the most frequent search order to still remain "MD5 -> SHA256 -> others" for many more years to come, so why not optimize the iteration order to take advantage of this?
8983c047e - isn't uintptr_t the proper choice for a bitmask, instead of intptr_t?
aac769cf0 - not sure if you've familiarized yourself with the idea behind catching `SIGSEGV` in opensips.  I noticed you later reverted this commit (so all good here!), but in short:  the idea is to prevent the attendant process from `SIGKILL`'ing a worker while it's dumping a corefile!  So by catching the handler and marking itself with a flag, a worker process is now able to freely write the corefile without fear of truncation if the disk is too slow, and the attendant starts forcefully terminating all processes because it's 60-second shutdown alarm rang out.

All in all, this is a great set of changes and I'm eager to see them merged, especially the turbo-comparison, hex conversion and other optimized routines!

Congratulations on a job very well done, @sobomax!

PR #2447, opened 2021-03-26
Fix for Makefile.defs

bambyster:
I was working on a new FreeBSD port for opensips and found a bug in the Makefile.defs which prevented LDFLAGS from including additional library search directory. The current order of the ifeq/else/endif is:

ifeq (,$(CFLAGS)) : line 826
ifeq ($(mode), release) : line 832
endif   # CFLAGS not set : line 1330
else    #mode,release : line 1362
endif #mode=release : line 1389

so I've changed the order of ifeqs and my build succeeded.

Another important feature which is needed by FreeBSD port is ability to override variables INSTALL_BIN, INSTALL_MODULES, etc so it is possible to be able to strip binaries of debug symbols if needed.

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -823,13 +823,13 @@ ifeq	($(CC_NAME), clang)
 	CC_OPTIMIZE_FLAG?=-O3
 endif
 
+ifeq ($(mode), release)
 ifeq (,$(CFLAGS))
 
 #common stuff
 CFLAGS+=$(DEBUGSYM)
 
 # setting more CFLAGS
-ifeq ($(mode), release)
 
 #if i386
 ifeq	($(ARCH), i386)
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -462,11 +462,11 @@ endif
 
 INSTALL_TOUCH = touch          # used to create the file first (good to
                                #  make solaris install work)
-INSTALL_CFG = $(INSTALL) -m 644
-INSTALL_BIN = $(INSTALL) -m 755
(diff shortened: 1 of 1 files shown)


razvancrainea:
Not sure I understand the underlying problem, but your PR changes the logic of the `release` mode. Can you explain exactly what was the problem if the `if`s are not reversed?

bambyster:
My problem was with the MOD_LDFLAGS not being set properly when I pass CFLAGS as an environment variable. So if CFLAGS is set then the section which initializes the MOD_LDFLAGS is never reached:
```
ifeq (,$(CFLAGS))
    ifeq ($(mode), release)
    endif # CFLAGS not set
    ...
    MOD_LDFLAGS=-shared $(LDFLAGS) : lines 1335-1377
    ...
else #mode,release
endif #mode=release
```
As you see the comments on elses and endifs contradict the actual logic in the Makefile.defs.

razvancrainea:
I somehow think the problem is somewhere else - an unhandled architecture or compiler.
What's the compiler you are using?

bambyster:
I'm using clang.

But still my main point here is that the real logic in the Makefile.defs differs from the logic stated in comments in the Makefile.defs. Which is why I tried to modify it to match the comments.

razvancrainea:
You are actually right, the comments are indeed reversed, and I think your PR is valid.
I was initially thinking you hit another bug I've noticed - in `mode=debug`, there's no handling for `clang`, but most likely this is a different bug.
Thanks for the PR, will merge it now.

razvancrainea:
sorry, accidentally closed it.

PR #2453, opened 2021-03-30
Update ip_addr.c

PBXForums:
Update RFC6598 address space

The current /22 is causing problems NAT traversal problems using space outside 100.64.0.0/22

RFC6598 para 7 defines the cgnat address space:
---------------------------------------------------------------------
7.  IANA Considerations

   IANA has recorded the allocation of an IPv4 /10 for use as Shared
   Address Space.

   The Shared Address Space address range is 100.64.0.0/10.


PBXForums:
Liviu, its Darren from another place, there is a ticket open for nat uac test 16 problem.

PR #2469, opened 2021-04-13
Make FORCE_SINGLE_REGISTRATION work again

wdoekes:
In 5b551ba35f, an always-false expression was added. The code was probably supposed to be as follows..

(REGISTERING_STATE == 1, AUTHENTICATING_STATE == 2)

----

@rvlad-patrascu, care to take a look?


wdoekes:
Don't forget this super easy fix :tada:

bogdan-iancu:
Thank you @wdoekes

PR #2470, opened 2021-04-13
Change CON_SET_CURR_PS pattern to always call it _directly_ before db call

wdoekes:
_This is a continuation of #2468._

_(It has that commit, and adds the following commit on top of that.)_

----

Because CON_SET_CURR_PS is always followed by an immediate call to CON_RESET_CURR_PS inside the DB function, it makes sense to call CON_SET_CURR_PS just before the DB call.

Otherwise you could get this:
```c
   CON_SET_CURR_PS(...);

   for (i = 0; i < n; ++i) {
       // first insert uses prepared statement
       db.insert(...);
       // but second time in the loop, we use "plain" text queries
   }
```
The updated pattern would look like this:
```c
   for (i = 0; i < n; ++i) {
       // the ps handle is always reset, so we reinit it every time
       CON_SET_CURR_PS(...);
       db.insert(...);
   }
```
Note that we don't need any explicit CON_RESET_CURR_PS() at odd points anymore (like was added in 57caa6c). A SET is immediately followed by an implicit RESET.

### DIFF
--- a/db/db_insertq.c
+++ b/db/db_insertq.c
@@ -131,11 +131,10 @@ void flush_query_list(void)
 			//Reset prepared statement between query lists/connections
 			my_ps = NULL;
 
-			CON_SET_CURR_PS(it->conn[process_no], &my_ps);
-
 			/* and let's insert the rows */
 			for (i=0;i<it->no_rows;i++)
 			{
+				CON_SET_CURR_PS(it->conn[process_no], &my_ps);
 				if (it->dbf.insert(it->conn[process_no],it->cols,it->rows[i],
 							it->col_no) < 0)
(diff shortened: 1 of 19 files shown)


wdoekes:
(Conflicts resolved.)

wdoekes:
(Conflicts resolved through the web interface. Let me know if you want it properly squashed+forcepushed instead.)

wdoekes:
(And properly force pushed this time, for cleaner git history.)

wdoekes:
:bell: Nice and shiny PR here :bell: 

I don't think this one is that questionable, is it?

bogdan-iancu:
Thank you @wdoekes , good catch !

PR #2479, opened 2021-04-21
Fix cookie collision when using rtpengine/rtpproxy modules

### LINKED ISSUE #2459: [BUG] Duplicate rtpengine cookies
```
version: opensips 3.1.1 (x86_64/linux)
main.c compiled on 02:22:09 Mar 25 2021 with gcc 8
```

**Describe the bug**
When running OpenSIPS inside of docker, I'm experiencing colliding rtpengine cookies between dockerized OpenSIPS instances that are using the same set of rtpengine servers.  In such a case, OpenSIPS logs `ERROR:rtpengine_offer_answer: failed to extract sdp body from proxy reply` and users experiences audio issues.  Looking at the rtpengine code, the only way `sdp` is not offered is when duplicate cookie is detected.

In the OpenSIPS `gencookie` snippet, PID is used to prefix the cookie value. The issue is in a dockerized environment, PIDs are likely to be the same between instances. This makes collisions more likely than traditional VMs, and is frequent right after doing an upgrade of an OpenSIPS cluster for example.  IMO, the prefix should be a random value instead of PID.

https://github.com/OpenSIPS/opensips/blob/ac2ce9960ab13c6a50110edc585e497572fef8da/modules/rtpengine/rtpengine.c#L1579-L1586

Example of PIDs between two separate dockerized OpenSIPS below:

Instance 1 `mi ps`:
```json
        {
            "ID": 5,
            "PID": 108,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 6,
            "PID": 109,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 7,
            "PID": 110,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 8,
            "PID": 111,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 9,
            "PID": 112,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 10,
            "PID": 113,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 11,
            "PID": 114,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 12,
            "PID": 115,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        }
```
Instance 2 `mi ps`:
```json
{
            "ID": 5,
            "PID": 109,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 6,
            "PID": 110,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 7,
            "PID": 111,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 8,
            "PID": 112,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 9,
            "PID": 113,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 10,
            "PID": 114,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 11,
            "PID": 115,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        },
        {
            "ID": 12,
            "PID": 116,
            "Type": "SIP receiver udp:xxx.xxx.xxx.xxx:5060"
        }
```

Let me know if you need more info!

**To Reproduce**
This is tough to reproduce, but generally speaking:
1. Start multiple dockerized OpenSIPS
2. Start traffic w/ rtpengine engaged
3. Periodically logs will show: `ERROR:rtpengine_offer_answer: failed to extract sdp body from proxy reply`

**Expected behavior**
OpenSIPS instances should generate unique enough cookies to not collide with other OpenSIPS instances.

john08burke:
When using Docker, PIDs often are the same between OpenSIPS instances and cannot be used as a good source for cookie prefix. Use prefix with random influence instead.

Fixes #2459

### DIFF
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -320,7 +320,7 @@ static int rtpengine_stats_used = 0;
 static int rtpengine_disable_tout = 60;
 static int rtpengine_retr = 5;
 static int rtpengine_tout = 1;
-static pid_t mypid;
+static int myprefix = 0;
 static unsigned int myseqn = 0;
 static str extra_id_pv_param = {NULL, 0};
 static char *setid_avp_param = NULL;
@@ -1500,7 +1500,7 @@ static int connect_rtpengines(void)
 static int
 child_init(int rank)
 {
-	mypid = getpid();
+	myprefix+=getpid()+rand()%10000;
 
 	if(*rtpe_set_list==NULL )
 		return 0;
@@ -1580,7 +1580,7 @@ static char * gencookie(void)
 {
 	static char cook[34];
 
-	sprintf(cook, "%d_%u ", (int)mypid, myseqn);
+	sprintf(cook, "%d_%u ", myprefix, myseqn);
 	myseqn++;
 	return cook;
 }
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -320,7 +320,8 @@ static int rtpengine_stats_used = 0;
 static int rtpengine_disable_tout = 60;
 static int rtpengine_retr = 5;
 static int rtpengine_tout = 1;
-static int myprefix = 0;
+static pid_t mypid;
+static int myrand = 0;
 static unsigned int myseqn = 0;
 static str extra_id_pv_param = {NULL, 0};
 static char *setid_avp_param = NULL;
@@ -1500,7 +1501,8 @@ static int connect_rtpengines(void)
 static int
 child_init(int rank)
 {
-	myprefix+=getpid()+rand()%10000;
+	mypid = getpid();
+	myrand = rand()%10000;
 
 	if(*rtpe_set_list==NULL )
 		return 0;
@@ -1580,7 +1582,7 @@ static char * gencookie(void)
 {
 	static char cook[34];
 
-	sprintf(cook, "%d_%u ", myprefix, myseqn);
+	sprintf(cook, "%d_%d_%u ", (int)mypid, myrand, myseqn);
 	myseqn++;
 	return cook;
 }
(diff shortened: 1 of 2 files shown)


modules/rtpengine/rtpengine.c
```diff
@@ -1500,7 +1500,7 @@ static int connect_rtpengines(void)
 static int
 child_init(int rank)
 {
-	mypid = getpid();
+	myprefix+=getpid()+rand()%10000;
```
razvancrainea, line 1503: Giving a deeper though about it, this approach might increase collision between instances that use different PIDs, or even between the same OpenSIPS instance, it increases the chances when `PID1+rand1 = PID2+rand2`. 
So probably the best idea is a combined one - have the cookie as `PID_RAND_MYSEQN`. If you don't have time to work this out by tomorrow (when we have a freeze), I will merge your PR and make the change myself, but if you do, we would all be grateful for your work :). Thanks!

john08burke: That's a good point, it could lead to the same situation in some cases. I will push your proposed schema shortly.

razvancrainea:
Awesome, I've backported the change on all branches. Thanks for helping us out!

PR #2481, opened 2021-04-22
Don't forget to release the insertq lock when downscaling workers

wdoekes:
Not sure if this is a problem or of the locks gets freed correctly after exit(). But it looks nicer this way.

----

I'm trying to find a root cause for the problem that causes #2467 and I bumped into this. It may very well not be a problem. But cleaning up after yourself is never a bad idea.

(Somewhere in the diff between 2.4 and 3.1, there is something that changes mysql reconnect stability, it seems. We now see lots of `error while submitting query` on lightly loaded (acceptance) hosts, which in turn likely causes the crash with the uninitialized op_t.)

### DIFF
--- a/db/db_insertq.c
+++ b/db/db_insertq.c
@@ -643,6 +643,8 @@ void ql_force_process_disconnect(int p_id)
 				it->dbf.close(it->conn[p_id]);
 				it->conn[p_id]=NULL;
 			}
+
+			lock_release(it->lock);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


wdoekes:
@bogdan-iancu: since you wrote this code, maybe you have some insight why the lock_release was omitted?

bogdan-iancu:
thanks @wdoekes , I think the missing unlock was a bug

PR #2485, opened 2021-04-25
add_uri_param can add $var()

wangduanduan:
add_uri_param can add $var():

```
$var(action) = "action=make";
add_uri_param("$var(action)");
```


bogdan-iancu:
@wangduanduan , the existing stable releases (like 2.4.x) do not accept any new features, only bug fixes. So, your PR cannot be accepted. But keep in mind that the starting 3.1 the issue you were addressing here was fixed.

PR #2487, opened 2021-04-26
dialog: Dont walk through empty dialog table

lemenkov:
We've got a following stacktrace:

(gdb) bt
.0  0x00007fba1e0edceb in _unref_dlg (dlg=0x7fba24689470, cnt=1) at dlg_hash.c:1007
.1  0x00007fba216c4a3e in empty_tmcb_list (head=head@entry=0x7fba26ab5da0) at t_hooks.c:53
.2  0x00007fba2169f978 in free_cell (dead_cell=0x7fba26ab5d30) at h_table.c:127
.3  0x00007fba216a24c3 in free_hash_table () at h_table.c:357
.4  0x00007fba216957d5 in tm_shutdown () at t_funcs.c:91
.5  0x0000555a8c6e4aec in destroy_modules () at sr_module.c:544
.6  0x0000555a8c6deb47 in cleanup (show_status=1) at main.c:366
.7  0x0000555a8c6df4cc in shutdown_opensips (status=<optimized out>) at main.c:530
.8  0x0000555a8c6dfe4a in handle_sigs () at main.c:568
.9  0x0000555a8c64d754 in main_loop () at main.c:868
.10 main (argc=<optimized out>, argv=<optimized out>) at main.c:1491
(gdb) p *d_table
Cannot access memory at address 0x0
(gdb)

Looks like dialog module was unloaded already. This means that OpenSIPS
already called `mod_destroy` in dialog.c and `destroy_dlg_table`. This
means `d_table` was assigned to `NULL`.

### DIFF
--- a/modules/dialog/dlg_hash.c
+++ b/modules/dialog/dlg_hash.c
@@ -1000,6 +1000,9 @@ void _ref_dlg(struct dlg_cell *dlg, unsigned int cnt)
 
 void _unref_dlg(struct dlg_cell *dlg, unsigned int cnt)
 {
+	if (d_table==0)
+		return;
+
 	struct dlg_entry *d_entry;
 
 	d_entry = &(d_table->entries[dlg->h_entry]);


liviuchircu:
Hi, Peter!  You discovered quite an interesting problem!  While I agree your solution will get the job done, the extra `if` will slow down the entire ref/unref mechanism of dialog, for all modules using the dialog API.  Also, most of the calls are done under lock, so it might be worth looking into the following solution:

What if we extended the module dependencies engine, so it not only invokes the `mod_init()` callback in a orderly fashion, but also `mod_destroy()`?  This way, since `dialog` depends on `tm` (see [here](https://github.com/OpenSIPS/opensips/blob/master/modules/dialog/dialog.c#L474)), `tm` will both _initialize_ and _destroy_ first, thus fixing the issue without introducing any overhead at runtime.

PS: for your platform, as a **dirty** quick fix, I am 90%+ sure that if you re-order `dialog` to load before `tm` in your `opensips.cfg` script, they will destroy in the _reverse_ order, thus `tm` will still find the `d_table` dialog table pointer.

lemenkov:
> What if we extended the module dependencies engine, so it not only invokes the `mod_init()` callback in a orderly fashion, but also `mod_destroy()`? This way, since `dialog` depends on `tm` (see [here](https://github.com/OpenSIPS/opensips/blob/master/modules/dialog/dialog.c#L474)), `tm` will both _initialize_ and _destroy_ first, thus fixing the issue without introducing any overhead at runtime.

This sounds much better, sure!

liviuchircu:
@lemenkov I just pushed all required code on `master` branch.  However, I'd give it a bit of time before backporting it.  May I ask if you are using some custom code in order to generate this backtrace?  I'm asking because @razvancrainea already pushed a similar shutdown ordering quick-fix in 2600e00217, so I don't see how the below backtrace can even happen on nowadays' code:

```
.0 0x00007fba1e0edceb in _unref_dlg (dlg=0x7fba24689470, cnt=1) at dlg_hash.c:1007
.1 0x00007fba216c4a3e in empty_tmcb_list (head=head@entry=0x7fba26ab5da0) at t_hooks.c:53
```

PR #2495, opened 2021-05-03
B2B: Contact user feature

nikbyte:


modules/b2b_logic/logic.c
```diff
@@ -4448,6 +4492,20 @@ int b2bl_bridge_msg(struct sip_msg* msg, str* key, int entity_no)
...
+		goto error;
+	}
+
+	local_contact = tuple->local_contact;
+	if (get_local_contact(msg->rcv.bind_address, &ct_uri.user, &local_contact) < 0)
+	{
+		LM_ERR("Failed to get received address\n");
+		local_contact = tuple->local_contact;
+	}
+
```
rvlad-patrascu, line 4508: The `local_contact` variable is never used after you build the Contact here, so this code is unnecessary.

modules/b2b_logic/records.c
```diff
@@ -117,10 +118,13 @@ b2bl_tuple_t* b2bl_insert_new(struct sip_msg* msg, unsigned int hash_index,
...
 	str extra_headers={0, 0};
 	str local_contact= server_address;
+	struct sip_uri *ct_uri = NULL;
+	struct to_body *pto;
 
 	if(msg)
 	{
-		if (get_local_contact(msg->rcv.bind_address, NULL, &local_contact) < 0)
+		if (contact_user && msg->to) ct_uri = parse_to_uri(msg);
+		if (get_local_contact(msg->rcv.bind_address, &ct_uri->user, &local_contact) < 0)
```
rvlad-patrascu, line 127: This would crash if the contact user feature is not used and the `ct_uri` pointer is NULL.

rvlad-patrascu (changes requested):
Hi @nikbyte ,

Overall, the patch looks OK besides the 2 issues I've pointed out. Can you push the fixes so I can merge this?

modules/b2b_logic/logic.c
nikbyte: Thank you for reviewing. It's my mistake when I ported from few patches, I fixed this, it's used by real in code later.

rvlad-patrascu (approved):

PR #2497, opened 2021-05-03
B2B: fix hardcoded 60 seconds tuple timeout, allow 0 as tuple timeout…

nikbyte:
… which means maximum duration


rvlad-patrascu (changes requested):
Hi @nikbyte ,

Can you also update the documentation about the changes here ? Firstly, a mention of the meaning of the new `0` timeout value would be helpful. Also, the docs currently explicitly state that the `t` flag in `b2b_init_request()` controls the timeout for the _topology hiding scenario_. After this commit this would no longer be the case (`t` flag will work for any scenario)

nikbyte:
@rvlad-patrascu Thank you for reviewing. I updated documentation.

rvlad-patrascu (approved):

PR #2498, opened 2021-05-03
Allow to pass To: header in b2b_init_request() as we do in create_top…

nikbyte:
…_hiding_entities().

RURI is not always a good candidate for To: header.

### DIFF
--- a/modules/b2b_logic/logic.c
+++ b/modules/b2b_logic/logic.c
@@ -3304,9 +3304,6 @@ str* b2b_process_scenario_init(struct sip_msg* msg, b2bl_cback_f cbf,
 	pkg_free(server_id);
 	tuple->servers[0]->type = B2B_SERVER;
 
-	pkg_free(to_uri.s);
-	to_uri.s = 0;
-
 	new_entity = NULL;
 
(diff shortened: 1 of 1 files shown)


modules/b2b_logic/logic.c
```diff
@@ -3304,9 +3304,6 @@ str* b2b_process_scenario_init(struct sip_msg* msg, b2bl_cback_f cbf,
 	pkg_free(server_id);
 	tuple->servers[0]->type = B2B_SERVER;
 
-	pkg_free(to_uri.s);
-	to_uri.s = 0;
```
rvlad-patrascu, line 3308: Removing this would introduce a memory leak. Can you move this code after `b2b_api.client_new()` ?

rvlad-patrascu (changes requested):

modules/b2b_logic/logic.c
nikbyte: @rvlad-patrascu thank you for reviewing. I fixed it as you recommended.

rvlad-patrascu (approved):

Issue #2425, opened 2021-02-28
b2b_logic can't handle more than 300-400 CPS and has memory corruption


nikbyte:
Hi team,

I found that if start load tests with common calls, starting from 300-400 CPS b2b_logic starts to have memory corruptions.
It's easy reproduceable in two cases:

1. Enable write-through db writing (db_mode=1) and you'll get corrupted dlg structures on b2bl_db_update. If enable db_mode=2 even with 1 second interval this doesn't reproduce.

2. From time to time call b2b_cback() in b2b_entities module crashes opensips because b2b_cback pointer points to some released memory.

Usually it occurs on BYE or final ACK when there is a large amount of requests and especially when there are retransmits.

rvlad-patrascu:
Hi @nikbyte ,

Did you test this with the "top_hiding" scenario ? If not, can you also post the script logic that triggers this?

nikbyte:
It seems, I found a reason. I'll post pool request soon.

nikbyte:
I'm testing with non-top hiding, but scenario is pretty simple: create server, create client, init b2b.

nikbyte:
@rvlad-patrascu this helps a lot: https://github.com/OpenSIPS/opensips/pull/2426
Also I have patch which helps, but I'm not sure it's proper:
```
diff --git a/modules/b2b_entities/dlg.c b/modules/b2b_entities/dlg.c
index f4f1370e8..d2281a6a5 100644
--- a/modules/b2b_entities/dlg.c
+++ b/modules/b2b_entities/dlg.c
@@ -1070,7 +1070,7 @@ logic_notify:

 				if(dlg->uas_tran && dlg->uas_tran!=T_UNDEFINED)
 				{
-					if(dlg->uas_tran->uas.request)
+					if(method_value != METHOD_BYE && dlg->uas_tran->uas.request)
 					/* there is another transaction for which no reply
 					 * was sent out */
 					{
```

Without this patch there is dlg->uas_tran exists, but dlg->uas_tran.uas points to released memory.
Maybe you know how to fix this properly.

With these two patches it's not crashing for me.

bogdan-iancu:
@nick, could you explain a bit your fix here, please ?

nikbyte:
@bogdan-iancu I'm not sure that this fix is proper, it may be more workaround. But.

Initial condition is if(dlg->uas_tran->uas.request) .
[...]

PR #2500, opened 2021-05-03
B2B: generation of 491 request pending on BYE is not the best idea

nikbyte:


rvlad-patrascu:
Hi @nikbyte 

If we're not generating a 491 reply I think that the BYE should end the dialog and not be simply discarded(as it would happen if merging this).

nikbyte:
Usually this occurred when we already answered BYE (and finished dialog) and these are retransmits. When dialog is already not found.

rvlad-patrascu:
So I guess that this change was only to fix the crashes reported in https://github.com/OpenSIPS/opensips/issues/2425, correct? In that case, I think that the better approach is to find a proper fix for that. Because as you were saying, this seems more of an work-around.

nikbyte:
Yes. It's for case #2425. If you see the root reason, of course proper fix appreciated.

rvlad-patrascu:
Closing this at the moment since it does not appear to be a proper fix for https://github.com/OpenSIPS/opensips/issues/2425 or the correct approach for handling a BYE when another request is pending.

PR #2503, opened 2021-05-03
B2B: thread safe NOTIFY generation

nikbyte:


rvlad-patrascu:
Hi @nikbyte ,

Could you explain a bit the problem that this is trying to solve? How did it manifest? And how are these changes going to fix the issue?

nikbyte:
Hi Vlad. Without this patch there is a situation when body buffer is corrupted and instead of `SIP/2.0 100 Trying` there will be some garbage. It was initially wrong idea to change pointer of static body.s variable because somewhere later in code this buffer can be overwritten. That's why I make separate body buffer and copy there what is required for this notify. It 100% solved issue with garbage instead of `SIP/2.0 100 Trying`.

rvlad-patrascu:
A more straightforward fix for this issue in  https://github.com/OpenSIPS/opensips/commit/947c7e3c295cffde48ddb4c35d928ae7339722dd.

PR #2504, opened 2021-05-03
Dialog module: skip bogus dialogs when loading from DB

nikbyte:


liviuchircu (approved):

liviuchircu:
Thanks for the patch, @nikbyte!  Although I am not sure how you can end up in a state where **some** DB dialogs are good and **some** have bad hashes (to me, it looks like either all dialogs are OK or all are broken)... it can only help to skip the bad ones and continue loading the good ones.  The administrator will eventually clean up the table if they don't want to keep running into that annoying error on startup or reload.

PR #2505, opened 2021-05-03
Drouoting: Use db_partitions_url if db_url field is empty in dr_parti…

nikbyte:
…tions table. Useful when you need different partitions in the same DB


bogdan-iancu:
@nikbyte , in `dr_partitions`, the `db_url` cannot be NULL (according to the DB definition) - how do you want to use this, with empty values ?

nikbyte:
Maybe I need to fix db structure. It works for me because my database is view and I use partitions from the same database, but want to have common url which is defined in opensips cfg. If there is null in database, it uses default url. Useful.

PR #2506, opened 2021-05-03
s.time transformation to print formatted time from unix time. example…

nikbyte:
…: {s.time,%Y-%m-%d %H:%M:%S}


bogdan-iancu:
Hi @nikbyte , if the input of this transformation is a unix timestamp, why it is a `s.` transformation ? definitely the input is not a string.

nikbyte:
@bogdan-iancu I opened to any offers. Please give me your advice how it must be and I change it.

bogdan-iancu:
Indeed, that is a good question. The integer, as input, in like 10 years or so (??), may not be enough for the timestamp. What about having a {time.} class of transformation and this one should be the first {time.print} or {time.format} supporting the input as both interger or string. What do you think ?

nikbyte:
still actual

bogdan-iancu:
@nikbyte , please see my questions above

nikbyte:
@bogdan-iancu I saw, it's reopened for me to redo this patch. I reassigned to myself and then reassign back to you when it's ready.

bogdan-iancu:
I mean this [comment](https://github.com/OpenSIPS/opensips/pull/2506#issuecomment-847073194).

nikbyte:
@bogdan-iancu I completely agree with you and accepted it as "to do". 🙂

Issue #2333, opened 2020-12-08
[BUG] Topology hiding without dialog-module destroys upstream RR


wdoekes:
**OpenSIPS version you are running**
```
version: opensips 3.2.0-dev (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 2f3ee99c6
main.c compiled on 15:00:45 Nov 27 2020 with gcc 9
```

**Describe the bug**

When using two or more proxies, dialog-less topology_hiding removes Route headers in the upstream direction.

In short:
* INVITE: alice -> proxy1+rr -> proxy2+topoh -> bob
* BYE: bob -> proxy2 -> alice (directly, bypassing proxy1)

Because:
* INVITE
  * alice
  * proxy1 adds Record-Route
  * proxy2 replaces Record-Route with fancy Contact with thinfo
  * bob gets request without RR (and no notion of proxies)
* 200
  * bob
  * proxy2 replaces Bob-contact with proxy2-contact, reads fancy RURI, replaces with Record-Route
  * proxy1
  * alice gets response with RR
* ACK
  * alice adds Route
  * proxy1 reads/strips Route
  * proxy2 reads fancy RURI and replaces RURI and Contact
  * bob gets request (and no notion of proxies)

All is good so far.

But now, the BYE from Bob:

* BYE
  * bob
  * proxy2 reads fancy RURI, replaces RURI and Contact (loose_routed to alice)
  * alice gets request from proxy2 directly (_and_ misses out on possible Route flags)

Here, Alice gets the BYE from proxy2, which should've gone to proxy1 first.

**To Reproduce**

[topoh_drop_route.zip](https://github.com/OpenSIPS/opensips/files/5658694/topoh_drop_route.zip)

In the attached ZIP file, there is:

* backend.cfg (proxy1 config)
* sbc.cfg (proxy2 config)
* sbc-w-topoh.cfg (proxy2 config with topology hiding)
* itsp-uac.xml (sipp scenario)
* phone-uas.xml (sipp scenario)
[...]

PR #2507, opened 2021-05-03
topology_hiding: Allow BYE with Route: header (ignore the header, not…

nikbyte:
… reject BYE)

### DIFF
--- a/modules/topology_hiding/topo_hiding_logic.c
+++ b/modules/topology_hiding/topo_hiding_logic.c
@@ -143,7 +143,7 @@ int topology_hiding_match(struct sip_msg *msg)
 
 	r_uri = &msg->parsed_uri;
 
-	if (check_self(&r_uri->host,r_uri->port_no ? r_uri->port_no : SIP_PORT, 0) == 1 && msg->route == NULL) {
+	if (check_self(&r_uri->host,r_uri->port_no ? r_uri->port_no : SIP_PORT, 0) == 1 && (msg->route == NULL || (msg->REQ_METHOD==METHOD_BYE))) {
 		/* Seems we are in the topo hiding case :
 		 * we are in the R-URI and there are no other route headers */
 		for (i=0;i<r_uri->u_params_no;i++)


wdoekes:
Hi @nikbyte: can you explain the use case? What problem does this solve?

nikbyte:
This solves the problem when broken UAC sends me BYE with Route header and opensips ignores it.
After patch opensips ignores Route header and processes BYE as usual.

wdoekes:
Okay @nikbyte: _that much_ I understood already :)

It might not be my place, as this is not my project. And it's really nice that you're filing all the PRs (thanks for that!). But, if you filed a PR like this on my project, I'd be concerned with the following questions:

- does the change have a negative impact on behaviour (strictness) for others? (this makes certain checks less strict: sloppiness is seldom a good thing)

- if it's a hack: is it documented what the hack is for? (I see no comments about why this is allowed in the source, for which kind of broken UAC)

- why does the UAC behave like this? which model UAC is it? does it use preloaded Routes? where did it get the Route from?

- is there a possibility to fix this in a script? you agree that the problem is with the UAC: then changing the opensips core/modules should be the last resort. first resorts would be: (1) replacing the UAC, (2) seeing if this can be fixed with additional scripting (re-route through itself after removing the Route header?)

The same question applies to some of your other PRs. Again, awesome that you've taken the trouble of filing them in the first place. But if it's not immediately obvious to others (1) why the change is beneficial, that (2) _this_ is the right place for the fix and (3) that the change doesn't adversely affect others by changing existing behaviour, then it's harder to evaluate and merge these commits.

Cheers :beers: 
Walter

wdoekes:
By the way, remotely related is #2333 -- there we have a BYE with too few Routes instead, but that's not the UAC's fault. It does increase my anxiety about adding this workaround though.

bogdan-iancu:
@nikbyte , @wdoekes , I will have to agree here with @wdoekes on the "politics" side, both technical and over management. The PR must describe as describe in details the problem they address and how they fix it. Otherwise, when we evaluate the PR we need to do revers-engineering and guess on "what the author of the PR was trying to do here".... :cry: 
So the tech details will guarantee that your PR will be understood and considered , otherwise it will simply stay there, as the effort of looking into is too high.

Also, I think it is not a good approach to cripple the OpenSIPS SIP stack, just to make whatever broken UAC happy.  I understand that OpenSIPS has to be flexible, but not with this cost. Even more, the change may trigger some false positive matching in the b2b module, so there are some negative side effects here.

IMO, the correct approach should be, for sequential request with `Route` headers, to check if there is only one hdr instance if the URI points to OpenSIPS itelself - that is an almost legitimate SIP behavior (forcing an Outbound Proxy with a Route hdr).

So, @nikbyte , is the your BYE holding a single Route pointing to OpenSIPS ?

nikbyte:
Hi Bogdan. The root issue that in topology_hiding module we have no any chance to process BYE from script in case if route header is there. Maybe we need to process it more smart than just remove, but root issue – a lack of control on this BYE request.

nikbyte:
?

PR #2508, opened 2021-05-03
rtpengine: support for zero-address flag

nikbyte:


modules/rtpengine/rtpengine.c
```diff
@@ -1857,6 +1857,13 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg,
 					break;
 				continue;
 
+			case 20:
+				if (str_eq(&key, "replace-zero-address")) {
+					if (!ng_flags->replace)
+						LM_DBG("%.*s not supported for %d op\n", key.len, key.s, *op);
+					else
+						BCHECK(bencode_list_add_string(ng_flags->replace, "zero address"));
+
```
wdoekes, line 1866: Sure? No break/continue?

razvancrainea: Both break and continue should be there - nevertheless, from functionality pov this is fine, but not optimal. Will push a fix, thanks for spotting it!

Issue #2512, opened 2021-05-04
[BUG] opensips 3.1 TCP socket search based on proto/ip/port doesn't respect source socket


kertor:
**OpenSIPS version you are running**

```
version: opensips 3.1.1 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-FUTEX-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 9d37433f5
main.c compiled on  with gcc 9
```

**Describe the bug**

I have two opensips listeners:
- `tls:1.1.1.1:5061`
- `tls:1.1.1.2:5061`

I have one destination gateway:
- `tls:3.3.3.3:5061`

Drouting send OPTIONS to `tls:3.3.3.3:5061` from both listeners.
When opensips starts and create TLS connection, all working correct.
But when opensips handle new OPTIONS send task, he send both OPTIONS from same `tls:1.1.1.1:5061`, because find first TCP transaction on DST attribute.
In opensips src code `net/net_tcp.c` i see the same: search TCP connection only on DST attribute.
So, gateway `tls:3.3.3.3:5061` receive both OPTIONS from same `tls:1.1.1.1:5061`, but in SIP message present correct `tls:1.1.1.2:5061` values in SIP headers.

**To Reproduce**

1) Create two TCP listeners.
2) Add in drouting two gateways with same DST but different socket_out values and enable OPTIONS pinging.
3) You will see on gateway that you receive OPTIONS from same real IP, but in SIP will present different IPs.

**Expected behavior**

Opensips must search TCP transactions on DST and on socket_out attributes and use it.

**Relevant System Logs**

**OS/environment information**
 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

**Additional context**

nikbyte:
Fixed in https://github.com/OpenSIPS/opensips/pull/2513.

PR #2513, opened 2021-05-04
Fix tcp/tls socket reusage for multihome opensips

nikbyte:
The patch for https://github.com/OpenSIPS/opensips/issues/2512

### DIFF
--- a/net/net_tcp.c
+++ b/net/net_tcp.c
@@ -427,7 +427,8 @@ int tcp_get_correlation_id( unsigned int id, unsigned long long *cid)
 /*! \brief _tcpconn_find with locks and acquire fd */
 int tcp_conn_get(unsigned int id, struct ip_addr* ip, int port,
 		enum sip_protos proto, void *proto_extra_id,
-		struct tcp_connection** conn, int* conn_fd)
+		struct tcp_connection** conn, int* conn_fd,
+		struct socket_info* send_sock)
 {
 	struct tcp_connection* c;
 	struct tcp_connection* tmp;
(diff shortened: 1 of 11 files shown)


nikbyte:
With this patch when opensips is matching existing tcp connection, it compares source socket and reuse only in case of source sockets are the same.

net/net_tcp.c
```diff
@@ -852,6 +857,8 @@ struct tcp_connection* tcp_conn_create(int sock, union sockaddr_union* su,
 	if (!send2main)
 		return c;
 
+	c->send_sock = si;
```
razvancrainea, line 860: I don't think there's any need of a new `c->send_sock` in the connection structure, as you already have that information in the `c->rcv.bind_address` field.

net/net_tcp.c
```diff
@@ -461,10 +462,14 @@ int tcp_conn_get(unsigned int id, struct ip_addr* ip, int port,
...
 					a->parent->rcv.src_port);
 				print_ip("ip=",&a->parent->rcv.src_ip,"\n");
+				if (send_sock != NULL && a->parent->send_sock != NULL)
+					print_ip("requered send_sock ip=",&send_sock->address,"\n");
+					print_ip("found send_sock ip=",&a->parent->send_sock->address,"\n");
 #endif
 				c = a->parent;
 				if (c->state != S_CONN_BAD &&
 				    c->flags&F_CONN_INIT &&
+				    send_sock == a->parent->send_sock &&
```
razvancrainea, line 472: If a socket is not forced (thus `send_sock` is NULL), then we might want to use any connection towards that destination, don't we?

razvancrainea (changes requested):

nikbyte:
@razvancrainea Thank you very much for your useful comments. I fixed regarding your notes.

net/net_tcp.c
nikbyte: Fixed

net/net_tcp.c
nikbyte: Fixed

razvancrainea (approved):
This looks all right now, merging it.

net/net_tcp.c
```diff
@@ -461,10 +462,14 @@ int tcp_conn_get(unsigned int id, struct ip_addr* ip, int port,
 					a, a->parent, a->parent->id, a->port,
 					a->parent->rcv.src_port);
 				print_ip("ip=",&a->parent->rcv.src_ip,"\n");
+				if (send_sock != NULL && a->parent->send_sock != NULL)
+					print_ip("requered send_sock ip=",&send_sock->address,"\n");
```
wdoekes, line 466: Assuming that should be `required`?

razvancrainea: This has been fixed before merging, but replaced with `requested` instead of `required`.

PR #2518, opened 2021-05-11
Reduce db_is_neq_type complexity

wdoekes:
The code was correct, but overly verbose. Boil it down to the intention and rely on the compiler to work its magic.

(Outsmarting the compiler _may_ sometimes work, but one can wonder whether the reduced readability and room for future error is worth it.)

### DIFF
--- a/db/db_ut.c
+++ b/db/db_ut.c
@@ -418,48 +418,34 @@ int db_print_set(const db_con_t* _c, char* _b, const int _l, const db_key_t* _k,
 int db_is_neq_type(db_type_t _t0, db_type_t _t1)
 {
 	// LM_DBG("t0=%d t1=%d!\n", _t0, _t1);
-	if(_t0 == _t1)
-		return 0;
-
-	switch(_t1) {
-	case DB_INT:
-		if(_t0==DB_BIGINT || _t0==DB_DATETIME || _t0==DB_BITMAP)
-			return 0;
-		break;
-
-	case DB_STRING:
-		if(_t0==DB_STR || _t0 == DB_BLOB)
-			return 0;
-		break;
-
-	case DB_STR:
-		if(_t0==DB_STRING || _t0==DB_BLOB)
-			return 0;
-		break;
-
+	switch(_t0) {
 	case DB_BIGINT:
-		if(_t0==DB_INT || _t0==DB_DATETIME || _t0==DB_BITMAP)
-			return 0;
-		break;
-
-	case DB_BLOB:
-		if(_t0==DB_STR || _t0==DB_STRING)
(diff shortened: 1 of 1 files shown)


wdoekes:
(This one might be for you @razvancrainea :wink: )

db/db_ut.c
```diff
@@ -416,46 +416,35 @@ int db_print_set(const db_con_t* _c, char* _b, const int _l, const db_key_t* _k,
...
+		case DB_INT:
 			return 0;
-		break;
-
-	case DB_BITMAP:
-		if(_t0==DB_INT || _t0==DB_BIGINT || _t0==DB_DATETIME)
+		default:
+			break;
+		}
+	case DB_BLOB:
```
liviuchircu, line 433: Optimization quirk: shouldn't there be a `break` after this curly brace?  E.g., on a `DB_INT` vs. `DB_BLOB` comparison, `_t0` will continue to fall through the BLOB, STR, STRING before finally exiting, rather than doing it immediately.

db/db_ut.c
```diff
@@ -416,46 +416,35 @@ int db_print_set(const db_con_t* _c, char* _b, const int _l, const db_key_t* _k,
...
-
+		default:
+			break;
+		}
 	default:
+		if(_t0 == _t1) {
+			return 0;
+		}
 		break;
 	}
```
liviuchircu, line 450: Personal preference here, not sure if you agree.  Wouldn't this code block read better if it were moved below into a single line, with the end of the function simply saying: `return _t0 != _t1;`?

liviuchircu (commented):
Hey, @wdoekes!  This function could definitely use some optimizing on the maintenance side :)  I left some suggestions, let me know your comments -- thank you!

db/db_ut.c
wdoekes: Ah. Good call. I mistakenly thought I was breaking from _t0 in default, but I'm breaking from _t1.

Fixing...
```c
...
                case DB_INT:
                        return 0;
                default:
                        break;
                }       
        case DB_BLOB:
```
->
```c
...
                case DB_INT:
                        return 0;
                default: // still needs this, to avoid complaints from compiler
                        break;
                }
                break;
        case DB_BLOB:
```
->
```c
...
                case DB_INT:
                        return 0;
                default:
                        return 1;
                }
        case DB_BLOB:
```
Chose the last one then.

db/db_ut.c
wdoekes: Agreed.

liviuchircu:
Nice work, @wdoekes, cheers!

PR #2519, opened 2021-05-11
Fix database purging of stale activewatchers

wdoekes:
This changeset contains two commits.

(1) To enable IS NULL support (it's ugly).

(2) Fix so activewatcher.sharing_tag IS NULL records get pruned, which was broken since 929ab4d4 (or 6022968fa for 2.4.x).

Without this changeset, people using clustering but no sharing tags will get stale records in the database.

See the commit messages for details.


wdoekes:
(This one might be for you @bogdan-iancu :wink: . See also PR #2520.)

wdoekes:
Rebased so the conflicts with 87f24165 are gone :tada:

hafkensite:
Any reason not to merge this request? 

I'm asking because our cluster is experiencing this issue, resulting in a lot of database stress and maintenance to prevent further issues.

PR #2520, opened 2021-05-11
presence: Fix database purge of activewatchers (clustering, fallback2db=0)

wdoekes:
When clustering sharing_tags were added to presence, they were added to the fallback2db "on" case only:

There are a couple of dimensions with differing behaviours:
```
                +------------+------------+
                | fallback2- | fallback2- |
                | -db = on   | -db = off  |
  +-clustering:-+------------+------------+
  | - no        | OK         | OK         |
  | - tagless   | PR-2519    | PR-2519    |
  | - active    | OK         | this       |
  +-------------+------------+------------+
```
The non-OK behaviour above refers to the activewatcher table getting filled up with stale/expired items.

fallback2db on or off:
  ```
  modparam("presence", "fallback2db", 0)  # or 1=on
  ```

The no-clustering case:
  ```
  handle_subscribe();
  ```

The tagless case:
  ```
  modparam("presence", "cluster_id", 1)
  modparam("clusterer", "my_node_id", 2)

  handle_subscribe();
  ```

The active case:
  ```
  modparam("presence", "cluster_id", 1)
  modparam("clusterer", "my_node_id", 2)
  modparam("clusterer", "sharing_tag", "node2/1=active")

  handle_subscribe("0", "node2");
  ```

Where PR #2519 fixes the tagless case, this PR fixes the `fallback2db=0` case by writing the sharing_tag to the database so the records can get found and cleaned up.

(Sidenote: subscriptions which ended with a timeout or 481 *would* get cleaned up. This makes sense in all cases: if they have an error before their expiry, it makes sense to purge them from the DB immediately. And if the periodic cleanup had cleaned those records already, it would not be an issue.)

### DIFF
--- a/modules/presence/subscribe.c
+++ b/modules/presence/subscribe.c
@@ -1363,9 +1363,10 @@ void update_db_subs(db_con_t *db,db_func_t *dbf, shtable_t hash_table,
 	db_op_t update_ops[2];
 	subs_t* del_s;
 	int pres_uri_col, to_user_col, to_domain_col, from_user_col, from_domain_col,
-		callid_col, totag_col, fromtag_col, event_col,status_col, event_id_col,
+		callid_col, totag_col, fromtag_col, event_col, status_col, event_id_col,
 		local_cseq_col, remote_cseq_col, expires_col, record_route_col,
-		contact_col, local_contact_col, version_col,socket_info_col,reason_col;
+		contact_col, local_contact_col, version_col, socket_info_col,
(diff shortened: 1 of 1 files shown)


wdoekes:
(This one might be for you @bogdan-iancu :wink: . See also PR #2519. Both affect both the 2.4.x and 3.1.x branches.)

wdoekes:
Rebased, just in case

PR #2558, opened 2021-06-24
[uac_registrant] Match reply Contact with binding_URI by parameter.

### LINKED ISSUE #2557: [FEATURE] uac_registrant match URI to Contact header partially

If we have a lot of external registrations, let's say 100000, we can distribute them through several, for example 10 servers. Put accounts to the DB, connect all OpenSIPS servers to this DB, create cluster of servers.Create for example 10 different tags for cluster, so each server will process only 10% of total amount of registrations. And if one server dies, just move its' tag to another server. It will increase load twice, but everything will be reliable.
Very simple scheme. Just every server substitutes its' IP-address instead of the hostname in "binding_URI" when sending outgoing REGISTER:
```
local_route {
	if (is_method("REGISTER") && $(ct.fields(uri){uri.host}) == "local.opensips") {
		$var(subst) = "/local.opensips/$socket_out(ip):$socket_out(port)/g";
		$var(new_contact) = $(ct{re.subst,$var(subst)});
		remove_hf("Contact");
		append_hf("Contact: $var(new_contact)\r\n");
	}
}
```
Everything works, except one problem. Expiration time for registration, provided in "200 OK" from external system is not applied to OpenSIPS. Because module uac_registrant compares URI from Contact header from "200 OK" to the full value of "binding_URI". They are different, as we are changing Contact in outgoing REGISTER.
It would be nice to have parameter for matching contact URI, to choose what parts of URI must match. And just ignore the rest.

Parameter to uac_registrant module to choose what parts of URI use to match.

**Implementation**
I'll send PR

Store in `reg_record->contact` not just raw value of `binding_URI`, but dynamic value with host depending on sending socket.

vasilevalex:
Added new integer module parameter 'match_contact' to choose, what
parts of URI from Contact header should match binding_URI when
OpenSIPS chooses expiration time for registration.

**Summary**
Solving #2557 

**Details**
It simplifies using uac_registrant module in cluster environment, when there is one DB per cluster, and each host uses its' own IP-address for SIP.

**Solution**
For example, setting `modparam("uac_registrant", "match_contact", 3)` allows matching only by user, ignoring host. As `binding_URI` contains specific host, and every server sends REGISTER with its' own IP-address.

**Compatibility**
Default behaviour is not changed.

**Closing issues**
closes #2557


vasilevalex:
up

vasilevalex:
up

vasilevalex:
up

vasilevalex:
up

liviuchircu (changes requested):
I like the idea behind this PR, as it brings a lot more flexibility to `uac_registrant` when trying to achieve interoperability with various UAS and their reply Contacts.

Still, I don't like that we are encouraging `opensips.cfg` code which uses these hardcore Contact matching algorithm bitmasks such as `3`, `11`, `19`, `67`, `273`, `1374`, etc., which increase the script reviewing difficulty, as well as forcing script developers to constantly consult the `parser/parse_uri.h` C program file, which I think should be avoided at all costs.

Suggestion to fix the above:  add a "string CSV options" converter function in `parser/parse_uri.h` to reverse human-readable input such as `"user|passwd|userparam|transport"` into the same `enum uri_match_flags` as you are using now.  The function will be called once at startup (during fixup phase) anyway, so it will come with zero runtime overhead, but with a big QoL improvement.

Finally, let's also add some documentation in the `modules/uac_registrant/doc/uac_registrant_admin.xml` file before we can merge this new feature (modparam).

vasilevalex:
@liviuchircu can you please review, when you have time?

jes:
I just had a look at this and noticed a couple of potential issues:

1. the "maybe we're lucky and straight-forward comparison succeeds" compares with `strncasecmp` which is not correct as detailed in https://github.com/OpenSIPS/opensips/pull/3263#pullrequestreview-1760759447
2. it allocates fixed-size buffers on the stack for `char unescaped_a[UNESCAPED_BUF_LEN], unescaped_b[UNESCAPED_BUF_LEN]`, but then writes into them without any bounds-checking

bogdan-iancu:
@vasilevalex , could you detail/describe the scenario where the host part should be ignored (in a clustered setup) ?

vasilevalex:
Hi, @jes @bogdan-iancu. I described why we need in some cases to compare Contact header from 200 Ok only with partial match in #2557 .
I can add here. If database with credentials is used by several OpenSIPS server with clusterer setup. All the servers use the same "binding_URI". But all the servers have different IP addresses. So we distribute registrations through cluster. In 200 Ok we will receive different Contact URI, depending of the OpenSIPS server which sent the REGISTER. It means that we must also always change binding_URI in database to have Contact matching. And in case of failover of one of the servers we must not just move shared tag to another server, but also fix all the records in DB changing IP-address in the binding_URI. That's the case, when we want to compare URI from Contact header only partially, excluding host part.
P.S. @jes I'll try to check the issues you mentioned.

PR #2591, opened 2021-08-08
fix: From Nick Altmann: Fix wrong variable usege when escape STR type…

NormB:
… in db_postgres

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**


razvancrainea:
This was initially solved by @nikbyte in #2423, however I find this version more easy to merge, that's why I chose it.

PR #2592, opened 2021-08-08
Tls postgres

NormB:
**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**


rvlad-patrascu (commented):
Hi @NormB ,

Overall this looks OK and I am able to connect to postgresql via TLS, but there are still a few issues here.

First of all, using the openssl library results in crashes, similar with problems we've had with db_mysql but the openssl tweaks we've done in that case don't seem to fix the issues here. I have tried using the  `PQinitOpenSSL()` function, which should prevent the initialization of openssl by libpq, but to no success. Maybe you could look into these crashes a bit further. If we cannot avoid them, we should at least not allow starting opensips with the `tls_openssl` module and include clear warnings about this in the documentation. Using the `tls_wolfssl` module or no TLS library at all, configured with `modparam("tls_mgm", "tls_library", "none")`, work fine on the other hand.

menuconfig/configs/opensips_residential_def.m4
```diff
@@ -1,6 +1,6 @@
 divert(-1)
-define(`ENABLE_TCP', `no') # OpenSIPS will listen on TCP for SIP requests
-define(`ENABLE_TLS', `no') # OpenSIPS will listen on TLS for SIP requests
+define(`ENABLE_TCP', `yes') # OpenSIPS will listen on TCP for SIP requests
+define(`ENABLE_TLS', `yes') # OpenSIPS will listen on TLS for SIP requests
```
rvlad-patrascu, line 3: This is not related to postgresql and I don't think we should enable TLS/TCP by default in the residential script anyway.

modules/db_postgres/doc/db_postgres_admin.xml
```diff
@@ -112,6 +112,56 @@ modparam("db_postgres", "max_db_queries", 2)
...
 modparam("db_postgres", "timeout", 2)
 ...
+</programlisting>
+		</example>
+	</section>
+
+        <section id="param_use_tls" xreflabel="use_tls">
+		<title><varname>use_tls</varname> (integer)</title>
+		<para>
+		Warning: <emphasis>wolfSSL</emphasis> is the recommended TLS/SSL Library
```
rvlad-patrascu, line 122: This could be expanded a bit, warning about the crashes with openssl and also mentioning that it is possible to use no TLS library for the `tls_mgm` module.

modules/db_postgres/ssl_tweaks.c
```diff
@@ -0,0 +1,50 @@
...
+		real_SSL_CTX_set_default_verify_paths = dlsym(RTLD_NEXT,
+			"SSL_CTX_set_default_verify_paths");
+		if (!real_SSL_CTX_set_default_verify_paths)
+			return 0;
+
+		init = 1;
+
+		return real_SSL_CTX_set_default_verify_paths(ctx);
+	}
+}
```
rvlad-patrascu, line 50: This is not useful as it doesn't prevent the crashes.

modules/db_postgres/pg_con.h
```diff
@@ -56,6 +57,10 @@ struct pg_con {
 	char**  row;		/* Actual row in the result */
 	time_t timestamp;	/* Timestamp of last query */
 
+	unsigned int init;       /* If the connection is initialized */
+	unsigned int disconnected; /* disconnection detected */
```
rvlad-patrascu, line 61: It seems that those fields are unused.

modules/db_postgres/pg_con.c
```diff
@@ -19,87 +19,169 @@
...
+
+	if (strncmp(id->parameters, DB_TLS_DOMAIN_PARAM_EQ, DB_TLS_DOMAIN_PARAM_EQ_LEN)) {
+		LM_ERR("Invalid URL (tls_domain) parameter: %s\n", id->parameters);
+		return NULL;
+	}
+
+	char *pos = strchr(id->parameters, '&');
+
+	dom.s = id->parameters + DB_TLS_DOMAIN_PARAM_EQ_LEN;
+	dom.len = pos - dom.s;
```
rvlad-patrascu, line 56: This will not work if the `tls_domain` is the only DB URL parameter.

modules/db_postgres/pg_con.c
```diff
@@ -19,87 +19,169 @@
...
-			ZSW(dbname));
+		/* Set the initial dbname parameter */
+		if (tls_domain_name->len && lenp > DB_TLS_DOMAIN_PARAM_EQ_LEN+tls_domain_name->len) {
+			/* For example: tls_domain=[dom]&hostaddr=1.2.3.4&application_name=opensips */
+			/* DB_TLS_DOMAIN_PARAM_EQ_LEN = length of 'tls_donain=' (11) */
+			/* tls_domain_name->len = length of [dom] */
+			/* +1 is because the first parameter starts with a '&' and there is no need to memcpy() it */
+			/* Result is 'hostaddr=1.2.3.4 application_name=opensips' */
+			/* https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS */
+			memcpy(dbname, id->parameters+DB_TLS_DOMAIN_PARAM_EQ_LEN+tls_domain_name->len+1, lenp - (DB_TLS_DOMAIN_PARAM_EQ_LEN+tls_domain_name->len)-1);
```
rvlad-patrascu, line 165: It seems that `tls_domain` should always be the first parameter. It would be nice to be able to use it in any position. If not, we should document this requirement.

rvlad-patrascu (changes requested):

menuconfig/configs/opensips_residential_def.m4
```diff
@@ -1,18 +0,0 @@
...
-define(`USE_MULTIDOMAIN', `no') # OpenSIPS will handle multiple domains for subscribers
-define(`USE_NAT', `no') # OpenSIPS will try to cope with NAT by fixing SIP msgs and engaging RTPProxy
-define(`USE_PRESENCE', `no') # OpenSIPS will act as a Presence server
-define(`USE_DIALPLAN', `no') # OpenSIPS will use dialplan for transformation of local numbers
-define(`VM_DIVERSION', `no') # OpenSIPS will redirect to VM calls not reaching the subscribers
-define(`HAVE_INBOUND_PSTN', `no') # OpenSIPS will accept calls from PSTN gateways (with static IP authentication)
-define(`HAVE_OUTBOUND_PSTN', `no') # OpenSIPS will send numerical dials to PSTN gateways (with static IP definition)
-define(`USE_DR_PSTN', `no') # OpenSIPS will use Dynamic Routing Support for PSTN interconnection
-define(`USE_HTTP_MANAGEMENT_INTERFACE', `no') # OpenSIPS will provide a WEB Management Interface on port 8888
-divert
```
rvlad-patrascu, line 18: These lines were not affected by the update so they should not be removed.

rvlad-patrascu:
Hi @NormB ,

Have you had a chance to look at the remaining changes requested here?

NormB:
Deleted branch - Will create a new branch with the latest code.

PR #2600, opened 2021-08-16
rtpengine: allow weight, setid params when calling `rtpengine_enable`

john08burke:
**Summary**
This PR adds the ability to (optionally) provide `weight` and `setid` params when calling the MI `rtpengine_enable` method.  Similar to the rtpproxy module, the `setid` param acts as a filter when enabling/disabling nodes.  The `weight` param updates the node(s) load-balancing weight.  Since the new params are optional, the MI signature is backwards compatible.

**Justification**
I was looking for a way to dynamically set the load-balancing weights of a single set of rtpengine nodes at startup without having to maintain multiple tables/views via a SQL backend. The current balancing algo is quite simple, so I didn't want to mess with it. However, the combo of `rtpengine_show` and `rtpengine_enable` (after this PR) gives the ability to modify via script!

### DIFF
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -1360,17 +1360,26 @@ rtpengine_play_dtmf("0"); # send the 0 code upstream
 				<listitem><para>
 					<emphasis>enable</emphasis> - 1 - enable, 0 - disable the &rtp; proxy.
 				</para></listitem>
+        <listitem><para>
+					<emphasis>weight</emphasis> (optional)  the weight to be used when load balancing.
+				</para></listitem>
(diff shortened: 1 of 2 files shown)


john08burke:
Hey @razvancrainea, can we reopen this PR?  I missed the bot marking as stale. Thanks!

bogdan-iancu:
@john08burke , ok let's end up with one PR for adding the set id filter. For the weights part, we can skip this for now, if you do not have a need anymore - theoretically you can achieve this by changing the DB definition of the rtpengine socks and fire a reload.

john08burke:
Hey @bogdan-iancu, please see PR #2657 which has just the setid functionality!  Going to close this one for now.

PR #2603, opened 2021-08-18
rtpengine: use dialog support when possible

john08burke:
**Summary**
Currently, the `rtpengine` module uses a simple algo based upon Call-ID hash to determine which node to select from a set and does so each time rtpengine needs to be engaged.  The upside to this approach is that it is simple and stateless.  However, issues can arise when making changes to the rtpengine set(s) via `rtpengine_reload` command (ex: removing/adding a node to/from the set).  In such a case, any on-going dialogs are unlikely (although this depends upon weighting) to communicate with the correct node after the reload, which leads to the dreaded `Unknown call-id` response from rtpengine and consequently audio / teardown issues.

The relevant node selection algo:
https://github.com/OpenSIPS/opensips/blob/06474f16e2f5cdeebd81f91d35ff25503818a98b/modules/rtpengine/rtpengine.c#L2544-L2583

Just to demonstrate the scenario:
1. Set of 3 rtpengine nodes exists: node1, node2, node3.
  - All weights = 1.
  - All nodes enabled.
2. INVITE with SDP arrives and `hash(Call-ID) = 1000`
  - `sumcut = 1000 % 3 = 1`.
  - Selected node = node2.
3. We need to remove node3, so issue `mi rtpengine_reload`.
  - Set is now: node1, node2.
4. 200OK with SDP arrives.
  - `sumcut= 1000 % 2 = 0`.
  - Selected node = node1, even though node2 was used and is still available.

**Solution**
Use dialog API to store the selected node URL after SDP offer.  On subsequent SDP answer, and re-negotiations, lookup the node based upon the stored dialog value.  If the node was removed, or possibly disabled, we use the existing algo to decide which node to use.

**Compatibility**
This PR is backwards compatible, as if dialog support is disabled we simply skip the dialog storage/lookup steps and use the existing algo.

### DIFF
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -297,6 +297,7 @@ static int fixup_free_set_id(void ** param);
 static int set_rtpengine_set_f(struct sip_msg * msg, rtpe_set_link_t *set_param);
 static struct rtpe_set * select_rtpe_set(int id_set);
 static struct rtpe_node *select_rtpe_node(str, struct rtpe_set *);
+static struct rtpe_node *lookup_rtpe_node(struct rtpe_set * rtpe_list, str *rtpe_url);
 static char *send_rtpe_command(struct rtpe_node *, bencode_item_t *, int *);
 static int get_extra_id(struct sip_msg* msg, str *id_str);
 
@@ -364,6 +365,9 @@ static unsigned int rtpe_number = 0;
 
 static int     setid_avp_type;
 static int_str setid_avp;
+
+/* keys to be stored in the dialog */
+static str rtpe_url_dlg_key = str_init("rtpengine_url");
 
 /* tm */
 static struct tm_binds tmb;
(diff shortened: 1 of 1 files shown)


john08burke:
keepalive

john08burke:
Hey @razvancrainea, we are still running 3.1 so I have not yet explored the new `rtp_relay` module. I see the what you mean though, where the dlg storage/lookup is not needed if using this module.

Maybe I can refactor the PR such that the dlg lookup is moved to the top-level func calls associated with the now "lower-level" `rtpengine_xxxx` funcs. This way if the developer wants to use the `rtp_relay` module, then the API calls will use the ctx for node lookup... but if `rtpengine_xxxx` funcs are directly called then we will use dlg for node lookup.  For ex, `rtpengine_offer_answer(msg, flags, node, ...)` instead of `rtpengine_offer_answer(msg,flags, NULL, ...)` like is currently done.  Thoughts?

razvancrainea:
@john08burke moving them in a "lower-level" place means basically duplicating the logic for both rtpproxy and rtpengine - that's what we are trying to get rid of, and move everything in a more unified interface such as rtp_relay.
Having them in 3.1 is probably good enough for you, but unfortunately we can't merge features in a released branch :(.

john08burke:
Okay I see your point on duplicated efforts.  I will just continue to run with a locally patched 3.1, not a big deal!  Are there plans on exposing the common underlying functions such as offer, answer, delete, play in the future from the `rtp_relay` module?  We currently use the manual calls instead of the `rtpengine_manage` or `rtpproxy_engage`.  I'm a bit concerned about losing the lower-level control... but TBH I haven't ever tested the "manager" funcs.

razvancrainea:
Hi, John! No, there are no equivalents for the manual calls, but the module should be quite flexible to allow you do whatever/whenever you want.

john08burke:
Okay, I'll plan on migrating to this new module when it comes time!

PR #2611, opened 2021-08-25
Initial support for partition reload

nikbyte:
Hello opensips team,

I offer to discuss implementation of drouting partition reloading.
This patch is initial to understand where to move further.

It adds `dr_partition_reload` MI command which allows to dynamically reload partitions from database only.

Current limitations:
- only database partitions, doesn't work with fixed partitions
- only deletes and adds partitions, not modify current

What do you think about it?


nikbyte:
up

nikbyte:
?

PR #2612, opened 2021-08-25
rtpengine: enhancement of set/node management

john08burke:
**Summary**
This PR seeks to improve the mechanism by which nodes (and their respective sockets) are managed:
1. When loading set data from the database or module param, only allow one node per set per URL to be stored in memory. This cuts out redundancy in probing the same endpoint and shortens delays within the SIP procs when the redundant node is unreachable. If a user is putting multiple nodes in their set(s) to achieve load-balancing, the better approach is to use the existing weighting mechanism.
2. Extend the work done in 9ccd25fc282db8b2452254790c5492fc5829e382 to now completely remove node probing from the SIP context. With this PR, probing will be done at startup by each process (via `child_init`), by the timer procs, and by the HTTPD proc after a `rtpengine_reload` command. One con to this approach is that nodes are currently set to enabled, added to the set, and then probed.  If an unreachable node is added to the set via `rtpengine_reload`, then there is a chance that a SIP proc could select the node while the HTTPD proc is probing it. With the existing logic, the SIP proc would have been forced to first probe the node, then go onto the next node.  To address this, we could add a module param, say`default_node_state`, that could be used to dictate what behavior is more desirable (ie add node to set disabled, and wait for probing to enable before the node is selectable by SIP procs).
3. Free write lock from HTTPD proc before probing nodes. The justification is similar to (2) in that probing is pulled out of SIP context, so no need to lock the set from SIP readers.
4. Introduce "soft" reload option to MI `rtpengine_reload` command. The general idea with this type of reload is to reuse sockets and leave nodes enablement as is (ie if disabled, stay disabled) to minimize disruption to SIP procs. If the node exists in memory, then the existing node is reused (including all sockets and disabled state). If the node does not exist in memory, then it is added to the set and sockets are built per process. Remaining nodes are removed from the set and sockets are destroyed, as they have no associated database record.
5. Introduce version history for set list. Essentially, when a proc determines that it's time to reload socket data it traverses the version history to determine what type of reload to perform (ie soft or hard). I implemented this to allow for proper handling of hard and soft reloads, as race conditions pop up since set data is in shared memory while sockets are maintained per process (and multiple `rtpengine_reloads` could occur before proc hits reload logic).

I have this as a Draft, as it depends upon a commit 6123943afe21031dec8f4207f6f2bbc0dbb06b5d in #2603.  I also wanted to get some feedback from the OpenSIPS team to see your thoughts, feedback, and suggestions!

**Compatibility**
There should be no compatibility issues. The default reload behavior is to perform a hard reload of nodes/sockets, which is the existing behavior.

### DIFF
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -1447,6 +1447,15 @@ $ opensips-cli -x mi rtpengine_show
 			Reloads all rtpengine sets from the database. Used only when the
 			<quote><xref linkend="param_db_url"/></quote> parameter is set.
 			</para>
+			<para>Parameters:</para>
+			<itemizedlist>
+				<listitem><para>
+					<emphasis>type</emphasis> (optional) soft - when reloading nodes
+					from the database, reuse any existing sockets and keep existing
+					node disabled state. If not provided, then all nodes and sockets
+					will first be torndown and then nodes will be loaded from the database.
+				</para></listitem>
+			</itemizedlist>
 			<para>
 			No parameter.
 			</para>
@@ -1456,6 +1465,7 @@ $ opensips-cli -x mi rtpengine_show
 			<programlisting format="linespecific">
 ...
 $ opensips-cli -x mi rtpengine_reload
+$ opensips-cli -x mi rtpengine_reload type=soft
 ...
 			</programlisting>
 			</example>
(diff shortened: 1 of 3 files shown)


john08burke:
keepalive

bogdan-iancu:
@john08burke , is this PR completed (from your perspective)? I'm asking as initially was WIP marked.

john08burke:
Hey @bogdan-iancu! I just looked through it again and removed 1 feature that was previously discussed with @razvancrainea in #2603 (dialog support now supported via `rtp_relay` module, so no need to include here).

From my point it's ready for review. Thanks!

bogdan-iancu:
Hi @john08burke , sorry for the long waiting here. Before doing a review on the functionality and patch, could you confirm that it is aligned with all potential changes done in the mean while?

john08burke:
Hey @bogdan-iancu, I just reviewed and pushed 1 commit (2a1b57d3746190a27922c0f6d68acb351aa09f78) to bring a recent code change in-line with this PR. Otherwise, I believe it is ready for review. Thanks!

PR #2614, opened 2021-08-27
Wait for more data if WS header doesn't have extended length yet

danlshields:
`ws_read_raw` is only ran to receive data from the TCP connection if `req->tcp.parsed >= req->tcp.pos`. We fail to update `req->tcp.parsed` if we have enough data to read a minimum WS header but we don't have enough data to read the extended length part of the header.

I've changed the `return 0;` in the corresponding conditions to `goto update_parsed;` which is the code that is run if we don't have enough data to read a minimum WS header.


danlshields:
@razvancrainea What would it take to get this PR merged for this file you're an author of? It handles a corner case that was missed where ws_parse has enough data to parse the first part of the WebSocket header, but not the extended length. We encountered it when dealing with WebSocket connections from iOS 15.

Issue #2608, opened 2021-08-20
[BUG] Memory leak when querying sqlite module.


aniketkuwarvonage:
**OpenSIPS version you are running**

Version : 2.2.5 and 2.4.9

version: opensips 2.2.5 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
git revision: 355a035b4
main.c compiled on 15:02:28 May 10 2021 with gcc 8

version: opensips 2.4.9 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: a11653539

**Describe the bug**
OpenSIPS process leaks memory when running lb_reload [ load balancer ] or dr_load [ dynamic routing  command ] when reading from sqlite db.
Since the problem is in the sqlite module , leak could happen in other scenarios which are querying sqlite.

**To Reproduce**
Use sqlite db to store the load_balancer table.
Run lb_reload  openSIPS command in a loop or large number of times.

**Expected behavior**
OpenSIPS process should not leak memory.

**Relevant System Logs**
Running lb_reload in while loop for small time interval with DBG_MALLOC enabled and then restarting opensips prints below in logs :

14429056 : **652** x [res.c: db_sqlite_allocate_rows, line 275]

This indicates there are 652 allocations present in memory and not freed till opensips was restarted. They consumed close to 14MB of memory.

Here is data from server running opensips2.4.9 after 12 hour test :

185151352 : **230437** x [res.c: db_sqlite_allocate_rows, line 275]

There are 230437 entries and total leaked memory was 176 MB

**OS/environment information**
 - Operating System: Debian ( buster )
[...]

PR #2654, opened 2021-10-14
fix sqlite memory leak for row values

aniketkuwarvonage:
**Summary**
Details in ticket : https://github.com/OpenSIPS/opensips/issues/2608
OpenSIPS process leaks memory after running lb_reload [ load balancer ] when reading load balancer data from sqlite db.

**Details**
Fix is to avoid memory leak when freeing sqlite query's result values

**Solution**
After applying the fix , memory leak does not occur.

### DIFF
--- a/modules/db_sqlite/dbase.c
+++ b/modules/db_sqlite/dbase.c
@@ -55,6 +55,7 @@ static int db_sqlite_store_result(const db_con_t* _h, db_res_t** _r, const db_va
 static int db_sqlite_bind_values(sqlite3_stmt* stmt, const db_val_t* _v, const int _n);
 #endif
 static int db_sqlite_free_result_internal(const db_con_t* _h, db_res_t* _r);
+static void db_sqlite_free_result_rows(db_res_t* _r);
 
 static int db_sqlite_submit_dummy_query(const db_con_t* _h, const str* _s)
 {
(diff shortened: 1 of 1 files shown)


aniketkuwarvonage:
Closes #2608

modules/db_sqlite/dbase.c
```diff
@@ -824,6 +822,36 @@ int db_sqlite_free_result(db_con_t* _h, db_res_t* _r)
...
+	{
+		values = _r->rows[0].values;
+		/* db_sqlite_allocate_rows allocates memory for rows and values separately.
+		 * Hence freeing rows using generic function and then values separately*/
+		db_free_rows(_r);
+		if(values)
+		{
+			pkg_free(values);
+			values = NULL;
+		}
```
liviuchircu, line 849: This entire block can be replaced with just `pkg_free(values);`, for better readability, because:

* `pkg_free()` can handle `NULL` values
* "cleaning up" values by setting is to `NULL` is unnecessary, as it is a short-lived variable (stack only)

razvancrainea:
Thanks for the patch, and my apologies for taking so long to look into it. It's now pushed in all supported branches, including 2.4, although we will never release it - but it is good to have it there.

Issue #2649, opened 2021-10-11
[BUG] sql_cacher get incorrect string value, after restart opensips


GangZhuo:
**OpenSIPS version you are running**

```
version: opensips 3.3.0-dev (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, CC_O0, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: b4061d3d7
main.c compiled on 11:53:33 Oct 11 2021 with gcc 8
```

**Describe the bug**

sql_cacher get incorrect string value, after restart opensips. 

**To Reproduce**

```
modparam("sql_cacher", "cache_table",
"id=cache_passwd
db_url=postgres://uuu:123@192.168.1.77:5432/table_name
cachedb_url=redis:passwd://:123@127.0.0.1:6379/
table=user_info
key=id
key_type=int
columns=password
on_demand=1
expire=3600")

# Get password from database.
$var(password) = $sql_cached_value(cache_passwd:password:$var(user_id));

```

Example:

If we have a user with id=1000 and password="1234".

The sql_cacher automatically create a new redis key "passwd1000" with value "AAAAAA==EAAAAA==1234",
and $sql_cached_value(cache_passwd:password:1000) can get correct value "1234". 

After restart the opensips,  $sql_cached_value(cache_passwd:password:1000) get incorrect value "16".

**Additional context**

1) No problem when on_demand=0.
2) Maybe: After restarting OpenSIPs, sql_cacher did not reload the values from the database, so incorrectly defaulted the field to an integer type, and then returned the offset of string value, which value is 16 in "AAAAAA==EAAAAA==1234".

kertor:
Hello! I have the same problem on opensips 3.1:
```
version: opensips 3.1.6 (x86_64/linux)
[...]

PR #2676, opened 2021-11-09
[sql_cacher] When on demand mode, get column types at mod_init

GangZhuo:
Fix https://github.com/OpenSIPS/opensips/issues/2649

On demand mode, if the cache key exists, sql_cacher do not load data
from the database, the get_column_types() function will not be called,
so c_entry->column_types will remaining the initial value 0.
This will cause the wrong data type to be used to decode the cached values.

PS: I am not perfect on OpenSIPs, and not sure if this repair has side effects.


nikbyte:
Still actual.

rvlad-patrascu (approved):

PR #2689, opened 2021-11-17
Improve xyz_challenge() API to allow generating challenges like qop="auth,auth-int" as well as qop="auth-int,auth"; Make qop mandatory for the MD5-sess as well.

sobomax:
**Summary**
Improve  {www/proxy}_challenge() API to allow generating challenges like qop="auth,auth-int" as well as qop="auth-int,auth"; Make qop mandatory for the MD5-sess as well.

**Details**
This is useful for testing purposes as well as to indicate a preference in which UAC is supposed to perform authentication. Also clean code a bit to remove redundant declarations.

**Solution**
www_authenticate(..., "auth,auth-int", ...) would generate qop="auth,auth-int", www_authenticate(..., "auth-int,auth", ...): qop="auth-int,auth".

**Compatibility**
Tested with few UACs at OpenSIPIt'02.

### DIFF
--- a/lib/digest_auth/dauth_nonce.c
+++ b/lib/digest_auth/dauth_nonce.c
@@ -59,7 +59,7 @@ _Static_assert(offsetof(struct nonce_context_priv, pub) == 0,
 
 struct nonce_payload {
 	int index;
-	uint64_t qop:2;
+	uint64_t qop:3;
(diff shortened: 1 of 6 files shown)


liviuchircu:
Thanks, @sobomax, LGTM!

PR #2695, opened 2021-11-25
Fix SIP message corruption due to uninitialized buffer

dsandras:
The SIP ping message could not be parsed by softphones because the space
is added to the end of p. It results in an unitialized charachter after
the protocol string.

### DIFF
--- a/modules/nathelper/sip_pinger.h
+++ b/modules/nathelper/sip_pinger.h
@@ -362,7 +362,6 @@ build_sipping(udomain_t *d, str *curi, struct socket_info* s,str *path,
 	sbranch.len = strlen(branch);
 
 	p = proto2str(s->proto, proto_str);
-	*(p++) = ' ';
 	st.s = proto_str;
 	st.len = p - proto_str;
 
@@ -383,7 +382,7 @@ build_sipping(udomain_t *d, str *curi, struct socket_info* s,str *path,
 	*((int *)st.s) &= ~((1 << 21) | (1 << 13) | (1 << 5));
 
 	if ( sipping_method.len + 1 + curi->len + s_len(" SIP/2.0"CRLF) +
-		s_len("Via: SIP/2.0/") + st.len + address.len +
+		s_len("Via: SIP/2.0/") + st.len + 1 + address.len +
 		1 + port.len + strlen(branch) +
 		(path->len ? (s_len(CRLF"Route: ") + path->len) : 0) +
 		s_len(CRLF"From: ") +  sipping_from.len + s_len(";tag=") + 8 +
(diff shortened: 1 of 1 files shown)


liviuchircu:
Hi, @dsandras!  Ouch, that is some dangerous-looking code I wrote some 8 years ago!  The main culprit is this fragile code below, which overwrites a `\0` on top of the `" "` (space) 4th character, corrupting the SIP message with a `Via: SIP/2.0/UDP\0 ...` sequence.

```
	/* quick proto uppercase */
	*((int *)st.s) &= ~((1 << 21) | (1 << 13) | (1 << 5));
```

Checking the RFC grammar, this fix won't fully fix the issue, as the `SCTP` protocol is also a possible value:

```
transport         =  "UDP" / "TCP" / "TLS" / "SCTP"
                     / other-transport

```

... . This means that even with your fix, the code could still produce `Via: SIP/2.0/SCTp ...` in theory, which is also broken SIP, albeit in a very small number of use-cases.  So we need to eliminate that bitwise operation spaghetti and replace it with a proper loop that performs an uppercase only on `_isalpha()` characters encountered in the protocol string name...

If you'd like to write it, I'll happily review it!  Otherwise, we can close here and I'll push the complete fix soon.

dsandras:
Hi @liviuchircu , long time!

It's ok for me if you write it. I did not dig further as I am running after time currently. I wonder if it would not be easier to add some argument to `proto2str` perhaps instead of a loop. There are certainly other places where uppercase could be needed.

But that decision depends on you.

dsandras:
@liviuchircu Btw, is it certain that the transport over there must be uppercase?

liviuchircu:
Good question.  Some RFC excerpts below, just to give us an idea about the right answer.  I have **no idea** if OpenSIPS respects this or not, and I have no intention of finding out :)

```
           numbers.  To be compliant with this specification,
           applications sending SIP messages MUST include a SIP-Version
           of "SIP/2.0".  The SIP-Version string is case-insensitive,
           but implementations MUST send upper-case.
```

```
   When comparing header fields, field names are always case-
   insensitive.  Unless otherwise stated in the definition of a
   particular header field, field values, parameter names, and parameter
   values are case-insensitive.  Tokens are always case-insensitive.
```

liviuchircu:
Pushed a commit to 2.4+ that eliminates the bitwise ops on strings in that function.

Looking deeper into this, my earlier hypothesis with a byte that gets overwritten with `\0` is not correct, because the bitwise op was an `&=` with an inverted mask (see the `~`).  This means that even if an extra byte was read/written to, it would be with the same value.

This explains why nobody reported issues with nathelper's `build_sipping()` function all these years (myself included).  The only problem is that I don't really understand your bug now... maybe your issue is related to some architecture-specific behavior, where `1` behaves as `1ULL` instead of `1U`, or a compiler-specific behavior we're not aware of, etc.

In any case, if there are still issues after testing the fix, let me know!  Cheers!

dsandras:
@liviuchircu It's a better implementation, thanks ;-)

Issue #2625, opened 2021-09-10
[FEATURE] Fuzzer integration into OpenSips with OSS-Fuzz


DavidKorczynski:
Hi! I am proposing to integrate fuzzing into OpenSIPS by way of OSS-Fuzz. OSS-Fuzz is a free service run by Google that enables fuzzing of important open source projects. The goal is to find reliability and security issues that may exist by writing fuzzers for OpenSIPS and then have OSS-Fuzz run the fuzzers continuously. 

We integrated Kamailio into OSS-Fuzz earlier this year  (see here for the initial integration https://github.com/google/oss-fuzz/pull/5279) and it has reaped nice results - so I thought it would make sense to also get OpenSIPS in as well. In this PR on the OSS-Fuzz repo https://github.com/google/oss-fuzz/pull/6420 I suggest an initial integration of OpenSIPS. I created two simple fuzzers targeting the parsing code which should give some initial results.

In order to integrate I only need one or more emails that will be receiving bug reports whenever bugs are found by the fuzzers. These emails should be affiliated with a Google account such that you can see the detailed reports (like stack traces and reproducing cases) on https://oss-fuzz.com 

At the moment I put the fuzzers in the OSS-Fuzz repository, but we can also move these upstream down the line if you are happy with getting fuzzing into OpenSIPS.

Let me know if you like the idea and would like to get integrated.

liviuchircu:
Hey, @DavidKorczynski!  This sounds awesome, please enrol "liviu {at} opensips {dot} org" for any crash reports generated by the tool.  Let me know if you need any help with calling the parsing functions or any other info around the C project itself.

Mid/long-term, both myself and a few other core developers (Bogdan, Vlad & Răzvan) also like the idea of getting this automation merged into the OpenSIPS codebase, although it is not quite clear what the integration will look like (GH Actions?  some other tool?).

DavidKorczynski:
Wonderful @liviuchircu - the project is now ready to get merged in on the OSS-Fuzz side and will update you when it finally gets merged in. 

Many of the fuzzers look a lot like unit-tests, so what we could do is place them in a location similar to tests. In Kamailio we put all of the fuzzers in a folder `mics/fuzz` and then include the build scripts in the OSS-Fuzz repositories. We can do something similar here? 

OSS-Fuzz will run the fuzzers continuously so in that sense it is not strictly needed to run the fuzzers by way of the CI, however, we can also set that up down the line.

[...]

PR #2697, opened 2021-11-25
ci: add cifuzz integration

DavidKorczynski:
Add cifuzz integration so oss-fuzz fuzzers are run 600 secs each PR

Signed-off-by: David Korczynski <david@adalogics.com>

**Summary**

This adds CI integration with CIFuzz. CIFuzz will run the opensips OSS-Fuzz fuzzers (see this issue here https://github.com/OpenSIPS/opensips/issues/2625) for a short period of time for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

**Details**

CIFuzz is the Github action for OSS-Fuzz.

**Solution**

**Compatibility**

No compatibility issues.

**Closing issues**

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,5 @@
 name: CIFuzz
-on: [pull_request]
+on: [push, pull_request]
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,25 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'opensips'
+        dry-run: false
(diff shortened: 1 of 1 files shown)


liviuchircu:
Hey, @DavidKorczynski and thanks for this!  Could you make it run on every `push` instead of every `pull_request`?  (_I hope i got the verb right..._)

The reason for this is that 90%+ of OpenSIPS contributions to `master` branch are made via `git push`, rather than by clicking "Merge Pull Request".

DavidKorczynski:
I added `push` now - I enabled both as I figured having on pull as well doesn't hurt?

liviuchircu (approved):
LGTM!

DavidKorczynski:
Argh, we are actually running into an issue here because of CIFuzz. The issue is that the fuzzers are placed in OSS-Fuzz repo rather than opensips, and CIFuzz does some stuff where it deletes the opensips repository and then reclones it, which causes the fuzzers to be removed. The issue is known (https://github.com/google/oss-fuzz/issues/6755) and I have run into it with another project. I should be able to do a quick fix on the OSS-Fuzz repository. Will ping you shortly when the fix has been deployed @liviuchircu

DavidKorczynski:
It works now in my fork @liviuchircu (https://github.com/DavidKorczynski/opensips/actions) so I think it should work if you retrigger the CI

liviuchircu:
That's great!  Thank you, @DavidKorczynski!

PR #2699, opened 2021-11-26
Update rtpengine.c

xuxin010033:
In parse_flags() in rtpengine_offer(), the flag parameters are ignored after its val.s checked because of the "continue" after the "if else";
"break"s and "continue"s are switched for the checked parameters to be added to dict with "break", as well as the wrong writen parameters are skiped with "continue".

**Summary**

Flags passed to rtpengine_offer() are ignored before communicating with rtpproxy.

**Details**

In the circle in parse_flags(), after every cases the program jumped to the start when it encounters "continue" because the flag string are checked through; on the contrary the wrong writen parameters that can't be recognized are added to the bencode dict.

**Solution**

"break"s and "continue"s' positions are switched, so that right parameters reach the bencode_add part while the wrong ones can't.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -1689,8 +1689,8 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg,
 					ng_flags->transport |= 0x100;
 					ng_flags->transport &= ~0x002;
 				} else
-					break;
-				continue;
+					continue;
+				break;
 
 			case 4:
 				if (str_eq(&key, "SRTP"))
@@ -1698,8 +1698,8 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg,
 				else if (str_eq(&key, "AVPF"))
 					ng_flags->transport |= 0x102;
 				else
-					break;
-				continue;
+					continue;
+				break;
 
 			case 6:
 				if (str_eq(&key, "to-tag")) {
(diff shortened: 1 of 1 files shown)


xuxin010033:
Keys like ICE, transcode, codec-strip, codec-mask are still not properly processed, more work needed in the future...

bogdan-iancu:
@razvancrainea , could you take a look please ?

razvancrainea:
This is exactly the desired behavior. Only parameters that should not end up in the dictionary, but in other nodes (such as `transport`, `direction`, `packetize`, etc) are explicitly treated, everything else should be blindly passed to in dict or list.

xuxin010033:
Eg., if we want to specify a media address in SDP I should write "media-address=111.222.333.444" in the input of function "rtpengine_answer()".
When the program reaches the circle at line 1700, it gets to case 13, then it gets to if (!val.s) which gives us false, then skips 2 "else"s, then it gets to "continue" which directs us all the way back to line 1673 at the "while (1)", so that the "media-address=111.222.333.444" was not added to the dictionary.

razvancrainea:
When using `media-address=111.222.333.444`, the code would go through the `case 13`:
https://github.com/OpenSIPS/opensips/blob/0a6b2607fcf3282c941e4dedae122ddfc2fd927a/modules/rtpengine/rtpengine.c#L1844-L1849
There is a value, so the `goto error` will not be triggered, and it will proceed to `break` (it will never get to `continue`), which will add it to the dictionary.
I believe the problem you were facing was sorted out before (in commit 83e0bf6, where was indeed a `break` missing) you refactored the code, but somehow missed it during rebase.
There is no issue now afaik, so I will close the ticket. If I am missing something else, please re-open it.

Cheers,
Răzvan

PR #2700, opened 2021-11-26
Fix b2bl clean segfault

ar45:
**Summary**

When `b2bl_clean` is run by timer current_processing_context is not set

**Details**

**Solution**

Allocate and set a context.
I'm not sure if this is the right way to do it, but at least in my case it avoids the crash.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/b2b_logic/b2b_logic.c
+++ b/modules/b2b_logic/b2b_logic.c
@@ -635,6 +635,7 @@ void b2bl_clean(unsigned int ticks, void* param)
 	int i;
 	b2bl_tuple_t* tuple, *tuple_next;
 	unsigned int now;
+	static context_p my_ctx = NULL;
 
 	now = get_ticks();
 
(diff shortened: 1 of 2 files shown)


ar45:
Yes. @razvancrainea can you please review this?

ar45:
b2b_clean_bt

ar45:
@razvancrainea @bogdan-iancu can you have a look at this?

razvancrainea:
Hi, @ar45 !

I don't think that this is the right approach, as you are creating a single context for both possible messages that are being sent. I think the right approach would be for the context to be created at the TM level - I believe that for request is, but looks like it is not for the a reply.
However, there is one thing that makes me curious - what is the scenario that leads to this crash? I am curious how a b2b clean triggers a dialog callback.

Best regards,
Răzvan

ar45:
Still open

bogdan-iancu:
@ar45 , to be able to progress here, could you check @razvancrainea 's question here ?

> However, there is one thing that makes me curious - what is the scenario that leads to this crash? I am curious how a b2b clean triggers a dialog callback.

PR #2701, opened 2021-11-26
b2b_logic: Fix missing lock release

ar45:
**Summary**

Fix deadlock when `new_entities_no` is not 2.


ar45:
Yes

liviuchircu (approved):
LGTM, @ar45, especially since all error-cases from above and below also use `goto error` instead of `return NULL`.  Merging and backporting to all stable branches.

PR #2703, opened 2021-11-28
compression module: mc_compact() - Added flag `n`

ar45:
**Summary**
compression module: mc_compact() - Added flag `n`

**Details**
FEATURE

Added flags param which accepts `n` for no compact headers.
    When specified, opensips will only remove headers not in the whitelist
    without using the compact form for From/To/CallID etc.
    `mc_compact( ... , "n")`

**Solution**
The motive for this is, because some UAs do not understand the compact form
    I've also encountered weird bugs using the compact form with some UAs.

**Compatibility**
The flags param is optional and defaults to the old behavior to using compact header names.

### DIFF
--- a/modules/compression/compression.c
+++ b/modules/compression/compression.c
@@ -128,7 +128,6 @@ static int fixup_whitelist_compact(void**);
 static int fixup_whitelist_compress(void**);
 static int fixup_whitelist_free(void **);
 static int fixup_mc_compact_flags(void **);
-static int fixup_mc_compact_flags_free(void **);
 
 static int mc_compact(struct sip_msg* msg, mc_whitelist_p wh_list, int* flags_p);
 static int mc_compact_cb(char** buf, struct mc_compact_args* mc_compact_args, int, int*);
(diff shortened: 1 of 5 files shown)


modules/compression/compression.c
```diff
@@ -450,6 +463,45 @@ static int fixup_whitelist_free(void **param)
 	return free_whitelist(*param);
 }
 
+static int fixup_mc_compact_flags(void **param)
+{
+	str *s = (str *) *param;
+	int st;
+	int *flags;
+
+	flags = pkg_malloc(sizeof *flags);
```
bogdan-iancu, line 472: There is no need to allocate here the int for the flags, you can store the value directly over the `void *` (its size is larger or the same as of an `int`). Check how is done here https://github.com/OpenSIPS/opensips/blob/master/modules/drouting/drouting.c#L3870

ar45:
Still waiting

modules/compression/compression.c
```diff
@@ -158,6 +164,8 @@ static cmd_export_t cmds[]={
 	{"mc_compact",	  (cmd_function)mc_compact, {
 		{CMD_PARAM_STR|CMD_PARAM_OPT|CMD_PARAM_FIX_NULL,
 			fixup_whitelist_compact, fixup_whitelist_free},
+		{CMD_PARAM_STR|CMD_PARAM_OPT|CMD_PARAM_FIX_NULL,
```
bogdan-iancu, line 167: Please add the details on this new `n` flag into the documentation file ( modules/compression/doc/compression_admin.xml ).

bogdan-iancu (changes requested):
Just update the README

bogdan-iancu:
@ar45 , as per my request, you need to update the `modules/compression/doc/compression_admin.xml ` file (switch is not part of your PR here). There is no need to commit changes on the actual README file as this one is auto-generated based on the xml one.

bogdan-iancu (approved):

bogdan-iancu:
All good @ar45 , thanks for the work and patience here. I was about to merge your PR when I realized that it is against 3.2 branch, not against master . The 3.2 is a stable release, not accepting new features, only fixes. New features go only into master - so please re-do this PR against the masters (and close this one)

ar45:
@bogdan-iancu Thank you and sorry for that. I updated the PR to be against master.

Issue #2706, opened 2021-12-07
[BUG] Cannot launch opensips with oracle database


GangZhuo:
**OpenSIPS version you are running**
```
version: opensips 3.2.3 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, CC_O0, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: fb82bacce
main.c compiled on 14:04:49 Nov 29 2021 with gcc 8
```

**Describe the bug**

Using oracle database with load_balancer module, there are error occurs during the start-up phase and opensips going to exit.

**To Reproduce**

Using db_oracle with load_balancer modules together.

**Expected behavior**

work normal.

**Relevant System Logs**

```
Dec  7 11:16:07 [532566] DBG:core:db_do_init: connection 0x7faad16230b8 not found in pool
Dec  7 11:16:07 [532566] DBG:db_oracle:db_oracle_new_connection: opening connection: oracle://xxxx:xxxx@xxx.xxx.xxx.xxx:1521/opensips
Dec  7 11:16:07 [532566] INFO:db_oracle:db_oracle_new_connection: server version is Oracle Database 19c Enterprise Edition Release 19.0.0.0.0 - Production
Dec  7 11:16:07 [532566] DBG:core:db_do_init: connection 0x7faad16230b8 inserted in pool as 0x7faad1622a18
Dec  7 11:16:07 [532566] DBG:core:db_new_result: allocate 48 bytes for result set at 0x7faad1622198
Dec  7 11:16:07 [532566] DBG:core:db_allocate_columns: allocate 28 bytes for result columns at 0x7faad1622820
Dec  7 11:16:07 [532566] DBG:db_oracle:get_columns: use DB_INT result type
Dec  7 11:16:07 [532566] DBG:core:db_allocate_rows: allocate 48 bytes for result rows and values at 0x7faad16221f8
Dec  7 11:16:07 [532566] DBG:core:db_free_columns: freeing result columns at 0x7faad1622820
[...]

PR #2707, opened 2021-12-07
[db_oracle] Using DB_STRING as character type

GangZhuo:
**Changes**

1) Using DB_STRING as character type in db_oracle module (Fix https://github.com/OpenSIPS/opensips/issues/2706)

Modules (e.g. load_balancer) using DB_STRING as character type to checking column type,
but the db_oracle module using DB_STR, this cause checking fail during the start-up phase and opensips going to exit.

2) Fix crash on db_oracle_raw_query()

The Function db_oracle_submit_query() required that CON_ORA(_h)->pqdata is not null, but db_oracle_raw_query() do not set.

### DIFF
--- a/modules/db_oracle/dbase.c
+++ b/modules/db_oracle/dbase.c
@@ -385,6 +385,8 @@ badparam:
 		return -1;
 	}
 
+	CON_ORA(_h)->pqdata = &cb;
+	CON_ORA(_h)->bindpos = 0;
 	CON_RESET_CURR_PS(_h); /* no prepared statements support */
 	memset(&cb, 0, sizeof(cb));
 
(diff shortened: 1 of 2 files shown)


liviuchircu (approved):
Hey @GangZhuo!  I reviewed your commits and they both LGTM:

* by now, it is well-known that **all** modules expect `DB_STRING` from an SQL database module, with most of them not handling `DB_STR` fully.  A similar adaptation was done for `cachedb_mongodb` in f03a958482 a year ago, to improve compatibility with various modules.
* your `->pqdata` fix looks correct, as it fixes an obvious crash: things can only go up from here!

Thank you!

GangZhuo:
That's great!

PR #2731, opened 2022-01-13
dispatcher: add optional partition param to MI `ds_list`

john08burke:
**Summary**
This PR adds an optional ***partition*** param to the `ds_list` MI command.

**Details**
When working with large dispatcher data, the response JSON can be quite large when calling `ds_list` as it currently returns all partition data. If partitions are used, it would be nice to include a way to filter the response at the server level to cut down on the response size.

**Compatibility**
Backwards compatible, as this is an optional param; default is still to return all data.

### DIFF
--- a/modules/dispatcher/dispatcher.c
+++ b/modules/dispatcher/dispatcher.c
@@ -333,7 +333,9 @@ static mi_export_t mi_cmds[] = {
 	},
 	{ "ds_list", 0, 0, 0, {
 		{w_ds_mi_list, {0}},
+		{w_ds_mi_list, {"partition", 0}},
 		{w_ds_mi_list_1, {"full", 0}},
+		{w_ds_mi_list_1, {"full", "partition", 0}},
 		{EMPTY_MI_RECIPE}}
 	},
 	{ "ds_reload", 0, 0, mi_child_init, {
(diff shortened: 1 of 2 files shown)


modules/dispatcher/dispatcher.c
```diff
@@ -1281,22 +1283,40 @@ mi_response_t *ds_mi_set(const mi_params_t *params,
...
 	mi_item_t *parts_arr, *part_item;
-	ds_partition_t *part_it;
+	ds_partition_t *partition, *part_it;
 
 	resp = init_mi_result_object(&resp_obj);
 	if (!resp)
 		return 0;
 
+	if (try_get_mi_string_param(params, "partition", &partition_name.s, &partition_name.len) == 0)
+		filter_partition = 1;
```
bogdan-iancu, line 1300: I think you can get rid of the `filter_partition` flag and use `partition` for that purpose too. Init `partition` to NULL and have here the block for searching the partition by name, under this initial `if`.

modules/dispatcher/dispatcher.c
```diff
@@ -1281,22 +1283,40 @@ mi_response_t *ds_mi_set(const mi_params_t *params,
...
+		if (partition == NULL)
+			return init_mi_error(500, MI_SSTR(MI_UNK_PARTITION));
+	}
+
 	parts_arr = add_mi_array(resp_obj, MI_SSTR("PARTITIONS"));
 	if (!parts_arr)
 		goto error;
 
 	for (part_it = partitions; part_it; part_it = part_it->next) {
+		if (filter_partition && str_strcmp(&partition->name,&part_it->name))
```
bogdan-iancu, line 1317: And here test if `partition` is NULL and if not, do the filtering. BTW, IMO, the partition matching can be directly done, by pointer, like `partition==part_it` as these are structures in shared mem.

modules/dispatcher/dispatcher.c
john08burke: Thanks Bogdan, the recent push should address this.

modules/dispatcher/dispatcher.c
john08burke: This should be addressed as well.

john08burke:
No problem, thanks for reviewing!

bogdan-iancu:
We can ignore for now this [report](https://github.com/OpenSIPS/opensips/pull/2731#issuecomment-1015563551) , it is not the fault of this PR, as it existed before it. The proper fix would be to rename the `static param_export_t params[]` to something else, its name it tooooo generic for a global variable.

PR #2735, opened 2022-01-18
Kworm83 stir shaken patch

kworm83:
**Summary**
Patch to bring Identity header more inline with current stir/shaken real world implementations

**Details**
In looking at standards/example documentation along with actual Identity headers received from North American carriers I have noticed that they include the "alg=ES256" parameter in the Identity header as well as not having double quotes around shaken in the ppt parameter.   I'm not sure that poses a problem with actual signing but this patch makes the Identity header for OpenSIPS match what is used by large carriers such as Bandwidth, Verizon, ANPI/Intelliquent/etc.  I ran across this while testing our own signing and hearing from other carriers that our Identity header was different in the above regards.

**Solution**
My PR adds the alg=ES256 parameter and remove the double quotes to make the ppt=shaken parameter

**Compatibility**
I do not believe this change will break any Identity headers as the largest carriers in NA are using them.  I do not have a way to test against any European/Latin America/etc carriers but assuming they are using the same standards it will be correct for them as well.

**Closing issues**
none

### DIFF
--- a/modules/stir_shaken/stir_shaken.c
+++ b/modules/stir_shaken/stir_shaken.c
@@ -784,7 +784,7 @@ static int add_identity_hf(struct sip_msg *msg, EVP_PKEY *pkey,
 
 	hdr_buf.len = IDENTITY_HDR_LEN + unsigned_buf.len + 1/*'.'*/ +
 		calc_base64_encode_len(RAW_SIG_LEN) + 1/*';'*/ + HDR_INFO_PARAM_LEN +
-		2/*'<','>'*/ + cr_url->len + 1/*';'*/ + HDR_PPT_PARAM_LEN + CRLF_LEN;
+		2/*'<','>'*/ + cr_url->len + 1/*';'*/ + HDR_ALG_PARAM_LEN + 1/*';'*/ + HDR_PPT_PARAM_LEN + CRLF_LEN;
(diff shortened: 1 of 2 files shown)


modules/stir_shaken/stir_shaken.c
```diff
@@ -814,6 +814,9 @@ static int add_identity_hf(struct sip_msg *msg, EVP_PKEY *pkey,
 	hdr_buf.len += cr_url->len;
 	hdr_buf.s[hdr_buf.len++] = '>';
 	hdr_buf.s[hdr_buf.len++] = ';';
+	memcpy(hdr_buf.s + hdr_buf.len, HDR_ALG_PARAM_S, HDR_ALG_PARAM_LEN);
++       hdr_buf.len += HDR_ALG_PARAM_LEN;
++       hdr_buf.s[hdr_buf.len++] = ';';
```
rvlad-patrascu, line 819: Can you remove the '+' characters at the start of these lines?

rvlad-patrascu (changes requested):

modules/stir_shaken/stir_shaken.c
kworm83: I have removed the '+' characters and build checks are running now.

rvlad-patrascu (approved):

PR #2742, opened 2022-01-27
domain: add support for `DB_BLOB` attrs type

john08burke:
**Summary**
The PR allows the domain module to accept BLOB type from database for attrs column.  If using the standard opensips [schema](https://github.com/OpenSIPS/opensips/blob/master/scripts/mysql/domain-create.sql) for the domain module, the column is always of CHAR type.  However, if using custom schema it's often times desirable to setup a VIEW to some other table/DB that provides configuration data in a different schema.  The problem is that simple SELECT statements in the VIEW definition can lead to BLOB fields instead of CHAR in the response, even when explicit CASTing is done.

For example, doing a simple GROUP results in BLOB if `group_concat_max_len` <= 512 (https://mariadb.com/kb/en/group_concat/):

>If group_concat_max_len <= 512, the return type is VARBINARY or VARCHAR; otherwise, the return type is BLOB or TEXT. The choice between binary or non-binary types depends from the input.

**Solution**
Also parse the DB_BLOB type.

**Compatibility**
There should be no compatibility issues.

### DIFF
--- a/db/db_val.h
+++ b/db/db_val.h
@@ -227,6 +227,17 @@ static inline void db_print_val(db_val_t *v)
 			} else { \
 				_str = VAL_STR(_val); \
 			} \
+		} else if ((_val)->type==DB_BLOB) { \
+			if ( VAL_BLOB(_val).s==NULL || VAL_BLOB(_val).len==0 ) { \
+				if (_not_empty) { \
+					LM_ERR("value in column %s cannot be empty\n", _col_name); \
+					goto _error_label;\
+				} else { \
+					_str = VAL_BLOB(_val) ;\
+				} \
+			} else { \
+				_str = VAL_BLOB(_val); \
+			} \
 		} else {\
 			LM_ERR("column %s does not have a string type (found %d)\n",\
 				_col_name,(_val)->type); \
--- a/modules/domain/domain.c
+++ b/modules/domain/domain.c
@@ -271,6 +271,8 @@ int reload_domain_table ( void )
 			attrs.len = strlen(attrs.s);
 		} else if (VAL_TYPE(val + 1) == DB_STR) {
 			attrs = VAL_STR(val + 1);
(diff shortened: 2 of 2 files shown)


john08burke:
I also just pushed a general patch for `DB_BLOB` support inside `get_str_from_dbval` as I ran into the same issue w/ the dispatcher module.

On a similar note (and assuming you're on-board w/ the idea behind the PR), the mysql/mariadb docs both state that **either** BLOB or TEXT can be returned:
- https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html#function_group-concat
>The return value is a nonbinary or binary string, depending on whether the arguments are nonbinary or binary strings. The result type is TEXT or BLOB unless group_concat_max_len is less than or equal to 512, in which case the result type is VARCHAR or VARBINARY.
- https://mariadb.com/kb/en/group_concat/
>If group_concat_max_len <= 512, the return type is VARBINARY or VARCHAR; otherwise, the return type is BLOB or TEXT. The choice between binary or non-binary types depends from the input.

I personally have only seen BLOB, but IMO support should also be added for TEXT as I'm not sure of what conditions dictate respective return type. Currently it doesn't look like the `DB_TEXT` type is supported in OpenSIPS.

liviuchircu (approved):
LGTM to me, @john08burke!  And the fact that you improved `get_str_from_dbval()` is even better - who knows in what ways it might help dispatcher, fraud_detection and freeswitch_scripting (the only modules which are currently using it).  Good job!

I will merge to `master` ASAP, but will also consider backporting these down to `3.1` in the meantime -- being more acceptable towards VIEWs can only lead to good!

liviuchircu:
Regarding the `DB_TEXT` idea... I haven't seen it either returned by the MySQL driver so far.  So maybe it really isn't that big of an issue, if any.  But at least we're aware that it may pop up - as soon as it does, we can build even more checks into the code and backport all of them.

PR #2757, opened 2022-02-22
Fix crash in tracer module when negative hop-by-hop ACK received during b2b tracing

nikbyte:
This PR fixed tracer module crash during b2b tracing.

end2end ACK forms a separate transaction
(even if TM will return a NULL transaction) and
we will trace it as standalone request, while a negative hop-by-hop ACK
(part of INVITE transaction) we will get T_UNDEFINED, so not to be traced

try to trace T_UNDEFINED ACK causes crash

### DIFF
--- a/modules/b2b_entities/dlg.c
+++ b/modules/b2b_entities/dlg.c
@@ -1091,9 +1091,15 @@ logic_notify:
 
 		tm_tran = tmb.t_gett();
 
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Accepted and backported to 3.2 also, thanks @nikbyte

PR #2764, opened 2022-02-28
incorrect error handling.

ihsinme:
**Summary**
strcpy does not return NULL. 

**Solution**
if the specified check is necessary to control undefined behavior in the case of `config_file` equal to `NULL` and an overflow occurs on `if(!strcat(backup_file, ".bak")){`, then it is proposed to simply eliminate this situation by nulling the `backup_file` array.

if in this way you are trying to implement an integer overflow of `strlen(config_file) + strlen (".bak") + 1)`, with large string lengths of `config_file`. Then this check leads to the same uselessness and `backup_file` will overflow on the heap.


modules/carrierroute/route_config.c
```diff
@@ -351,10 +351,8 @@ static int backup_config(void) {
 		LM_ERR("out of private memory\n");
 		return -1;
 	}
-	if(strcpy(backup_file, config_file) == NULL){
-		LM_ERR("can't copy filename\n");
-		goto errout;
-	}
+	backup_file[0] = 0;
```
razvancrainea, line 354: I totally agree `strcpy` should not be checked against `NULL`, as `config_file` can never be `NULL`. Enough space should be in `backup_file` pointer as well, so I can't see the reason of checking the return of `strcat` either. But I can't see the reason when "nulling" the `backup_file` might be useful, can you please elaborate?
Personally, in order to sort out this warning/issue, I would drop `strcpy/strcat` in favor of `memcpy`.

liviuchircu: One concise way of merging both blocks could be using a simple `sprintf(backup_file, "%s.bak", config_file);` - no need to check the result either.  Regarding "nulling" the `backup_file` beforehand, it seems unnecessary to me as well: the `dst` string is just a buffer.

liviuchircu: It's superfluous: assuming `config_file` could be `NULL` (which is not possible, as it has a default value, so now the script writer must explicitly set it to `NULL` from script - but that's illegal), then the code will crash anyway much earlier, on the very 1st `strlen(config_file)` on line 350.

So @ihsinme: could you rewrite both of those code blocks into a simple `sprintf()`, since the buffer is sufficiently large to hold all data?  Then we can merge it.  Thanks!

ihsinme: @liviuchircu 
look please

ihsinme: @liviuchircu look please

PR #2773, opened 2022-03-08
Use library function to fix edge case in require parsing

carstenbock:
**Summary**
Use library function to fix edge case in "Require" parsing

**Details**
We've noticed, that in some edge-cases the parsing of 100rel failed in the Require-Header

**Solution**
Using the internal parser from the core for the Requires header fixes this issue.

Kudos go to my colleague Andrii (andrii@ng-voice.com) for identifying and fixing this.


bogdan-iancu:
Thank you @carstenbock . Indeed, the old code worked only if you had only the 100rel token in the Required hdr, nothing else; and normally there is a list there.

PR #2781, opened 2022-03-16
Add parameters for PRACK generation and Route checking

carstenbock:
**Summary**
Add options for locally generating PRACK and for checking of additional route headers.

**Details**
We needed the PRACK to be end-2-end in some scenarios. Also, we have some scenarios, where we have multiple Route headers and the b2b_entities failed to associate the requests based on these multiple route headers.

**Solution**
This PR allows the configuration of this behavior.

**Compatibility**
Default configuration is the same as previously - it should not break anything.


modules/b2b_entities/dlg.c
```diff
@@ -762,7 +762,7 @@ int b2b_prescript_f(struct sip_msg *msg, void *uparam)
 				rt = (rr_t*)route_hdr->parsed;
 			}
 		}
-		if (rt) {
+		if (check_more_routes && rt) {
```
bogdan-iancu, line 765: In terms of routing, the b2b is the end point. So, no route hdr should point to something else than the B2B - and this only in the case of a pre-loaded route. 
By turning the `check_more_routes` off, the b2b will check only the first route (to see if belongs to it), ignoring any additional Route that may point to another SIP server (maybe the request you have here is not even for the B2B, it may be a request to be proxies at script level, with a complex stack of Route hdrs). So this option may actually break the routing from SIP perspective. 
Do you have a clear example of the issue you are trying to address here ? (just trying to understand the this new option and to be sure we do not end up with broken behaviors )

bogdan-iancu (changes requested):

modules/b2b_entities/dlg.c
```diff
@@ -3204,7 +3204,7 @@ void b2b_tm_cback(struct cell *t, b2b_table htable, struct tmcb_params *ps)
 					}
 					hdr = hdr->sibling;
 				}
-				if(hdr)
+				if(hdr && generate_prack)
```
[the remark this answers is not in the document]
carstenbock: Refactored this into "passthru_prack" + updated the documentation. Changed it to exclude the checking of the Require Header. Thanks!

modules/b2b_entities/dlg.c
carstenbock: As a matter of fact, the issue came up, since we are using preloaded routes for the initial request. The preloaded route of the incoming leg comes in the following format:

Route: <b2bua>
Route: <next-hop-after-b2bua> <= This needs to be passed on when originating the request.

As the second Route header needs to be passed through to the B-Leg, we added it to the custom_headers parameter of b2b_logic:

modparam("b2b_logic", "custom_headers", "[...];Route;[...]")

However, this caused it to be passed-through also for subsequent requests, which resulted with an issue with the Route-Headers at a different instance, where it found the second Route header (one for the dialog and the other passed through).

I will come up with better approach and remove that parameter for now.

bogdan-iancu: hmm, passing thru a `Route` hdr via a B2B .... it looks to me you are asking it for troubles :) . Maybe you should do something at script level, for the initial INVITE (with the preloaded Route), to get the second value and try to push it further in the B leg manually (via the b2b context, so you can add it later in local_route) ?

bogdan-iancu (approved):

PR #2783, opened 2022-03-16
b2b_entities: parse from & to structs if they aren't

svanstone-resilient:
This fix aims to solve a crash in case the from/to headers are not parsed (yet). We are parsing them in a different structure, to make sure it doesn't overlap with any ongoing processing of the request.

(Cherry-picked from 3.2 and 5e5fc98ad07889c6d285a913aee7caf769c2878b )


svanstone-resilient:
Hi. I still care about this one.
It caused a crash when the B2BUA module was used and there was a timeout.
3.1 is still supported, so it'd be good for it to continue getting bug fixes.

PR #2784, opened 2022-03-17
B2B Logic: Dynamic Contact-Header + Trigger Scenario from Script

carstenbock:
**Summary**
This PR adds the following functionality:
- Allow pseudo-variables in the contact header
- Trigger a scenario from the script (e.g. for out-of-dialog REFERs)

**Details**
1) The Contact-Header parameter has some limitations, as it becomes challenging when having different aliases defined for different sockets. With this PR, you can use all kinds of Pseudo-Variables for the Contact header, e.g. the Contact header could even be defined in the script.
2) Added functionality to trigger a scenario from Script. This may be used to for example handle an out of dialog REFER or if you want to process multiple in-dialog REFERs without touching the initial session.

**Compatibility**
No impact on existing configurations, the default behavior stays the same.

### DIFF
--- a/modules/b2b_logic/logic.c
+++ b/modules/b2b_logic/logic.c
@@ -4362,7 +4362,7 @@ int b2b_bridge_extern(struct sip_msg* msg, str *id, str * params,
 		goto end;
 	}
 	tuple->lifetime = 60 + get_ticks();
-	LM_ERR("Key: %.*s (%p)\n", b2bl_key->len, b2bl_key->s, tuple);
+	LM_DBG("Key: %.*s (%p)\n", b2bl_key->len, b2bl_key->s, tuple);
 
 	if (params && params->s && params->len > 0) {
 		param_list = parse_csv_record(params);
@@ -4384,7 +4384,7 @@ int b2b_bridge_extern(struct sip_msg* msg, str *id, str * params,
 				remote_tuple = s;
 				s = param_list->next->next ? &param_list->next->next->s : NULL;
 				if (s && s->s && s->len) {
-					if (s->s[0] == 1) remote_tuple_party = 1;
+					if (s->s[0] == '1') remote_tuple_party = 1;
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


modules/b2b_logic/logic.c
```diff
@@ -4106,6 +4134,360 @@ int b2bl_bridge(str* key, str* new_dst, str *new_proxy, str* new_from_dname,
...
+	tuple = b2bl_insert_new(msg, hash_index, &init_params,
+		NULL, NULL, -1, &b2bl_key, INSERTDB_FLAG, TUPLE_NO_REPL);
+	if(tuple== NULL)
+	{
+		LM_ERR("Failed to insert new scenario instance record\n");
+		rc = -1;
+		goto end;
+	}
+	tuple->lifetime = 60 + get_ticks();
+	LM_ERR("Key: %.*s (%p)\n", b2bl_key->len, b2bl_key->s, tuple);
```
rvlad-patrascu, line 4365: This should be a LM_DBG log.

modules/b2b_logic/logic.c
```diff
@@ -4106,6 +4134,360 @@ int b2bl_bridge(str* key, str* new_dst, str *new_proxy, str* new_from_dname,
...
+			goto end;
+		}
+		if (s->s[0] == 'n') {
+			tuple->bridge_flags = B2BL_BR_FLAG_NOTIFY | B2BL_BR_FLAG_DONT_DELETE_BRIDGE_INITIATOR;
+			s = param_list->next ? &param_list->next->s : NULL;
+			if (s && s->s && s->len) {
+				remote_tuple = s;
+				s = param_list->next->next ? &param_list->next->next->s : NULL;
+				if (s && s->s && s->len) {
+					if (s->s[0] == 1) remote_tuple_party = 1;
```
rvlad-patrascu, line 4387: It seems that this param identifies the old entity that receives the NOTIFY, given as an index in the bridged entities array, so the comparison should probably be `if (s->s[0] == '1')`. Can you also clarify a bit in the docs that this is an extra parameter besides the b2b_logic key?

rvlad-patrascu (commented):
Besides the small fixes suggested, the PR looks ok and ready to be merged.

modules/b2b_logic/logic.c
carstenbock: Found a few more places, where it should have been LM_DBG => resolved!

modules/b2b_logic/logic.c
carstenbock: Improved the comparison and improved the docs. Thank you for noticing!

rvlad-patrascu (approved):

PR #2790, opened 2022-03-24
Add TOS in IPV6

kpouer:
**Summary**
TOS support is not working in IPV6

**Details**
When using IPV6 the tos option do not apply

**Solution**
Check the transport type to apply appropriate TOS flags

**Compatibility**
Should be fully compatible

### DIFF
--- a/net/net_udp.c
+++ b/net/net_udp.c
@@ -180,18 +180,18 @@ int udp_init_listener(struct socket_info *si, int status_flags)
 		goto error;
 	}
 	/* tos */
-    	optval=tos;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
@kpouer , thank you for the patch, but please update it and use onyl tabs (no spaces) for indentation of the new code, so we will keep the proper alignment. Thank you !

kpouer:
Hey @bogdan-iancu thanks for taking the time to take a look at this pull request. I fixed the indentation, but maybe you would prefer another pull request with all changes within only one commit ?

bogdan-iancu:
Thank you @kpouer !

PR #2791, opened 2022-03-24
Fix /var/run to /run

vasilevalex:
**Summary**
Filesystem Hierarchy Standard 3.0 released in 2015.
It's time to fix /var/run to /run

### DIFF
--- a/etc/opensips.cfg
+++ b/etc/opensips.cfg
@@ -69,7 +69,7 @@ loadmodule "sipmsgops.so"
 
 #### FIFO Management Interface
 loadmodule "mi_fifo.so"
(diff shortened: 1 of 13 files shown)


bogdan-iancu:
Hi @vasilevalex . Indeed, but:

1. I see that most of the system do sym link from /var/run to /run, so we are still good
2. what are the chances to have a 7 years old OS ? do we rule this completely out (as OpenSIPS will no longer run on something like that)

just my 2 cents :)

vasilevalex:
Hi @bogdan-iancu 
1) Yes, I agree, that OpenSIPS can work without these changes. They are not mandatory. Just in modern systems you'll catch some warnings.
2) All the changes are not dangerous. There are only config files, docs and default values. For old system directories can be changed in the config files.

PR #2820, opened 2022-04-24
ip_addr: simplify check for RFC 1918

lemenkov:
This function is used everywhere as follows:

```c
flag = (ip_addr_is_1918(x)==1) ? 1 : 0;
```

Let's just make it return the expected 0 or 1.


bogdan-iancu:
:+1:  @lemenkov

PR #2825, opened 2022-04-27
core: fix bogus "delay in execution" warnings

### LINKED ISSUE #2812: [BUG] Bogus warning in the logs about timers being late

All recent versions of opensips.

**Describe the bug**

The ijiffies are adjusted every 5 min by the run_timer_process_jif() in a dedicated proc named time_keeper.

The timer tasks are dispatched by the run_timer_process() in a dedicated proc named timer.
Before dispatching the time jobs, the wait time between each job dispatch is computed via compute_wait_with_drift() and the jobs are dispatched via (u)timer_ticker() procedure. Most of the time (u)timer_ticker() procedure doesn't need to dispatch any jobs (on most deployments it is called every 100ms and for tm-timer is called every 10th iteration and for tm-utimer is called every 2nd iteration).

If the ijiffies are adjusted at the same time when a timer is scheduled for execution, we will have a false warning about the timer being late. This is happening when the ijiffies are adjusted after the trigger_time is set inside the (u)timer_ticker in the timer process and before the timer task is handled by the handle_timerjob() inside the "Timer handled" proc.

This could be avoided by keeping track of the ijiffies drift and tacking that into consideration while evaluating the print of the warning (in addition to the ITIMER_TICK).

ovidiusas:
- time ijiffies are adjusted by the time_keeper process
 - when the trigger time for a time job is set right
   before adjusting the ijiffies and the processing
   of the timer job is done right after the ijiffies
   adjustment, there's an artificial delay (the ijiffies
   adjutment) that triggers a bogus "delay in execution"
   warning
 - by taking into consideration the ijiffies adjutment
   we avoid printing the bogus "delay in execution" warning

closes #2812

### DIFF
--- a/timer.c
+++ b/timer.c
@@ -62,7 +62,7 @@
 /* define internal timer to 10 milliseconds */
 #define ITIMER_TICK 10000
 
-/* try to synchronize with system time every 5 seconds */
+/* try to synchronize with system time every 5 second(s) */
 #define TIMER_SYNC_TICKS 5000000
 
 /* synchronize if drift is greater than internal timer tick */
@@ -77,6 +77,8 @@ static struct os_timer *utimer_list = NULL;
 static unsigned int  *jiffies=0;
 static utime_t       *ujiffies=0;
 static utime_t       *ijiffies=0;
+static utime_t       *ijiffies_drift=0;
+static utime_t       *ijiffies_drift_base=0;
 static unsigned short timer_id=0;
 static int            timer_pipe[2];
 static struct scaling_profile *s_profile=NULL;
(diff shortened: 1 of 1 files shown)


volga629-1:
in

volga629-1:
In progress

sobomax:
All this jiffies nonsense needs to be ripped out and replaced with clock_gettime(CLOCK_MONOTONIC). In My Humble Opinion.

ovidiusas:
> All this jiffies nonsense needs to be ripped out and replaced with clock_gettime(CLOCK_MONOTONIC). In My Humble Opinion.

Someone needs to bite the bullet, implement it and test it! :)

bogdan-iancu:
@ovidiusas , I merged your fix into master (with a bit of a change), please see https://github.com/OpenSIPS/opensips/commit/1d83a0c77849a8d0741f0e2824f73288b3032af8 - hopefully nothing got broken, but if you could try it and confirm me that everything is ok...so I can do the backport ;)
Thanks !!

ovidiusas:
I see that the printing of the WARNING was optimized (hopefully the behavior was preserved - it's been a long time since I implemented the fix and I don't recall all the details).
Looks pretty safe to backport.

bogdan-iancu:
Thank you @ovidiusas , backports done !

PR #2830, opened 2022-05-06
clusterer: add statistics for number of nodes and number of up nodes

hafkensite:
**Summary**
Basic node statistics for clusterer module.

**Details**
To monitor the cluster health state we want to have statistics about the number of nodes in the `Up` state, compared to the total number of nodes.

**Solution**
I copy/pasted some of the statistics code from other modules, should work the same as other modules that have custom statistics.
Not sure about the code-style and consequences of reloading or updating the nodes. 

**Compatibility**
No, should be good. Statistics can be disabled if desired.

**Closing issues**


razvancrainea:
A reload adding and/or removing a new node will definitely break the `clusterer_nodes_up` statistic, as it will not account for new/deleted nodes.. I think the easier way to fix this would be to use a function for computing it as well by only accounting for nodes that `link_state == LS_UP`.

PS: please also document the new statistics in the [xml](https://github.com/OpenSIPS/opensips/blob/master/modules/clusterer/doc/clusterer_admin.xml)

hafkensite:
> A reload adding and/or removing a new node will definitely break the `clusterer_nodes_up` statistic, as it will not account for new/deleted nodes.. I think the easier way to fix this would be to use a function for computing it as well by only accounting for nodes that `link_state == LS_UP`.

Good point, I separated the counting into a function, calling it for each of the stats. 
 
> PS: please also document the new statistics in the [xml](https://github.com/OpenSIPS/opensips/blob/master/modules/clusterer/doc/clusterer_admin.xml)

Done, I assume the README will be automagically updated?

hafkensite:
I would still like to see this metric added, a healthy cluster is pretty important

hafkensite:
@razvancrainea, would you be willing to review this PR again? I processed the changes you suggested, but if you find more issues, I can do another iteration.

rvlad-patrascu (changes requested):
Hi @hafkensite, 

Can you improve the docs a bit by adding a separate "Exported Statistics" section that describes each provided statistic ([example)](https://opensips.org/docs/modules/3.3.x/dialog.html#exported_statistics)?

Also, I think that for the _clusterer_nodes_down_ statistic it would be more useful to account for any link state other that `LS_UP` since any probing state effectively means that the node is not reachable at that time (for replication traffic at least). This would also be consistent with the "node down" meaning of the _E_CLUSTERER_NODE_STATE_CHANGED_ event.

rvlad-patrascu (approved):

Issue #2846, opened 2022-05-27
[BUG] ERROR:db_oracle:done_timelimit: driver: ORA-03126: network driver does not support non-blocking operations


GangZhuo:
**OpenSIPS version you are running**

```
version: opensips 3.2.6 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, CC_O0, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 1628c902a
main.c compiled on 13:15:40 May 27 2022 with gcc 8
```

**Describe the bug**

Using oracle database via db_oracle module, after some time, the opensips cannot perform insert/delete operations.

**To Reproduce**

The database has only the following tables: 

- address
- clusterer
- dialog
- domain
- load_balancer
- version

I think only dialog table need to be read and write data, other tables should be read data once at startup, and no write after startup.

**Expected behavior**

Can be to read and write to the database normally

**Relevant System Logs**

```
May 27 13:16:13 localhost systemd[1]: Starting OpenSIPS is a very fast and flexible SIP (RFC3261) server...
May 27 13:16:13 localhost opensips[131639]: Listening on
May 27 13:16:13 localhost opensips[131639]:              udp: 192.168.1.8 [192.168.1.8]:5060
May 27 13:16:13 localhost opensips[131639]:              udp: 192.168.1.8 [192.168.1.8]:6070
May 27 13:16:13 localhost opensips[131639]:              tcp: 192.168.1.8 [192.168.1.8]:5060
May 27 13:16:13 localhost opensips[131639]:              tcp: 192.168.1.8 [192.168.1.8]:6070
May 27 13:16:13 localhost opensips[131639]:              bin: 192.168.1.8 [192.168.1.8]:5555
May 27 13:16:13 localhost opensips[131639]: Aliases:
May 27 13:16:13 localhost opensips[131642]: NOTICE:core:main: version: opensips 3.2.6 (x86_64/linux)
[...]

PR #2849, opened 2022-06-02
[db_oracle] Do not change asynch mode when connection lost.

GangZhuo:
**Summary**

The PR try closes https://github.com/OpenSIPS/opensips/issues/2846.

**Details**

According to the contents of the paragraph "Setting Blocking Modes" at [OCI Programming Basics](http://www.cis.famu.edu/support/10g/Oracle_Database_10g/doc/appdev.102/b14250/oci02bas.htm),
We must set OCI_ATTR_NONBLOCKING_MODE attribute only after OCISessionBegin() or OCILogon2() has been called. Otherwise, an error will be returned.

In my test, when the connection lost due to a network error,
change OCI_ATTR_NONBLOCKING_MODE attribute also cause an error,
the error is ORA-03126 in my test.

In the case, the function done_timelimit will always fail with ORA-03126 error, 
and the con->errhp will overwritten with the new error code, 
the error code is not considered as a connection loss,
and subsequent operations will always throw this error,
and the lost connection never reconnect.

**Solution**

In the done_timelimit function, don't call change_mode function when the db connection lost, so the function do not override the error, and return success, the caller can continue to check the error, and detect that whether need to reconnect the database.

### DIFF
--- a/modules/db_oracle/asynch.c
+++ b/modules/db_oracle/asynch.c
@@ -189,9 +189,9 @@ sword begin_timelimit(ora_con_t* con, int connect)
 }
 
 
-static sword remap_status(ora_con_t* con, sword status)
+static sword remap_status(ora_con_t* con, sword status, sword *errcode)
 {
-	sword code;
+	sword code = 0;
 
 	if (   status == OCI_ERROR
 	    && OCIErrorGet(con->errhp, 1, NULL, &code,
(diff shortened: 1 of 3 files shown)


liviuchircu:
Hey, @GangZhuo!  And sorry for the delay on this one - if you could re-sync your `ORA-03126` branch against `master`, so we can get this PR re-opened and merged, that would be great!  Or, you could just open a new one (same thing).

The fix makes sense to me and it's also well-tested on your Oracle DB setup, so we should merge & backport it ASAP!

PR #2855, opened 2022-06-10
Fix in-dialog ACK being relayed with a wrong CSEQ

### LINKED ISSUE #1071: Wrong Cseq on ACK during race with re-Invite
I believe this is related to issue #1034

If a re-Invite is received at the same time as the ACK for the initial call setup the ACK gets forwarded with the Cseq of the re-Invite and not the initial Invite. Since the ACK has the wrong Cseq it may cause the call to be dropped/ended, so it is a big inter-op issue.

I haven't seen this affects overlapping requests later in the dialog, but I am not 100% sure it doesn't.

Seems like there may be some underlying Cseq management issues in the dialog module. I don't see (at least, haven't caught) this issue on my 1.11 systems, just 2.2.

Attached is an example capture.

[Overlap_Invite_ACK_Cseq.zip](https://github.com/OpenSIPS/opensips/files/831929/Overlap_Invite_ACK_Cseq.zip)

Opensips Version:
version: opensips 2.2.3 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, F_MALLOC, FAST_LOCK-FUTEX-ADAPTIVE_WAIT
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
git revision: 52813f4
main.c compiled on 16:33:40 Mar  7 2017 with gcc 5.2.1

dsandras:
**Summary**
Fixes wrong CSEQ on ACK requests

**Details**
There are cases where opensips can handle a reINVITE before the ACK that
acknowledges the previous INVITE/200 OK exchange.

In that case, the ACK would be relayed with the reINVITE CSEQ, leading
to interoperability issues.

**Solution**
We now never use the generated invite cseq for ACK requests. This change
also works with reINVITE pings.

It seems the problem was introduced in
27b0b21b2f2aa7b29fdaf0b7bc35274979fc55b0

**Closing issues**
Closes #1071

**Compatibility**

### DIFF
--- a/modules/dialog/dlg_handlers.c
+++ b/modules/dialog/dlg_handlers.c
@@ -2129,12 +2129,8 @@ after_unlock5:
 
 			dlg_lock (d_table, d_entry);
 
-			if (dlg->legs[dst_leg].last_gen_cseq ||
-			dlg->legs[dst_leg].last_inv_gen_cseq ) {
-				if (dlg->legs[dst_leg].last_inv_gen_cseq)
-					update_val = dlg->legs[dst_leg].last_inv_gen_cseq;
-				else
-					update_val = dlg->legs[dst_leg].last_gen_cseq;
+			if (dlg->legs[dst_leg].last_gen_cseq) {
+				update_val = dlg->legs[dst_leg].last_gen_cseq;
 				dlg_unlock( d_table, d_entry );
 
 				if (update_msg_cseq(req,0,update_val) != 0) {


dsandras:
It is worth double-checking it does not break anything, and especially https://github.com/OpenSIPS/opensips/commit/27b0b21b2f2aa7b29fdaf0b7bc35274979fc55b0

razvancrainea:
I don't think you fix is complete, as it may break scenarios where there are multiple requests generated from the proxy side. I am thinking of a different approach, where a mapping between INVITEs and their ACKs is used. I will update in #1071 .

PR #2868, opened 2022-07-15
presence: Add old-etag to event

hafkensite:
**Summary**
The events emitted from the presence module are missing the old_etag value of the presentity. 

**Details**
In order to track presentities fully the events should contain the old_etag. This allows the consumer to keep track of the old/new tag and properly update its state. Without the old_etag un-publishing a presentity can not be tracked.

**Solution**
Emit publish event after update is successful, and include the `old_etag` field in it.

Also fixed a few typos in the presence module comments.

**Compatibility**
I would think adding an extra field to the events does not break compatibility.

**Closing issues**

### DIFF
--- a/modules/presence/publish.c
+++ b/modules/presence/publish.c
@@ -66,6 +66,7 @@ void presence_raise_event(event_id_t event, presentity_t* presentity)
         static str parameter_eventname_str = { "event", 5 };
         static str parameter_expires_str = { "expires", 7 };
         static str parameter_etag_str = { "etag", 4 };
+        static str parameter_old_etag_str = { "old_etag", 8 };
         static str parameter_body_str = { "body", 4 };
         if (event == EVI_ERROR) {
                 LM_ERR("event not registered %d\n", presence_event_id);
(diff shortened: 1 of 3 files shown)


modules/presence/publish.c
```diff
@@ -616,6 +619,9 @@ int handle_publish(struct sip_msg* msg, str* sender_uri)
 		goto error;
 	}
 
+	/* send event E_PRESENCE_PUBLISH */
+	presence_raise_event(presence_event_id, &presentity);
+
```
bogdan-iancu, line 624: @hafkensite , is the event raising moved after the update_presentity() related to obtaining the old_etag info ?

hafkensite:
Indeed, that way that logic does not need to be duplicated. It als prevents sending events for publishes that are invalid for some reason.

On August 19, 2022 5:45:30 PM GMT+02:00, Bogdan Andrei IANCU ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -616,6 +619,9 @@ int handle_publish(struct sip_msg* msg, str* sender_uri)
> 		goto error;
> 	}
> 
>+	/* send event E_PRESENCE_PUBLISH */
>+	presence_raise_event(presence_event_id, &presentity);
>+
>
***@***.*** , is the event raising moved after the update_presentity() related to obtaining the old_etag info ?

Issue #2874, opened 2022-07-20
[BUG]Under extreme network conditions, the recvmsg function is blocked on Linux system calls for a long time and does not return


zbyslb:
**OpenSIPS version you are running**

version: opensips 3.2.6 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
main.c compiled on 00:37:02 Apr 21 2022 with gcc 8

**Describe the bug**

in opensips3.2.6, when the registered users reach about 300 and the number of concurrent calls is about 300, sometimes t_relay will be failed after receiving the BYE message, Once this scenario occurs, it will continue to appear until the system resources are fully occupied

**To Reproduce**

**Expected behavior**

**Relevant System Logs**

**OS/environment information**
 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

in the script, I added print information to the process of calling the function t_relay of each request message. in the process of forwarding the BYE message, there was no result returned after calling the function t_relay

Jul 20 11:01:59 [12552] DBG:tm:t_lookup_request: no transaction found
Jul 20 11:01:59 [12552] DBG:tm:run_reqin_callbacks: trans=0x7fee7c9d93f0, callback type 1, id 0 entered
Jul 20 11:01:59 [12552] DBG:core:mk_proxy: doing DNS lookup...
Jul 20 11:01:59 [12552] DBG:core:build_req_buf_from_sip_req: extra param added: <;rport>
Jul 20 11:01:59 [12552] DBG:core:parse_headers: flags=ffffffffffffffff
Jul 20 11:01:59 [12552] DBG:tm:set_timer: relative timeout is 500000
Jul 20 11:01:59 [12552] DBG:tm:insert_timer_unsafe: [4]: 0x7fee7c9d9610 (13171800000)
Jul 20 11:01:59 [12552] DBG:tm:insert_timer_unsafe: [0]: 0x7fee7c9d9640 (13181)
[...]

PR #2878, opened 2022-07-27
Under extreme network conditions, the recvmsg function is blocked on Linux system calls for a long time and does not return

zbyslb:
**Summary**

for #2874 
**Details**

**Solution**

**Compatibility**

**Closing issues**


zbyslb:
?

Issue #2891, opened 2022-08-15
[BUG] Call not terminated until FR_INV_TIMEOUT expires if t_wait_for_new_branches() was calling in main route


man1207:
**OpenSIPS version you are running**
version: opensips 3.2.5 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
main.c compiled on  with gcc 9

**Describe the bug**
I call t_wait_for_new_branches() in main route by example from https://opensips.org/html/docs/modules/3.2.x/event_routing.html#idp5535408
Two devices register their contacts and OS injects two branches
Then one device reject call by 486 BUSY.
I call t_wait_no_more_branches() and then call t_cancel_branch("a") in ONREPLY_ROUTE
But caller still waiting until FR_INV_TIMEOUT expires. 

**To Reproduce**
1. Add Push Notification EBR functionality to script form https://opensips.org/html/docs/modules/3.2.x/event_routing.html#idp5535408
2. Make call
3. Register two contacts to create two branches
4. Reject call by first device with 486 BUSY
5. Call t_wait_no_more_branches() in ONREPLY_ROUTE and then cancell second branch by t_cancel_branch("a")
6. Call at all will not terminated but should.

**Expected behavior**
Call should terminated if t_wait_no_more_branches() was called and cancelled all pending/ongoing non-2xx branches.

bogdan-iancu:
Same discussion on the mailing list....

bogdan-iancu:
Let me review the report and the provided patch

bogdan-iancu:
Thank you @man1207 for the all the details and explanations here - very useful and helpful.
I agree with your solution, still I do not feel comfortable with moving `t_wait_no_more_branches()` AFTER the `t_forward_nonack()`, as this may open the gates for signaling races - having the max branch set after the INVITE is out and maybe a reply back !
So, please check the attached patch with a slightly different approach on the fix, but same main idea.
[...]

PR #2898, opened 2022-08-24
fix t_wait_no_more_branches behaviour

man1207:
1. Issue: https://github.com/OpenSIPS/opensips/issues/2891
2. Logic:
a) t_wait_no_more_branches(): max number of allowed branches is set to the current number of branches
b) but really it add +1 to the max number of allowed branches. This leads to that OS consider that we are waiting for one more branch. Such logic comes from t_inject_branch() internal logic. First, OS calles t_wait_no_more_branches() inside t_inject_branch() and add +1 to max allowed branches before branch will be injected.
It works right for t_inject_branch() logic. But when we call "naked" t_wait_no_more_branches() in route script - OS adds +1 branch to max allowed, so we can never reach one branch yet. 
c) in my fix t_wait_no_more_branches(): max number of allowed branches is set to the current number of branches (but not add +1 to max allowed branches
d) reffering to "c", we need to change logic in t_inject_branches()
e) in my fix, first we create branch inside t_inject_branches() and only if branch was created clear - call t_wait_no_more_branches(). So OS set number of allowed branches to the current number of branches properly

**Details**
Issue: https://github.com/OpenSIPS/opensips/issues/2891
1. Problem. In Push Notification scenario (and any other), when we call t_wait_for_new_branches() in route script, we are able to stop wait for branches by calling t_wait_no_more_branches in ONREPLY_ROUTE, FAILURE_ROUTE etc as mentioned in docs.
It means that when t_wait_no_more_branches() was calling and all branches got 4XX status, call will be terminated.
But in real it does'nt happen due to the wrong logic in t_wait_no_more_branches() internal code. So call is active until FR_INV_TIMEOT expires.
2. Fix. Now t_wait_no_more_branches() stops waiting for new branches correctly. And call will be terminated if all branches will reach 4XX status.

**Solution**
1. PR fix properly max allowed branches when calling t_wait_no_more_branches()
2. PR corrects t_inject_branches() according to above item.

**Closing issues**
Issue: https://github.com/OpenSIPS/opensips/issues/2891

**Summary**
PR will fix behaviour of t_wait_no_more_branches() calling inside route script.
OS will set number of allowed branches to the current number of branches properly.
When all branches will reach 4XX status, call will be terminated. (In current implementation OS is still waiting for one more branch)

### DIFF
--- a/modules/tm/t_fwd.c
+++ b/modules/tm/t_fwd.c
@@ -1029,7 +1029,7 @@ int t_wait_no_more_branches( struct cell *t)
 	 * number of branches) */
 	for ( b=t->nr_of_outgoings-1; b>=t->first_branch ; b-- ) {
 		if (t->uac[b].flags & T_UAC_IS_PHONY) {
-			t->uac[b].br_flags=t->nr_of_outgoings+1;
+			t->uac[b].br_flags=t->nr_of_outgoings;
 			return 0;
 		}
 	}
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Alternative proposed in #2891

PR #2914, opened 2022-09-28
Corrected 'getting contacts' log

MayamaTakeshi:
**Summary**
Hi this is a small correction in a log line in lib/reg/lookup.c

The line intends to print the current AOR being processed but it is actually using the old AOR that was already processed.

**Details**
I detected this issue originally in opensips 3.2 commit b243666098be44226ade6a7df2b62851efcb5de8 as it was causing confusion to understand debug logs.


bogdan-iancu:
nice catch @MayamaTakeshi !

PR #2915, opened 2022-09-28
dialog: try to get a options/reinvite last direction

lemenkov:
**Summary**
This patch adds extra details in case of SIP ping failures within dialog.

**Details**
This patch improves visibility of call disconnection reason/direction in case of SIP ping failures within dialog. Before we lack information about the party which caused timeout ad now we know it. I've added one extra if(..) clause for some unknown case.

Tested in a real environment.

### DIFF
--- a/modules/dialog/dlg_timer.c
+++ b/modules/dialog/dlg_timer.c
@@ -840,7 +840,17 @@ void dlg_options_routine(unsigned int ticks , void * attr)
 		shm_free(it);
 		it = curr;
 
-		init_dlg_term_reason(dlg,"Ping Timeout",sizeof("Ping Timeout")-1);
+		if (dlg->legs[DLG_CALLER_LEG].reply_received == DLG_PING_FAIL)
+			init_dlg_term_reason(dlg,"Downstream Ping Timeout",sizeof("Downstream Ping Timeout")-1);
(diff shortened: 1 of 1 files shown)


modules/dialog/dlg_timer.c
```diff
@@ -840,7 +840,17 @@ void dlg_options_routine(unsigned int ticks , void * attr)
 		shm_free(it);
 		it = curr;
 
-		init_dlg_term_reason(dlg,"Ping Timeout",sizeof("Ping Timeout")-1);
+		if (dlg->legs[DLG_CALLER_LEG].reply_received == DLG_PING_FAIL)
+			init_dlg_term_reason(dlg,"Downstream Ping Timeout",sizeof("Downstream Ping Timeout")-1);
```
bogdan-iancu, line 844: @lemenkov , in the dialog module, the convention is the call flows from caller to callee, so caller -> callee is a downstream direction and callee -> caller is upstream direction ; sew https://opensips.org/html/docs/modules/3.2.x/dialog.html#pv_DLG_dir
So if the pinging failed on the caller side, it should be `Upstream Ping Timeout`, so the other way around :)
Or, if you want to keep it simple and 100% clear, better use `Caller Ping Timeout` .

PR #2919, opened 2022-10-01
EVI: Event subscribers expiry enhancements

goharahmed:
**Summary**

Following issues or rather minor inconvenience were faced when printing "subscribers_list" from MI modules.

1 - The subscribers of any event could only print their static time of subscription i.e 3600 and from output of MI command users can't tell when the subscription will actually expire. or how long this subscription is valid for.
2 - The "expired" subscription to an event continue to print their subscription data with original values when "subscriber_list" MI command is executed.
3 - The expired subscription to an event was ONLY removed when that event is raised. Until the event is raised the MI command keep printing the possibly expired subscription which is incorrect information to the users.

**Details**

Writing an external monitor of opensips Events Interface which can create new subscriptions to events when they are expired or not found, The outputs of the MI command "opensips-cli -x mi subscribers_list" continued to give wrong information which made if impossible to judge when an event subscription will expire. Hence I present this PR with following results:

1 - EVI: prints the subscriber expiry as a countdown value relevant to the subscriber's original expiry time.
2 - EVI: marks the subscriber as "expired" if it goes beyond its expiry time
3 - EVI: remove the expired subscribers from the list if MI command "subscribers_list" is executed.

**Solution**

1 - Create a function `evi_remove_expired_subs` that takes the event-id and parses all its subscribes; for each subscriber its subscription_time is added with expiry time and compared with current time. 
2 - if the time difference goes negative then the subscriber is removed from the list and subs list is assigned back to the event.
3 - Before printing the subscribers of an event, the event-id is passed to this new function to refresh its subscribers list.

**Compatibility**

Tested with opensips 3.3.1
```
version: opensips 3.3.1 (x86_64/linux)
flags: STATS: On, EXTRA_DEBUG, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, CC_O0, FAST_LOCK-ADAPTIVE_WAIT
git revision: d187a0bd5
main.c compiled on 18:01:16 Sep 20 2022 with gcc 8
```
**Closing issues**

### DIFF
--- a/evi/event_interface.c
+++ b/evi/event_interface.c
@@ -578,18 +578,25 @@ static int evi_print_subscriber(mi_item_t *subs_obj, evi_subs_p subs)
 			socket.len, socket.s) < 0)
 		return -1;
 	if (sock->flags & EVI_EXPIRE) {
+		/* indicate the expiration time used for the event*/
+		if (add_mi_number(subs_obj, MI_SSTR("expire"), sock->expire) < 0)
+				return -1;
 		now = time(0);
 		/* calculate the remaining time for the subscriber to be expired */
 		expiry_countdown = sock->expire - (now - subs->reply_sock->subscription_time);
 		if (expiry_countdown > 0) {
-			if (add_mi_number(subs_obj, MI_SSTR("expire"), expiry_countdown) < 0)
+			if (add_mi_number(subs_obj, MI_SSTR("ttl"), expiry_countdown) < 0)
+				return -1;
+		}else{
+			/* Mark event as expired if time-to-expire is reached*/
+			if (add_mi_string(subs_obj, MI_SSTR("ttl"), MI_SSTR("expired")) < 0)
 				return -1;
 		}
-		/* Mark event as expired if time-to-expire is reached*/
-		if (add_mi_string(subs_obj, MI_SSTR("expire"), MI_SSTR("expired")) < 0)
-			return -1;
 	} else {
 		if (add_mi_string(subs_obj, MI_SSTR("expire"), MI_SSTR("never")) < 0)
+			return -1;
+		/* If an event is never expired then its timeout should be never as well*/
+		if (add_mi_string(subs_obj, MI_SSTR("ttl"), MI_SSTR("never")) < 0)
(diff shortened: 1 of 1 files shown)


evi/event_interface.c
```diff
@@ -533,9 +577,16 @@ static int evi_print_subscriber(mi_item_t *subs_obj, evi_subs_p subs)
...
-
 	if (sock->flags & EVI_EXPIRE) {
-		if (add_mi_number(subs_obj, MI_SSTR("expire"), sock->expire) < 0)
+		now = time(0);
+		/* calculate the remaining time for the subscriber to be expired */
+		expiry_countdown = sock->expire - (now - subs->reply_sock->subscription_time);
+		if (expiry_countdown > 0) {
+			if (add_mi_number(subs_obj, MI_SSTR("expire"), expiry_countdown) < 0)
+				return -1;
+		}
```
razvancrainea, line 587: I believe an `else` is required here.
Moreover, to preserve backwards compatibility, I'd suggest to leave the `expire` node as it is, to indicate the expiration time used for the event, and add a new node, let's call it `timeout` or something similar, which contains the number of seconds left.

goharahmed: @razvancrainea while adding the `timeout` node does it make sense to add the same node in the area where expire=never

```
	} else {
		if (add_mi_string(subs_obj, MI_SSTR("expire"), MI_SSTR("never")) < 0)
			return -1;
		/* If an event is never expired then its timeout should be never as well*/
		if (add_mi_string(subs_obj, MI_SSTR("timeout"), MI_SSTR("never")) < 0)
			return -1;
	}
```

razvancrainea: no, definitely not, only when it is still active.

razvancrainea: I see that `timeout` is still printed for `expire=never` - I don't think it is necessary. And instead of `timeout`, does `remaining` sound better? your call :).

razvancrainea (approved):

PR #2927, opened 2022-10-15
Fix for access due to low memory.

ihsinme:
**Summary**
If you check after a memory access, then you always have access from outside. with stricter control of memory boundaries, this can lead to an abnormal termination of the program.


liviuchircu:
Nice catch, @ihsinme!  Still, you would need `100` timer routes in the _opensips.cfg_ before the invalid memory access would start taking place, which could explain why very few (if any) users have ran into this crash so far.

PR #2929, opened 2022-10-17
Fix rollback of transfer in case of using of separate media server url

nikbyte:
This is important bugfix which I use in production more than 1.5 years.
if fixes the case when you do transfer and using b2b_bridge() command with [provmedia_uri] parameter.
without this patch if transfer failed then rollback from transfer is broken.


modules/b2b_logic/logic.c
```diff
@@ -2831,6 +2892,7 @@ int process_bridge_action(struct sip_msg* msg, b2bl_tuple_t* tuple,
 				LM_ERR("Failed to create new b2b entity\n");
 				goto error;
 			}
+			entity->force_send_socket = msg?(msg->force_send_socket?msg->force_send_socket:0):0;
```
rvlad-patrascu, line 2895: This line should be removed as there is no such field in the `b2bl_entity_id_t` structure currently.

rvlad-patrascu (changes requested):
This looks ok aside from the change I mentioned.

modules/b2b_logic/logic.c
nikbyte: Done. thanks for note. It's a part of another patch.

PR #2931, opened 2022-10-17
b2b: inherit sending socket when create new dialog after negative ACK

nikbyte:
This small fix allows to inherit sending socket when create new dialog using previous dialog data.

### DIFF
--- a/modules/b2b_entities/dlg.c
+++ b/modules/b2b_entities/dlg.c
@@ -3196,6 +3196,7 @@ dummy_reply:
 			new_dlg->prev = dlg->prev;
 			new_dlg->add_dlginfo = dlg->add_dlginfo;
 			new_dlg->last_method = dlg->last_method;
+			new_dlg->send_sock = dlg->send_sock;
 			new_dlg->tracer = dlg->tracer;
 
 //			dlg = b2b_search_htable(htable, hash_index, local_index);


nikbyte:
up

nikbyte:
up

nikbyte:
up

bogdan-iancu:
@nikbyte , trying to understand what this is doing (as per your description) - the new dialog I assume is a new client created in the B2B session. But what's the "old" dialog? is the dialog being replaced by the new one  (during a bridging) ?

nikbyte:
Hi @bogdan-iancu . It was hard to remember because time passed, but I did.

If use b2b as a some balancer under which is another SIP server and if that 2nd SIP server makes few outgoing legs, like.

1. INVITE to 1st direction, 503
2. INVITE to 2nd direction, 500
3. INVITE to 3rd direction, call established

And all these few calls goes through b2b, then this flow is the same dialog for b2b, but #2 and #3 forgets outgoing socket if you set it for 1st call. The patch allows to keep outgoing socket for all legs.

bogdan-iancu:
@nikbyte , if I understand correctly, you do multiple re-briges in B2B, and you want the following bridges / UACs to remember the socket that was use by the prev failing UAC ?

PR #2932, opened 2022-10-18
Exported t_lookup_callid

MayamaTakeshi:
**Summary**
Function t_lookup_callid is currently not available via load_tm()

**Details**
This PR just exposes the function in the struct tm_binds.
I built it and tested with a custom module and it works.


bogdan-iancu:
@MayamaTakeshi , the policy here is to avoid adding code/functions that are not actually used by the public project. If you opted to keep your module provide, consider keeping this patch private also.

PR #2942, opened 2022-10-27
Fix To: header and sockets for b2b transfer

nikbyte:
This patch fixes the case for transfers. When new leg for transfer created, it generates To: header using RURI of previous header. In case if RURI has some parameters, new To: also has these parameters. It breaks Microsoft Teams transfers. We need make To: header and skip RURI parameters to avoid of this. Also this patch helps to set sockets properly for transfer legs.


nikbyte:
up

nikbyte:
up

nikbyte:
It requires rework of the patch, let it be opened

PR #2953, opened 2022-11-16
[db_oracle] Do not change asynch mode when connection lost.

GangZhuo:
The PR fix issue https://github.com/OpenSIPS/opensips/issues/2846, and have a slight difference from PR (https://github.com/OpenSIPS/opensips/pull/2849)

The sessions may be ended due to a network error, and change OCI_ATTR_NONBLOCKING_MODE attribute with no session will cause "ORA-03126: network driver does not support non-blocking operations" error.

When connection lost, we got "ORA-03135: connection lost contact" error, and then the function `done_timelimit` try to restore session to Non-Blocking mode by function `change_mode`, it's always failed and returned the true and overrided the error with ORA-03126. The subsequent db operations will always failed with an error ORA-03126, the error is not considered as connection loss (see function `db_oracle_errorinfo`), so the function `db_oracle_disconnect` never called, and also it is can not reconnect.

**Solution**

My solution is to modify the function `done_timelimit` function to close the connection instead of call `change_mode` function when the db connection lost, so the function `done_timelimit` return false and do not override the error, the caller can continue to check the raw error (ORA-03135), and detect that whether need to reconnect the database. Refer to https://github.com/OpenSIPS/opensips/blob/60c1d54f9193e8fca733eb1d0fe92fd60ffe7a27/modules/db_oracle/dbase.c#L299

After patched, the log like below:

```
Nov 16 11:35:31 xxxxxxx opensips[xxxxx]: ERROR:db_oracle:db_oracle_errorinfo: conneciom dropped
Nov 16 11:35:31 xxxxxxx opensips[xxxxx]: ERROR:db_oracle:db_oracle_submit_query: driver: ORA-03135: connection lost contact
                                             Process ID: 2xxxx
                                             Session ID: 5xxx Serial number: 4xxxx
Nov 16 11:35:31 xxxxxxx opensips[xxxxx]: NOTICE:db_oracle:db_oracle_submit_query: attempt repeat after reconnect
```
The opensips print the raw error, and attempt repeat after reconnect.

### DIFF
--- a/cachedb/cachedb_id.c
+++ b/cachedb/cachedb_id.c
@@ -44,6 +44,7 @@ static int dupl_string(char** dst, const char* begin, const char* end)
 
 	*dst = pkg_malloc(end - begin + 1);
 	if ((*dst) == NULL) {
+		LM_ERR("pkg malloc failed on %p/%p\n", begin, end);
 		return -1;
 	}
 
@@ -81,7 +82,7 @@ static int parse_cachedb_url(struct cachedb_id* id, const str* url)
 	};
 
 	enum state st;
-	unsigned int len, i, ipv6_flag=0;
+	unsigned int len, i, ipv6_flag=0, multi_hosts=0;
 	char* begin;
 	char* prev_token,*start_host=NULL,*start_prev=NULL,*ptr;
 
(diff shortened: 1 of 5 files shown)


liviuchircu:
@GangZhuo, I just manually downloaded & merged your branch, just to fix a small typo.  But it seems GitHub doesn't detect the change anymore, as the commit no longer matches against the real `GangZhuo:ORA-03126` branch.  I will also take care of backporting down to `3.1`.

Closing as completed (you are still the full author of the commit).  Many thanks!

liviuchircu:
Sorry for re-opening, I was just hoping the PR would be detected as "merged" automatically, now that the conflicts are solved.  [This blog](https://tighten.com/blog/adding-commits-to-a-pull-request/) shows how to properly do my commit next time from command-line, so your PR ends up on GitHub as "Merged" instead of "Closed"!

GangZhuo:
OK, thanks.

PR #2981, opened 2023-01-05
Fix flag usage in compression module

rrb3942:
The compression module seems to erroneously use the 'flags' field instead of the 'msg_flags' field for tracking if callbacks have already been set on the message.

The 'flags' field is for script level flags (setflag(), isflagset(), etc, script level functions).
This can lead to collisions with script level flags and unexpected behavior. As an example, calls to mc_compact began failing after I had introduced a new flag into my script which caused a collision with the flag used by the compression module.


liviuchircu:
Good catch, @rrb3942!  Cheers!

PR #3020, opened 2023-02-28
Fix bug in weight-based sort

nexbridge:
**Summary**
Quotient used in random number generation in weight_based_sort() function should always be less than 1 to avoid unexpected behaviour

**Type**
Bug fix

**Solution**
- Changed RAND_MAX to RAND_MAX + 1 so that the quotient will always be less than 1
- Changed casting from float to double to avoid rounding errors, particularly when the value of rand() nears RAND_MAX and thus the quotient could round to 1 unintentionally

**Compatibility**
The proposed solution is a strict improvement with no risk of side-effects

### DIFF
--- a/modules/drouting/drouting.c
+++ b/modules/drouting/drouting.c
@@ -2913,11 +2913,11 @@ static int weight_based_sort(pgw_list_t *pgwl, int size, unsigned short *idx)
 		}
 		if (weight_sum) {
 			/* randomly select number */
-			rand_no = (unsigned int)(weight_sum*((double)rand()/((double)RAND_MAX+1)));
+			rand_no = (unsigned int)(weight_sum*((double)rand()/((double)RAND_MAX)));
 			LM_DBG("random number is %d\n",rand_no);
 			/* select the element */
 			for( i=first ; i<size ; i++ )
-				if (running_sum[i]>rand_no) break;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Hi @nexbridge , thanks for your contribution. Indeed the missing `+1` may lead to `bug in weight sort` situations if the generated number `rand_no` is the same as `weight_sum`. But this may be fixed by changing the `if (running_sum[i]>rand_no)` condition to `>=` also. IMHO this will be more intuitive (basically shifting a bit the matching intervals for each destination to also include the its own running sum.
On the float versus double, agreed!

nexbridge:
Hi @bogdan-iancu,

Ah OK - should I close this then and leave you to apply your own bug fix?

Thanks

bogdan-iancu:
@nexbridge if you could update the PR to replace that `+1` with change in the test, it will be super, I will accept it.

nexbridge:
@bogdan-iancu OK I've removed the `+1` now and changed the `>` to `>=` as suggested, although I haven't checked your logic on that one so am relying on you! 🙂

bogdan-iancu:
perfect, thank you. Once the checks/tests will complete, I will merge it

bogdan-iancu:
Merged and backported, thank you @nexbridge

PR #3023, opened 2023-03-02
route: fix crash when get an integer parameter by $param(i)

GangZhuo:
When using $var(x)=$param(1) to obtain integer parameter values for route(xx, 3, ..), maybe duplicate a string from a fly pointer due to a type mismatch.

opensips.cfg

```
route {
    ...
    route(lb, 3);
}

route[lb] {
    $var(load_balancer_group) = $param(1);
    ...
}
```

Associated commit: https://github.com/OpenSIPS/opensips/commit/6b9e140f6576ba66be6f74e686a5fd33c0a16603

### DIFF
--- a/route.c
+++ b/route.c
@@ -862,7 +862,7 @@ static int eval_elem(struct expr* e, struct sip_msg* msg, pv_value_t *val)
 					{
 						if(val!=NULL)
 							memcpy(val, &lval, sizeof(pv_value_t));
-						if(lval.flags&PV_VAL_STR)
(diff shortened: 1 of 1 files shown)


GangZhuo:
> I am afraid your patch is not 100% complete. But could you please check if [16f5212](https://github.com/OpenSIPS/opensips/commit/16f5212d865fd75ca6f519381358e031ca9a10e2) fixes your issue?

@razvancrainea Yes, it is fixed.

PR #3033, opened 2023-03-13
Don't increase tcp_workers_no above tcp_workers_max_no (fixes segfault in debug mode)

jes:
**Summary**

This pull request fixes segfaults in debug mode caused by trying to start more TCP workers than the amount that memory was allocated for, when the config file contains `tcp_workers=1`.

(I also included fixes for 2 typos in comments & 1 slightly erroneous error message).

**Details**

When running in debug mode, `main()` checks if there are to be any TCP processes, and if so, it sets `tcp_workers_no` to 2, even if `tcp_workers_no` was previously only 1. This happens after `tcp_init()` has run, which has already allocated memory based on `tcp_workers_max_no`. Increasing `tcp_workers_no` above `tcp_workers_max_no` causes buffer overruns leading to segfaults.

**Solution**

I have changed the code in `main()` so that it can only decrease, and never increase, `tcp_workers_no`, so that it is no longer at risk of exceeding `tcp_workers_max_no`. I have also added a safety check in the start of `tcp_start_processes()`, comparable to the one in `fork_dynamic_tcp_processes()`, that calls `LM_BUG()` if it tries to start more processes than the `tcp_workers_max_no`.

(It is possible that there is a comparable bug related to `main()` in debug mode increasing `udp_workers_no` in the same manner, although there is no corresponding `udp_workers_max_no` and I didn't look any harder than that.)

**Compatibility**

I don't expect this change breaks any scenarios.

**Closing issues**

### DIFF
--- a/main.c
+++ b/main.c
@@ -764,7 +764,7 @@ try_again:
 			}
 		}
 		if (tcp_count_processes(NULL)!=0) {
-			if (tcp_workers_no!=2) {
+			if (tcp_workers_no>2) {
 				LM_NOTICE("setting TCP children to 2 (found %d)\n",
 					tcp_workers_no);
 				tcp_workers_no = 2;
--- a/main.c
+++ b/main.c
@@ -342,6 +342,7 @@ int main(int argc, char** argv)
 	int ret;
 	unsigned int seed;
 	int rfd;
+	int want_tcp_workers_no;
 
 	/*init*/
 	ret=-1;
@@ -764,10 +765,12 @@ try_again:
 			}
 		}
 		if (tcp_count_processes(NULL)!=0) {
-			if (tcp_workers_no>2) {
-				LM_NOTICE("setting TCP children to 2 (found %d)\n",
+			want_tcp_workers_no = (tcp_workers_max_no>=2) ? 2 : tcp_workers_max_no;
+			if (tcp_workers_no != want_tcp_workers_no) {
+				LM_NOTICE("setting TCP children to %d (found %d)\n",
+					want_tcp_workers_no,
 					tcp_workers_no);
-				tcp_workers_no = 2;
+				tcp_workers_no = want_tcp_workers_no;
 			}
 		}
 
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -292,7 +292,7 @@ static int mod_init(void)
 		register_timer("uac_reg_check", timer_check, (void*)(long)param, _timer,
 			TIMER_FLAG_DELAY_ON_DELAY);
 	} else {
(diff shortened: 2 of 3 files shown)


bogdan-iancu:
Hi @jes , thank you for your contribution here.

While your finding and fix is correct, I was wondering if maybe a better logical fix will be, in debug mode, to set `tcp_workers_no` to max between 2 and `tcp_workers_max_no`

jes:
I'm not actually sure what that logic is for. I guess the intention is to limit the amount of processes you have to contend with when debugging, but I don't know why you would want the number of TCP workers set to 2 instead of 1. 

Maybe it thinks the TCP main process is included in that count so the minimum possible value is 2?

Maybe it is just supposed to say 1 instead of 2?

Or maybe it is trying to make sure there are at least 2 workers so that you have a chance of encountering race conditions? (In which case it needs to do its thing before `tcp_init()` so that `tcp_workers_max_no` can be set correctly).

I'm not fussy about how it's fixed, whatever you think is best is probably best.

bogdan-iancu:
Yes, while enabling the debug_mode, the idea is to limit the number of procs and increase verbosity in order to make opensips easier to troubleshoot. There is no magic behind 2, just a low number of workers, but still higher than 1 :).
So, my idea is, when lowering the number of tpc procs in debug mode, to be sure we are still under the configured `tcp_workers` - so the new tcp_workers to be set to the mininum between 2 and already set tcp_workers.

BTW, wondering if not the same situation with UDP workers also.....

jes:
Yes I wondered if the same applied to UDP but it looked different enough that I gave up trying to check - at any rate it wasn't causing me segfaults.

bogdan-iancu:
OK, for the TCP part, will you alter the PR according to my suggestion or should extract parts of it and make myself a commit ?

jes:
Isn't that what it already does?

bogdan-iancu:
I mean [this line](https://github.com/OpenSIPS/opensips/blob/master/main.c#L770) should be
```
tcp_workers_no = (tcp_workers_max_no>=2) ? 2 : tcp_workers_max_no ;
```

And remove the test you added (between tcp_workers_no and tcp_workers_max_no) as it be useless

jes:
How does the new commit look?

I've deleted the sanity check you suggested deleting, although personally I think it is a good idea to keep it, given the demonstrated risk of tcp_workers_no exceeding tcp_workers_max_no for surprising reasons.

Would you also propose deleting the test in `fork_dynamic_tcp_processes` - https://github.com/OpenSIPS/opensips/blob/master/net/net_tcp.c#L1924 ?

bogdan-iancu:
I manually merged it, see e84297e7138f4545608e4dade9d37b6aa27b1bfa
Thank you @jes

jes:
Thanks!

PR #3039, opened 2023-03-24
Add REUSE_PORT socket option

jes:
**Summary**
Add a REUSE_PORT option for TCP sockets.

**Details**

This allows outgoing TCP connections to reuse a listening port as a source port. This is useful for example when using "forced_socket" with the "uac_registrant" module, or when using "force_send_socket()". Without this change, "forced_socket" and "force_send_socket()" have no obvious effect for TCP sockets.

The behaviour of sockets that do not use the REUSE_PORT option is unchanged.

I also noticed some of the other socket options (ANYCAST, FRAG) are documented in uppercase but only implemented in lowercase, so I made these accept either uppercase or lowercase as well.

We found Kamailio's similar option ( https://www.kamailio.org/wiki/cookbooks/5.4.x/core#tcp_reuse_port ) useful in the past.

**Solution**

Adds another `enum si_flags` value called `SI_REUSEPORT`. When this flag is enabled, TCP sockets have the `SO_REUSEADDR` and `SO_REUSEPORT` sockopt's enabled, and no longer override the port number to 0 before calling `bind()`. When the SI_REUSEPORT flag is not enabled, the behaviour is the same as before.

Remaining concerns: maybe you don't like changing the API for `tcp_init_sock_opt` and `tcp_sync_connect_fd` (which both gained a parameter to take an `enum si_flags` argument). Another way to do it would be to make every place that *calls* `tcp_init_sock_opt` or `tcp_sync_connect_fd` check the flags and enable SO_REUSEADDR/SO_REUSEPORT if necessary. I think the way I have done it is neater, but I don't know how important API stability is for the modules.

**Compatibility**

I can't imagine any scenarios that this change breaks.

**Closing issues**

### DIFF
--- a/cfg.lex
+++ b/cfg.lex
@@ -328,8 +328,9 @@ DOT			\.
 CR			\n
 
 ANY		"any"
-ANYCAST "anycast"
-FRAG "frag"
+ANYCAST	("anycast"|"ANYCAST")
+FRAG	("frag"|"FRAG")
+REUSE_PORT	("reuse_port"|"REUSE_PORT")
 
 
 COM_LINE	#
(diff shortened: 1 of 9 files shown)


jes:
Is there anything I can do to make this easier to merge?

PR #3059, opened 2023-04-18
feat: add ds_inherit_dst_state

wangduanduan:
**Summary**

for example
step1: dispatcher table 

| id  | setid | destination | state |
| --- | --- | --- | --- |
| 1 | 1 | sip:1.2.3.4:5060 | 0

step2: run opensips
step3: opensips mi show the sip:1.2.3.4:5060 is active
step4: change the dist state to 1, from db table

| id  | setid | destination | state |
| --- | --- | --- | --- |
| 1 | 1 | sip:1.2.3.4:5060 | 1

step5: run opensips mi ds_reload
step6: opensips mi show dest sip:1.2.3.4:5060 is active, not disabled

**Details**

**Solution**

i add the param ds_inherit_dst_state(default is 1) to 0,  to make opensips not inherit the old data。

**Compatibility**

because ds_inherit_dst_state default is 1, so it not change old behave

**Closing issues**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ out
 log
 *.log
 *.html
+.vscode
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -1164,7 +1164,10 @@ int ds_reload_db(ds_partition_t *partition)
 	if (old_data) {
 		/* copy the state of the destinations from the old set
 		 * (for the matching ids) */
-		ds_inherit_state( old_data, new_data);
+		if (ds_inherit_dst_state) {
+			ds_inherit_state( old_data, new_data);
+		}
+		
 		ds_destroy_data_set( old_data );
 	}
 
--- a/modules/dispatcher/dispatch.h
+++ b/modules/dispatcher/dispatch.h
@@ -59,6 +59,7 @@
 
 
 extern int ds_persistent_state;
+extern int ds_inherit_dst_state;
 
 typedef struct _ds_dest
 {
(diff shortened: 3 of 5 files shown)


bogdan-iancu:
Hi @wangduanduan , the original idea (of preserving the state) was for a reload NOT to impact the in-memory state of the in-used destinations. Actually the DB state is used only to seed the in-memory / runtime state when opensips starts, nothing more. This is the reason why  the DB state is ignored during reload, when the destination persists as definition.
May I ask why do you want to discard the in-mem status and have it overwritten with the the DB one?

bogdan-iancu:
Why not directly using the `ds_set_state` MI cmd ? 
https://opensips.org/html/docs/modules/3.2.x/dispatcher.html#mi_ds_set_state

wangduanduan:
i know ds_set_state will work， but the web server don't know what fields of the dest be changed, maybe not the state field, so it only can reload the dispatcher.

if you change the state and some other attr, you should call ds_reload and ds_set_state.

If dispatcher have param to discard old dispatcher dest state, just one command ds_reload will refresh all data like opensips restart.

wangduanduan:
up

bogdan-iancu:
@wangduanduan , I see this more as a problem to solve at the web server ....

But, if to do something in OpenSIPS, rather than adding a modparam (which is not so flexible, and not easy to correlate with the MI function), maybe a better approach is to add an optional param to the `ds_reload` MI function, like the  if the state should not be preserve

wangduanduan:
i will give a try, thank you

PR #3060, opened 2023-04-18
uac_auth: fix qop-value for MD5 hashing

benceszigeti:
**Summary**

The digest authentication response incorrect when the qop value is set to `auth,auth-int` and the MD5 hashing algorithm is being used. This PR fixes that.

**Details**

According to the documentation, the `uac_auth` module supports both qop values of `auth` and `auth-int`, but when both values are presented, `auth` is preferred. (Source: https://opensips.org/html/docs/modules/3.2.x/uac_auth.html)

The qop value is being filled with `auth,auth-int` for hashing instead of `auth` as described in RFC2617. The parsing set the qop flag correctly, but keeps the original qop value and using that later for hashing which causing the issue.

**Solution**

Based on the parsing result, use the appropriate qop value (`auth` or `auth-int`) for MD5 hashing.

**Compatibility**

The issue has been present since e53751e (v3.2.0), and requires a backport to v3.2.x and v3.3.x

**Closing issues**

N/A.

### DIFF
--- a/modules/uac_auth/auth.c
+++ b/modules/uac_auth/auth.c
@@ -240,6 +240,7 @@ int do_uac_auth(str *msg_body, str *method, str *uri, struct uac_credential *crd
 	const struct digest_auth_calc *digest_calc;
 	str_const cnonce;
 	str_const nc;
+	str_const qop;
 
 	digest_calc = get_digest_calc(auth->algorithm);
 	if (digest_calc == NULL) {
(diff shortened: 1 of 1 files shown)


genesys-adam (approved):

davidtrihy-genesys (approved):

bogdan-iancu:
Thank you @benceszigeti , good catch here!

PR #3069, opened 2023-04-28
feat: add inherit_state to ds_reload

wangduanduan:
**Summary**

after close https://github.com/OpenSIPS/opensips/pull/3059, i create this new pr.

if you change the state and some other attr, you should call ds_reload and ds_set_state

...

i add a param to ds_reload, `inherit_state`,  the default value of inherit_state is y, This will be compatible with previous behaviors.

```json
{"partition":"default", "inherit_state":"n"}
```

**Details**

**Solution**

**Compatibility**

**Closing issues**


modules/dispatcher/dispatcher.c
```diff
@@ -1360,15 +1361,30 @@ mi_response_t *ds_mi_reload_1(const mi_params_t *params,
...
 
+	if (get_mi_string_param(params, "inherit_state", &inherit_state.s, &inherit_state.len) >= 0) {
+		LM_DBG("inherit_state is: %s \n", inherit_state.s);
+
+		if (inherit_state.s[0] == '0' || inherit_state.s[0] == 'n' || inherit_state.s[0] == 'N') {
+			is_inherit_state = INHERIT_STATE_NO;
+		}
+		else if (inherit_state.s[0] == '1' || inherit_state.s[0] == 'y' || inherit_state.s[0] == 'Y') {
+			is_inherit_state = INHERIT_STATE_YES;
+		}
```
bogdan-iancu, line 1378: Maybe you can add here a warning log saying that the `inherit_state` values was not recognized, so ignored.

.gitignore
bogdan-iancu, line 1: please remove this file from the PR

modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -1607,14 +1607,23 @@ opensips-cli -x mi ds_list
 		<itemizedlist>
 			<listitem><para>
 				<emphasis>partition</emphasis> (optional) - name of
-				the partition to be reloaded.
+				the partition to be reloaded. default partition is partition.
```
bogdan-iancu, line 1610: I think the default partition is actually named `default`

modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -1607,14 +1607,23 @@ opensips-cli -x mi ds_list
...
 			</para></listitem>
+
+			<listitem><para><emphasis>inherit_state</emphasis> (optional) : whether inherit old state of the destination , default is y. you must set partition, and then set inherit_state. </para>
```
bogdan-iancu, line 1613: Actually it is not necessary to provide the name of the partition when using this new param. To avoid ambiguity of multiple optional params, you can use named params like this:
```
opensips-cli -x mi ds_reload inherit_state=n
```

bogdan-iancu:
Thank you @wangduanduan , please check my comments

wangduanduan:
@bogdan-iancu please check the update

modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -1607,14 +1607,23 @@ opensips-cli -x mi ds_list
 		<itemizedlist>
 			<listitem><para>
 				<emphasis>partition</emphasis> (optional) - name of
-				the partition to be reloaded.
+				the partition to be reloaded. default partition is default.
```
bogdan-iancu, line 1610: Change with : `Default partition is "default"`

modules/dispatcher/dispatcher.c
```diff
@@ -1360,15 +1362,34 @@ mi_response_t *ds_mi_reload_1(const mi_params_t *params,
 {
 	ds_partition_t *partition;
 	str partname;
+	str inherit_state;
+	int is_inherit_state = INHERIT_STATE_YES;
 
-	if (get_mi_string_param(params, "partition", &partname.s, &partname.len) < 0)
-		return init_mi_param_error();
+	if (get_mi_string_param(params, "partition", &partname.s, &partname.len) < 0) {
```
bogdan-iancu, line 1368: Not sure if this was in the original PR, but it is not a good idea to fall back to "default" on error here, if may have unpredictable side effects, please take this change out

wangduanduan:
@bogdan-iancu please check the udpate

bogdan-iancu:
oh, something that missed my big time - as it is a feature, it must be against master, not 3.2 - please re-generated

wangduanduan:
> oh, something that missed my big time - as it is a feature, it must be against master, not 3.2 - please re-generated

I think there should be some documentation in this project on how to contribute to it.

wangduanduan:
> oh, something that missed my big time - as it is a feature, it must be against master, not 3.2 - please re-generated

i think may be some link

PR #3070, opened 2023-04-29
Fix: regex segfault if max_groups is exceeded

NormB:
**Summary**

Fix: regex segfault if max_groups is exceeded

**Details**

The example provided in the doc has a regex_group file that contains 3 regex groups.
If the max_groups parameter is set to 2 a segfault happens as shown here:

Apr 29 14:01:38 [1] INFO:regex:mod_init: initializing module...
Apr 29 14:01:38 [1] NOTICE:regex:mod_init: loading pcres...
Apr 29 14:01:38 [1] ERROR:regex:load_pcres: max patterns exceeded
Apr 29 14:01:38 [1] CRITICAL:regex:mod_init: failed to load pcres
Apr 29 14:01:38 [1] CRITICAL:core:qm_free_dbg: freeing already freed pointer, first free: regex_mod.c: free_shared_memory(526) - aborting

**Solution**

The variable "i" is used to count the number of processed groups and it is compared against the max_groups variable.
The problem is that at the end of the loop that processes each group the "num_pcres_tmp" variable is set to "i + 1".
The "num_pcres_tmp" variable now has a value which is too large.

The fix is to decrement the variable "i" if the max_groups check fails.

If the max_groups check failed, all processing of the regex module would stop because of the "goto err" statement.  This has been changed so that any already processed regex groups are used and the module is allowed to continue.

Logging in this case has been enhanced to more clearly indicate that max_groups has been exceeded and that subsequent groups will not be processed.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/regex/regex_mod.c
+++ b/modules/regex/regex_mod.c
@@ -326,7 +326,7 @@ static int load_pcres(int action)
 			i++;
 			/* Check if there are more patterns than the max value */
 			if (i >= max_groups) {
-        i--;
+				i--;
 				LM_ERR("max_groups: %d exceeded, continuing without processing subsequent groups\n", max_groups);
 				break;
 			}
--- a/modules/regex/regex_mod.c
+++ b/modules/regex/regex_mod.c
@@ -327,7 +327,7 @@ static int load_pcres(int action)
 			/* Check if there are more patterns than the max value */
 			if (i >= max_groups) {
         i--;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thanks @NormB for the fix. To keep backward compatibility I suggest to have in this fix only the index decremented, while keeping the "goto err" behavior. 
As a second step, you can submit another PR for head only, to change the error behavior and skip the exceeding / bogus records - but this must be done in a consistent way, there are some following checks in the code  (like line 341) that may require the same change.

PR #3076, opened 2023-05-04
feat: add inherit_state to ds_reload

wangduanduan:
Summary

after close https://github.com/OpenSIPS/opensips/pull/3059, i create this new pr.

if you change the state and some other attr, you should call ds_reload and ds_set_state

...

i add a param to ds_reload, inherit_state, the default value of inherit_state is y, This will be compatible with previous behaviors.

```
opensips-cli -x mi ds_reload inherit_state=n
```


bogdan-iancu:
Thank you @wangduanduan

PR #3078, opened 2023-05-05
Optional verification of hostnames in TLS certificates

### LINKED ISSUE #3064: [BUG] OpenSIPS does not verify hostnames in TLS certificates (?)

```
version: opensips 3.4.0-dev (x86_64/linux)
git revision: 6faf77b
main.c compiled on 14:58:32 Mar 24 2023 with gcc 4.8.5
```

**Describe the bug**
When OpenSIPS connects to a TLS server that presents a certificate that does not match its hostname, OpenSIPS thinks the certificate passes validation and allows communication anyway.

There is probably a comparable bug regarding checking client certificates.

**To Reproduce**

1. Get OpenSIPS to connect using TLS to a server that has a good certificate (I did this using uac_registrant, but any method would be fine).
2. Verify that OpenSIPS successfully connects and sends SIP.
3. Now connect to exactly the same server but using a hostname that is not in the certificate (e.g. a bare IP address)
4. Observe that OpenSIPS *still* successfully connects and sends SIP, even though the certificate is not valid without a matching hostname.

**Expected behavior**

I expected OpenSIPS to reject a certificate when the common name (or subject alternate names) don't match the hostname it is trying to connect to.

**Relevant System Logs**

 - Operating System: CentOS 7
 - OpenSIPS installation: git
 - other relevant information:

**Additional context**

jes:
**Summary**

This PR adds a "verify_hostname" modparam to TLS domains in `tls_mgm`, which turns on OpenSSL verification of hostnames when connecting to TLS servers. The option is off by default, which leaves the default behaviour unchanged.

**Details**

Without verifying hostnames, it is almost completely pointless to verify TLS certificates at all*, because anyone can get a valid TLS certificate very easily (e.g. from LetsEncrypt). The point of verifying the hostname is so that you know you're speaking to the *correct peer*.

Prior to this PR, OpenSIPS **did not support verification of hostnames in TLS certificates**.

(* unless you are running a custom CA, in which case you can of course make sure that your CA only issues certificates to people you trust)

**Solution**

The main complication with this was how to let `tls_openssl` know what hostname it is trying to connect to. Mostly IP addresses were passed around inside OpenSIPS with `union sockaddr_union`s. I have turned this into `struct sockaddr_union_struct`, with an extra `char hostname[256]` field. This means when OpenSIPS is passing IP addresses around, they keep the hostname attached to them.

**Compatibility**

The main downside is that there is an increase in memory usage of 256 bytes per instance of `struct sockaddr_union_struct`.

I did not check that all of the modules compile, so it is likely that some modules that I'm not using do not compile any more, so this PR is probably not ready to merge straight away. Where a piece of code tries to access a field of `union sockaddr_union`, that needs to change to go through the struct (e.g. `foo.sin => foo.u.sin`). Nothing else needs to change.

If this PR is likely to be accepted, I am happy to go through all the modules and fix them.

**Closing issues**

Closes #3064.


jes:
This code still has the bug that OpenSIPS can reuse a connection to a given IP address even if the hostname is different. So if an attacker can cause your server to connect to mybadname.com which resolves to 1.2.3.4, and then your server tries to connect to mygoodname.com which also resolves to 1.2.3.4, it can send it through the existing connection which is encrypted with the mybadname.com certificate.

bogdan-iancu:
Thank you @jes for your work here. I would say we are at a (b) right now considering that

- the verification will be optional, so you can do it in both ways
- we need to check a bit deeper the implementation alternatives in terms of impacting as few code as possible (ideally avoiding changing the sock_addr union) - but here we need to allocate some time and evaluate alternatives.

But the idea is `Desirable`, let us evaluate the approach for the implementation

jes:
OK, thanks encouraging, thanks for the update @bogdan-iancu !

This evening I had a thought about an alternative way to do this that does not require modifying everywhere else that touches `sockaddr_union`.

The original `sockaddr_union` was:

    union sockaddr_union {
        struct sockaddr     s;
        struct sockaddr_in  sin;
        struct sockaddr_in6 sin6;
    };

So one option is to add an extra type to this union, whose size is as big as the largest of those (i.e. as big as a sockaddr_in6), plus the `char hostname[256]`.

One way to achieve this is:

    union sockaddr_union {
        struct sockaddr     s;
        struct sockaddr_in  sin;
        struct sockaddr_in6 sin6;
        struct {
            union {
                struct sockaddr     s;
                struct sockaddr_in  sin;
                struct sockaddr_in6 sin6;
            } u; /* this field is unused, but just serves to position the hostname field past the end of the sockaddr_in6 */
            char hostname[256];
        } h;
    };

So we'd add another member to the union which starts out with its *own* union that is the same as the other elements (just so that it ends up at the same size), plus it gains a `hostname` field. So to access the hostname of a `sockaddr_union` called `u`, it would just be `u.h.hostname`, and accessing the other fields is unchanged, and the type name is unchanged.

Or maybe this is more clear:

    /* this must have the same size as the largest field of `sockaddr_union` */
    union sockaddr_union_no_hostname {
        struct sockaddr     s;
        struct sockaddr_in  sin;
        struct sockaddr_in6 sin6;
    };

    union sockaddr_union {
        struct sockaddr     s;
        struct sockaddr_in  sin;
        struct sockaddr_in6 sin6;
        struct {
            char _padding[sizeof(union sockaddr_union_no_hostname)];
            char hostname[256];
        } h;
    };

What do you think about this sort of idea?

sobomax:
The idea is good, but the implementation is too broad IMHO. The core function that does translation between hostname and sockaddr_union is hostent2su() and it is only referenced in couple of places. The function can be extended to store reference to a hostname into a struct that wraps sockaddr_union only in places where it's necessary (i.e. in the TLS path) and pass if through the call chain only in those places, not in other 150 places where sockaddr_union happens to be used.

jes:
I didn't mean to close this; I've done the same again but now using the idea from this comment: https://github.com/OpenSIPS/opensips/pull/3078#issuecomment-1546111607

How are we looking now?

jes:
`hostent_shm_cpy` didn't previously copy the `h_name` field, which is needed with the TLS hostname verification code otherwise it segfaults if it tries to access the `h_name` field of a hostent that was populated by `hostent_shm_cpy`, so I've added that now.

It still doesn't copy `h_aliases` which could lead to similar surprises in the future.

Maybe we'd want `hostent_shm_cpy` to be exactly the same as `hostent_cpy` in all cases, except for the `shm`/`pkg` distinction?

sobomax:
> I agree that would be simpler in principle, the issue is it's not obvious to me how to work out which places need to be able to do TLS and which ones don't. I thought it was more straightforward to apply the change everywhere.
> 
> Similarly, I put the char[256] inside the struct, instead of a pointer, because that way you don't need to worry about when the string can be free'd.
> 
> But maybe it's easy, and I just don't know the codebase well enough?

Well, I have a conceptual problem with the approach you are taking: looks like you've started from the wrong end, i.e. you are pushing hostname into each and every name resolution the OpenSIPS performs, just so that you can consume that hostname at one or two very specific places. Having structure that can have different sizes is Bad Idea[tm[ as well, and is going to backfire in many places where sizeof() can be using for allocation, copying etc. You are just seeing some of it, much more is probably not visible just yet because we don't have any appreciable amount of code coverage, especially for code in in modules. And as evident from the V1 of the patch, it touches A LOT of that code.

And this is not the question of knowing the codebase well enough. In reality nobody probably does to a smaller or larger extent. 

I still don't see much need in showing in hostname into sockaddr_union. Instead I'd suggest approaching it from the other end: i.e. see where do you need that hostname and how you would modify call chain to save the hostname *ALONG* with the resolved IP just for that specific code flow. Then create union structure for that specific use case and push it in places where it really needs to be. 

struct sockaddr_hostname {
  struct sockaddr_union addr;
  struct str hostname;
};

Yes, that's probably slightly longer route, but at the end it might be actually faster way for you to get this to the mergeable state.

-Max

resolve.c
```diff
@@ -1881,8 +1881,18 @@ struct hostent* sip_resolvehost( str* name, unsigned short* port,
 	}
 
 	he = do_srv_lookup( tmp, port, dn);
```
sobomax, line 1883: @jes why don't we pass name->s in there directly (along with the length of the buffer, perhaps)? This way you won't need to memcpy() the result twice, first into tmp[] and then into the name->s.

sobomax:
@jes 

> I don't understand what you mean about structures that can be different sizes. Is it the sockaddr/sockaddr_in/sockaddr_in6 thing? That's just how sockaddr's work, it's already that way in the existing codebase. If you're talking about the way a hostent was copied with some fields missing if the destination is in shared memory, that's also already that way in the existing codebase. Otherwise, I'm not sure what you are referring to.
> 
> > save the hostname ALONG with the resolved IP just for that specific code flow
> 
> The "specific code flow" is every single place that could send a SIP packet, which is like everywhere. If the SIP packet turns out to be transmitted over TLS, you need the hostname attached to the sockaddr so that you can check the hostname on the certificate. I think it's more complicated and surprising (and more likely to be done incorrectly) if everything that wants to send a SIP packet has to work out whether it is using TLS or not and then behave differently. It makes more sense to me if we just tack the hostname on to the sockaddr in all cases.

@jes yes, you are right, sorry for confusion. I checked the latest version of the patch and it looks much saner now. I am still wondering if extra allocation and copying can somehow be only limited to the TCP targets only. Stacking hostname in all cases has obvious disadvantage of increasing CPU usage as well as increasing memory usage, including shared memory, each time sockaddr_union is used or allocated. In 99.99999% of real-use cases that extra information is not going to be used. How about adding a flag into sockaddr_union to indicate whether it has a name recorded, and avoid any extra work if it has not? Then you can check that flag and emit an error in your checking code if somehow that got misused somehow. Then you can turn it on in sip_resolvehost() based on the value of a configuration flag. It would still cost is some CPU usage on that branch, but since it's going to be a constant, the cost should be properly absorbed by the branch prediction logic.

Thanks!

-Max

sobomax:
OK, I've bitten the bullet and went to re-implement the proposed change by introducing a new type: "struct sockaddr_hu", which explicitly combines normal sockaddr_union with semi-optional hostname field and propagated it through the stack. Some work in progress is here: https://github.com/OpenSIPS/opensips/compare/master...sippy:opensips:pr_3078, any comments/suggestions are very welcome. Changes are big, but by no means unsurmountable.

jes:
@sobomax Nice one, thanks for looking at this.

I just added a commit that fixes a memory leak that I inadvertently introduced in https://github.com/OpenSIPS/opensips/pull/3078/commits/2efbf6439f88012112b8aa51089c9db79b6ceb97

The problem is that the `tm` module free's the hostent's fields manually instead of using `free_shm_hostent()`.

I don't know if you'd prefer to make `tm` use `free_shm_hostent()` rather than have it free the fields manually, but either way it needs to make sure `h_name` gets free'd if `h_name` is getting allocated.

jes:
Btw @sobomax I wonder if you could educate me - why is it that the shm hostent stuff doesn't allocate the name? Apparently this is the obvious behaviour, because `tm` doesn't bother freeing the name, and you prefer to put the name in the `h_addr_list` field instead of the `h_name` field. Could you help me understand?

sobomax:
> Btw @sobomax I wonder if you could educate me - why is it that the shm hostent stuff doesn't allocate the name? Apparently this is the obvious behaviour, because `tm` doesn't bother freeing the name, and you prefer to put the name in the `h_addr_list` field instead of the `h_name` field. Could you help me understand?

Well, that's kind of easy. In the hostent_shm_cpy() we allocate buffer sufficient to hold h_addr_list entries (plus some extra for a terminating NULL), so instead of allocating another shared memory chunk I am just allocating a *bigger* chunk to also have sufficient storage for the hostname and point h_name to the start of that section in that bigger chunk. 

So we are going from this:

```
                  +---shm_malloc#1----+
dst->h_addr_list->| h_addr_list[0..N] |
                  +-------------------+


                  +-shm_malloc#2-+
dst->h_name------>| h_name       |
                  +--------------+
```
To this:
```
                  +---shm_malloc#1----+
dst->h_addr_list->| h_addr_list[0..N] |
                  +-------------------+
dst->h_name------>| h_name       |
                  +--------------+
```
Not only it simplifies deallocation code, but is also going to improve cache locality as well as avoid shared memory fragmentation.

sobomax:
P.S. Lots of complexity of the current name resolution code steams from the unfortunate fact that OpenSIPS historically (ab)uses system-provided "struct hostent" to hold all this data and pass it between threads and that locks us into a certain structure. As such we have no control over the layout of that struct and also forced to use nil-terminated string instead of more safe and convenient str / str_const. I am really inclined to rip that off and roll our very own version of the said struct.

bogdan-iancu:
Hi @jes, where are we with this ? I see the patch is reasonable so far ... should we go for a review ?

jes:
Yes please!

bogdan-iancu:
@jes , while working out the review, I came some important questions.

1) the overall idea is to transport the "name" of the host we are connecting, so to be able to check later when performing the TLS handshake. And for this, the "name" goes to via (a) hostend ,  (b) sockaddr and finally (c) tcp conn. 
Question - this TLS verification will make sense only when opensips acting as TLS client right ?

2) if in the first place I thought the "name" is actually what the DNS lookup returns in `h_name` of the hostent struct, I see you actually grab the "name" we do DNS lookup over. And it is not something returned by DNS resolver. If my understanding is correct, there are some side issues here:

2.1) what if you go via NAPTR and SRV lookups? if the you do NAPTR over "foo.com" to discover that SIP service is provided via "sip.foo.com"...and of course, we will end up with a hostend struct (after SRV and AAA lookups for "sip.foo.com") which holds as name "foo.com" ; not sure how the certificate is expected to be generated (by the server side) as you can get to the same "sip.foo.com" via NAPTR over "foo.com" or direct AAA lookup over "sip.foo.com"... wouldn't one of these conflict with the certificate ? 

2.2) as the "name" is actually something we already have (as piece of info) (is it not returned by DNS lookup), isn't there a simpler/easier path to get it into the TCP conn, without touching all the intermediary structs (hostend & sockaddr) ?

For now, please ignore the ongoing review, let's try to clarify the above. Thanks.!

jes:
1. Yes, verifying the hostname is only applicable when connecting out as a client.

2.1 I don't actually know too much about SIP. In the HTTP world you make the `commonName`/`subjectAltName` fields say the hostname that end users programs will put in URIs. I assume the same is true in SIP.

2.2 I couldn't see a simpler path to get it into the TCP connection. At the time something tries to send a packet it is ambivalent as to whether it is going out over UDP or TCP or TLS. Would you propose doing it differently? Potentially everywhere that can send a packet needs to be able to send it over TLS.

bogdan-iancu:
Ok @jes , thanks for the feedback, let me resume the review

sobomax:
I agree with @bogdan-iancu, the proposed way to attach a name is still a bit too intrusive. I actually started a rewrite of the PR about a year ago, based off @jes code, but then ran out of time. Just pushed it for a reference here:

https://github.com/OpenSIPS/opensips/compare/master...sippy:opensips:wip-pr_3078

The main difference is that I started from opposite direction: i.e. defined an extended version `struct host_sock_info` an extended version of `union sockaddr_union` to be used in cases where name is relevant and started to pushing it up from TLS code.

Where I got stuck is the name resolution. That code is really hairy also it abuses system-defined structures, so some semi-major cleanup effort is needed there to make this all work.

PR #3084, opened 2023-05-08
Add options to set_dlg_profile and unset_dlg_profile to remove all values on a profile

rrb3942:
**Summary**
This adds the ability for set_dlg_profile() and unset_dlg_profile() to optionally remove all other values associated with a given dialog profile when called. In the case of set_dlg_profile() the removal happens before setting the new value.

**Details**
This makes working with profile values that you tend to overwrite much easier. You don't need to explicitly call unset_dlg_profile() for every previous value when setting new ones.

**Solution**
This:
unset_dlg_profile("caller",$var(oldfu));
set_dlg_profile("caller",$fu);

Becomes just:
set_dlg_profile("caller",$fu, 1);

You can also clear all values on a profile with:
unset_dlg_profile("caller",,1);

**Compatibility**
The new parameter is optional, and does not break prior usage.

We have been running this patch in production on 3.2 without issue.

### DIFF
--- a/modules/dialog/dialog.c
+++ b/modules/dialog/dialog.c
@@ -123,7 +123,7 @@ static int w_match_dialog(struct sip_msg *msg, void *seq_match_mode_val);
 static int api_match_dialog(struct sip_msg *msg, int _seq_match_mode);
 static int w_validate_dialog(struct sip_msg*);
 static int w_fix_route_dialog(struct sip_msg*);
-static int w_set_dlg_profile(struct sip_msg *msg, str *prof_name, str *value, str *remove_value);
+static int w_set_dlg_profile(struct sip_msg *msg, str *prof_name, str *value, int *clear_values);
 static int w_unset_dlg_profile(struct sip_msg *msg, str *prof_name, str *value);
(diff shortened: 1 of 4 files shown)


bogdan-iancu:
Wouldn't be a simpler approach to have only  `unset_dlg_profile("caller","")` meaning remove from the profile with all the values - the marker will be the empty string value?

rrb3942:
@bogdan-iancu 

Great suggestion. I extended it to set_dlg_profile as well, so you can specify a value to remove, or all, while setting a new one. Please see the revised patch.

This:
unset_dlg_profile("caller",$var(oldfu));
set_dlg_profile("caller",$fu);

Becomes:
set_dlg_profile("caller",$fu, $var(oldfu));

Or to clear all values while setting:
set_dlg_profile("caller",$fu,"");

You can also clear all values on a profile with:
unset_dlg_profile("caller","");

liviuchircu:
Hey, @rrb3942!  After looking into this more closely, we agreed that extending both the `set` and `unset` functions with the new parameter is ideal.  Also, the new param should be a boolean, per the initial PR.  Let me merge your latest branch locally, amend the functions one final time and merge it!

PR #3086, opened 2023-05-09
add inhert_state param for mi dr_reload

wangduanduan:
**Summary**

the command dr_reload can't reset the state of gateway , because it will inherit old state and ignore the new state in db.

if you change the gateway state, you should call dr_gw_status.

if you change the state and some other attr, you should call dr_reload and dr_gw_status

...

i add a param to dr_reload, inherit_state, the default value of inherit_state is y, This will be compatible with previous behaviors.

```
opensips-cli -x mi dr_reload inherit_state=n
```

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/drouting/drouting.c
+++ b/modules/drouting/drouting.c
@@ -2329,32 +2329,12 @@ static mi_response_t *mi_dr_get_partition(const mi_params_t *params,
 	return NULL;
 }
 
-static inline int get_inherit_state (const mi_params_t *params) {
-	str inherit_state;
-	int is_inherit_state = 1; // default is inhert old state
-
-	if (get_mi_string_param(params, "inherit_state", &inherit_state.s, &inherit_state.len) >= 0) {
-		LM_DBG("inherit_state is: %s \n", inherit_state.s);
-
-		if (inherit_state.s[0] == '0' || inherit_state.s[0] == 'n' || inherit_state.s[0] == 'N') {
-			is_inherit_state = 0;
-		}
-		else if (inherit_state.s[0] == '1' || inherit_state.s[0] == 'y' || inherit_state.s[0] == 'Y') {
-			is_inherit_state = 1;
-		} else {
-			LM_WARN("inherit_state values was not recognized, ignored, use default 1\n");
-		}
-	}
-
-	return is_inherit_state;
-}
-
 mi_response_t *dr_reload_cmd(const mi_params_t *params,
 								struct mi_handler *async_hdl)
 {
-	LM_INFO("dr_reload MI command received!\n");
+	int is_inherit_state = get_mi_bool_like_param(params, "inherit_state", 1);
 
-	int is_inherit_state = get_inherit_state(params);
+	LM_INFO("dr_reload MI command received!\n");
 
 	if (dr_reload_data(0, is_inherit_state) != 0) {
 		LM_CRIT("failed to load routing data\n");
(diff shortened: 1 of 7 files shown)


modules/drouting/drouting.c
```diff
@@ -1266,36 +1269,38 @@ static inline int dr_reload_data_head(struct head_db *hd,
...
+			/* copy the state of gw/cr from old data */
+			/* interate new gws and search them into old data */
+			for (map_first(new_data->pgw_tree, &it);
+					iterator_is_valid(&it); iterator_next(&it)) {
+				dest = iterator_val(&it);
+				if(dest==NULL)
+					break;
 
-			gw=(pgw_t *)*dest;
+				gw=(pgw_t *)*dest;
```
bogdan-iancu, line 1281: please check if the indentation is with spaces  and if so, replace with TAB  `\t`

modules/drouting/drouting.c
```diff
@@ -1266,36 +1269,38 @@ static inline int dr_reload_data_head(struct head_db *hd,
...
-				break;
+			/* interate new crs and search them into old data */
+			for (map_first(new_data->carriers_tree, &it);
+					iterator_is_valid(&it); iterator_next(&it)) {
+				dest = iterator_val(&it);
+				if(dest==NULL)
+					break;
 
-			cr=(pcr_t *)*dest;
+				cr=(pcr_t *)*dest;
```
bogdan-iancu, line 1296: also here, check the indentation please

modules/drouting/drouting.c
```diff
@@ -2324,12 +2329,34 @@ static mi_response_t *mi_dr_get_partition(const mi_params_t *params,
...
+	str inherit_state;
+	int is_inherit_state = 1; // default is inhert old state
+
+	if (get_mi_string_param(params, "inherit_state", &inherit_state.s, &inherit_state.len) >= 0) {
+		LM_DBG("inherit_state is: %s \n", inherit_state.s);
+
+		if (inherit_state.s[0] == '0' || inherit_state.s[0] == 'n' || inherit_state.s[0] == 'N') {
+			is_inherit_state = 0;
+		}
+		else if (inherit_state.s[0] == '1' || inherit_state.s[0] == 'y' || inherit_state.s[0] == 'Y') {
```
bogdan-iancu, line 2342: In these checks, you should also check the len of the param value to be sure it is 1. Otherwise the `yesterday` value will be considered valid and "do inherit" :)

modules/drouting/drouting.c
```diff
@@ -2324,12 +2329,34 @@ static mi_response_t *mi_dr_get_partition(const mi_params_t *params,
...
+			is_inherit_state = 1;
+		} else {
+			LM_WARN("inherit_state values was not recognized, ignored, use default 1\n");
+		}
+	}
+
+	return is_inherit_state;
+}
+
```
bogdan-iancu, line 2351: This is not necessarily related to this PR, but as this is 100% the same idea as in `dispatcher`, I would like to have some consistency, in terms of:

- using a function like `get_inherit_state` for evaluating the param value
- using enum or simply 0/1 values for the internal value of the inherit variable

modules/drouting/drouting.c
```diff
@@ -2324,12 +2329,34 @@ static mi_response_t *mi_dr_get_partition(const mi_params_t *params,
...
 mi_response_t *dr_reload_cmd(const mi_params_t *params,
 								struct mi_handler *async_hdl)
 {
 	LM_INFO("dr_reload MI command received!\n");
 
-	if (dr_reload_data(0) != 0) {
+	int is_inherit_state = get_inherit_state(params);
```
bogdan-iancu, line 2357: please do all var declaration in the very beginning of the function

bogdan-iancu (changes requested):

modules/drouting/drouting.c
wangduanduan: i checked, this is tab

modules/drouting/drouting.c
wangduanduan: this is tab

bogdan-iancu (approved):

bogdan-iancu:
Thank you @wangduanduan , great work !

PR #3087, opened 2023-05-09
Enhance module version check to also check the SCM ID

sobomax:
**Summary**

This change enhances module version check to also check the SCM ID (i.e. git hash) and refuse to load module if they differ.

**Details**

Often when the OpenSIPS internal APIs or structures change this also requires all modules to be recompiled. Otherwise the OpenSIPS would crash or misbehave in an unpredictable and very hard to debug way. It is believed to be the cause of failure reported here #3081 and possibly many other issues that cannot be reproduced.

**Solution**

We embed SCM type (i.e. "git") and SCM revision (i.e. "abcdef123") into the module_exports struct and compare them with respective values compiled into code once module is loaded. Error is reported and OpenSIPS exits with an error if an mismatch is found:

***Trying to load module compiled before this change***
```
$ ./opensips -FE -f o1.cfg
May  8 18:51:56 [2810577] DBG:core:add_module_dependency: adding type 1 dependency proto_tcp - (module proto_hep)
May  8 18:51:56 [2810577] DBG:core:load_module: loading module modules/signaling/signaling.so
May  8 18:51:56 [2810577] DBG:core:register_module: register_pv: signaling
May  8 18:51:56 [2810577] DBG:core:pv_add_extra: extra items list is not initialized
May  8 18:51:56 [2810577] DBG:core:add_module_dependency: adding type 1 dependency signaling - (module tm)
May  8 18:51:56 [2810577] DBG:core:add_module_dependency: adding type 1 dependency signaling - (module sl)
May  8 18:51:56 [2810577] DBG:core:load_module: loading module modules/sl/sl.so
May  8 18:51:56 [2810577] DBG:core:load_module: loading module modules/db_mysql/db_mysql.so
May  8 18:51:56 [2810577] CRITICAL:core:version_control: BUG - SCM type not defined in module <modules/db_mysql/db_mysql.so> (try `make clean all modules' and reinstall everything)
```

***Trying to load mis-matched revision***
```
$ ./opensips -FE -f o1.cfg
May  8 20:19:22 [2830822] DBG:core:add_module_dependency: adding type 1 dependency proto_tcp - (module proto_hep)
May  8 20:19:22 [2830822] DBG:core:load_module: loading module modules/signaling/signaling.so
May  8 20:19:22 [2830822] DBG:core:register_module: register_pv: signaling
May  8 20:19:22 [2830822] DBG:core:pv_add_extra: extra items list is not initialized
May  8 20:19:22 [2830822] DBG:core:add_module_dependency: adding type 1 dependency signaling - (module tm)
May  8 20:19:22 [2830822] DBG:core:add_module_dependency: adding type 1 dependency signaling - (module sl)
May  8 20:19:22 [2830822] DBG:core:load_module: loading module modules/sl/sl.so
May  8 20:19:22 [2830822] DBG:core:load_module: loading module modules/db_mysql/db_mysql.so
May  8 20:19:22 [2830822] ERROR:core:version_control: module SCM revision mismatch for db_mysql
core: fcad64a53
module: 54572a71f (try `make clean all modules' and reinstall everything)
```

### DIFF
--- a/sr_module.c
+++ b/sr_module.c
@@ -199,28 +199,51 @@ error:
 
 static inline int version_control(const struct module_exports* exp, char *path)
 {
-	if ( !exp->version ) {
+	const char *hint = "(try `make clean all modules' and reinstall everything)";
+	const char *scm_nm = "version control system";
+	if ( !exp->ver_info.version ) {
 		LM_CRIT("BUG - version not defined in module <%s>\n", path );
 		return 0;
 	}
-	if ( !exp->compile_flags ) {
+	if ( !exp->ver_info.compile_flags ) {
 		LM_CRIT("BUG - compile flags not defined in module <%s>\n", path );
 		return 0;
 	}
-
-	if (strcmp(OPENSIPS_FULL_VERSION, exp->version)==0){
-		if (strcmp(OPENSIPS_COMPILE_FLAGS, exp->compile_flags)==0)
-			return 1;
-		else {
-			LM_ERR("module compile flags mismatch for %s "
-				" \ncore: %s \nmodule: %s\n",
-				exp->name, OPENSIPS_COMPILE_FLAGS, exp->compile_flags);
-			return 0;
-		}
+	if ( !exp->ver_info.scm.type ) {
+		LM_CRIT("BUG - %s type not defined in module <%s> %s\n",
+			scm_nm, path, hint );
+		return 0;
 	}
-	LM_ERR("module version mismatch for %s; core: %s; module: %s\n",
-		exp->name, OPENSIPS_FULL_VERSION, exp->version );
-	return 0;
+	if ( !exp->ver_info.scm.rev ) {
+		LM_CRIT("BUG - %s revision not defined in module <%s> %s\n",
+			scm_nm, path, hint );
+		return 0;
+	}
+
+	if (strcmp(OPENSIPS_FULL_VERSION, exp->ver_info.version)!=0) {
+		LM_CRIT("module version mismatch for %s; core: %s; module: %s\n",
+			exp->name, OPENSIPS_FULL_VERSION, exp->ver_info.version );
+		return 0;
+	}
+	if (strcmp(OPENSIPS_COMPILE_FLAGS, exp->ver_info.compile_flags)!=0) {
+		LM_CRIT("module compile flags mismatch for %s "
+			" \ncore: %s \nmodule: %s\n", exp->name,
+			OPENSIPS_COMPILE_FLAGS, exp->ver_info.compile_flags);
+		return 0;
+	}
+	if (strcmp(core_scm_ver.type, exp->ver_info.scm.type) != 0) {
+		LM_CRIT("module %s type mismatch for %s "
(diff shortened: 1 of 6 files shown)


sr_module.c
```diff
@@ -199,28 +199,48 @@ int register_module(const struct module_exports* e, char* path, void* handle)
...
-		if (strcmp(OPENSIPS_COMPILE_FLAGS, exp->compile_flags)==0)
-			return 1;
-		else {
-			LM_ERR("module compile flags mismatch for %s "
-				" \ncore: %s \nmodule: %s\n",
-				exp->name, OPENSIPS_COMPILE_FLAGS, exp->compile_flags);
-			return 0;
-		}
+	if (strcmp(OPENSIPS_FULL_VERSION, exp->ver_info.version)!=0) {
+		LM_ERR("module version mismatch for %s; core: %s; module: %s\n",
```
bogdan-iancu, line 221: All these logs about mismatch should be CRIT (as it prevents the startup), rather then the regular ERR... imho

sr_module.c
```diff
@@ -199,28 +199,48 @@ int register_module(const struct module_exports* e, char* path, void* handle)
...
-		exp->name, OPENSIPS_FULL_VERSION, exp->version );
-	return 0;
+	if (strcmp(OPENSIPS_COMPILE_FLAGS, exp->ver_info.compile_flags)!=0) {
+		LM_ERR("module compile flags mismatch for %s "
+			" \ncore: %s \nmodule: %s\n",
+			exp->name, OPENSIPS_COMPILE_FLAGS, exp->ver_info.compile_flags);
+		return 0;
+	}
+	if (strcmp(core_scm_ver.type, exp->ver_info.scm.type) != 0) {
+		LM_ERR("module SCM type mismatch for %s "
```
bogdan-iancu, line 232: The term of `SCM` may be confusing to regular user (even for me, it was unknown by now). The idea of a log is to be clear and understandable by the user. So, let's try to explain / replace this SCM with something that can be understood by user - even like GIT revision, version, etc...something to ring a  :bell:

razvancrainea (commented):
Looks good to me!
My only feedback is that it would be nice to have some sort of mechanism to skip this check when you are actually intended to replace a module you know its interface was not modified. This could be either a compile flag (at the module level, to instruct the core to skip its check - this might be a bit dangerous), either a start parameter.

razvancrainea:
Giving a thought, perhaps an easier way to accomplish my goal would be to be able to force a specific VERSION of the module, overwriting the auto-detected one.

sr_module.c
sobomax: Thanks for reminding how old I am, Bogdan! 😂 Will do.

sobomax:
> Giving a thought, perhaps an easier way to accomplish my goal would be to be able to force a specific VERSION of the module, overwriting the auto-detected one.

Well, if it bothers you (or anyone else) too much then such feature will not take long to implement. 💯

sr_module.c
sobomax: Adjusted as suggested. Thanks, @bogdan-iancu!

PR #3103, opened 2023-05-25
Improve passthru_prack behaviour so it actually passes through PRACKs

Giedriusj1:
**Summary**

The b2b_entities module has a passthru_prack toggle, which in theory should cause PRACKs to behave in an end-to-end manner when set to 1.

Unfortunately the existing behaviour only does half of what's expected. It stops our b2b module from responding with PRACKs itself (upon seeing Requires: 100rel), but it has no logic to actually deal with PRACK replies when they eventually come back.

**Details**
This patch implements the missing behaviour so the PRACKs are truly end-to-end.

**Solution**

When a message requesting 100rel is identified, dlg is updated with the RAck header that the subsequent PRACK will have to use.

PRACKs are now correctly proxied with the appropriate RAck set to match the message which requested the 100rel.

b2b_dlg_t struct now has an additional prack_tran, which is used to keep track of the PRACK transaction, so we know how to deal with the 200 OK which comes for the PRACK.

**Compatibility**
Yes, it changes how passthru_prack behaves, but I don't think the existing behaviour is correct. Or at least it's very confusing, because it contradicts the docs.

**Closing issues**

### DIFF
--- a/modules/b2b_entities/dlg.c
+++ b/modules/b2b_entities/dlg.c
@@ -1030,23 +1030,30 @@ search_dialog:
 
 	if (method_value == METHOD_PRACK)
 	{
-		B2BE_LOCK_RELEASE(table, hash_index);
-		LM_DBG("Received a PRACK - send 200 reply\n");
-		str reason={"OK", 2};
-		/* send 200 OK and exit */
-		tmb.t_newtran( msg );
-		tm_tran = tmb.t_gett();
-		if (dlg)
-			b2b_run_tracer(dlg, msg, tm_tran);
-		tmb.t_reply(msg, 200, &reason);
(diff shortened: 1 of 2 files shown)


Giedriusj1:
It sems that the passthru_prack functionality was initially implemented by @carstenbock

It would be great if Carsten could also take a look at my changes and advise on any impact these could have.

carstenbock (approved):
Looks good to me - nothing to add. Thanks for the contribution!

carstenbock:
(Sorry for the delay - I was out and at some conferences)

gjonikas:
Brilliant, thanks for the review!

rvlad-patrascu (approved):

PR #3105, opened 2023-05-29
completely disabled state persists over opensips restart

wangduanduan:
**Summary**

Calling the lb_status MI command for disabling (on demand) the destination. the destination will be completly disabled。

but once opensips restart, the **completly disabed** state will be lost.

i expect the **completely disabled** will keep persists over opensips restart

**Details**

**Solution**

Add Probing mode  12 (completly disabled)

1. set the probing mode  to 12
2. call lb_reload

**Compatibility**

**Closing issues**


wangduanduan:
@bogdan-iancu thank you。

PR #3132, opened 2023-07-26
`proto_hep`: Add TLS transport

benceszigeti:
**Summary**

Send HEP messages over TLS.

**Details**

**Solution**

**Compatibility**

**Closing issues**


bogdan-iancu:
Thank you @benceszigeti for your contribution. On a first check it looks ok, but let;s also see how it works :)

PR #3154, opened 2023-08-15
The proxy address does not specify the protocol to give an error mess…

button-chen:
…age and is not added to the node list

**Summary**
The proxy address does not specify the protocol to give an error message and not added to the node list

**Details**
Configure the RTP Proxy address without the protocol part. you will see an error log: "ERROR:rtpengine:send_rtpe_command: proxy <ip:port> does not respond, disable it". but it is not easy to see what is the problem

**Solution**
The proxy address does not specify the protocol to give an error message and not added to the node list

**Compatibility**

**Closing issues**


razvancrainea:
This looks only like a documentation issue, specifying that if the protocol is missing, `unix:` will be considered. I don't think we should ignore the socket, as this is not backwards compatible. I will push the docs change and close this PR.
Thanks for documenting the issue!

PR #3165, opened 2023-08-22
Jes/uac registrant null pointer dereference

jes:
**Summary**
I was seeing segfaults when trying to make opensips register against a server with a much-too-short interval. I traced the problem back to a uac_registrant null pointer dereference of `msg->min_expires`. 

**Details**
This commit makes uac_registrant only try to dereference `msg->min_expires` if it is non-NULL.

**Solution**
I wrapped the dereference of `msg->min_expires` in `if (msg->min_expires)`.

**Compatibility**
It is possible that we would instead want `parse_min_expires()` to return nonzero if `msg->min_expires` is NULL. I couldn't quite work out what logic was intended. I think the code in this PR will fix the segfault, but it may not correct the intended behaviour.

**Closing issues**

### DIFF
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -674,7 +674,8 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 			goto done;
 		}
 		if (0 == parse_min_expires(msg)) {
-			rec->expires = (unsigned int)(long)msg->min_expires->parsed;
+			if (msg->min_expires)
+				rec->expires = (unsigned int)(long)msg->min_expires->parsed;
 			if(send_register(cb_param->hash_index, rec, NULL)==1)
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Hi @jes , imho the proper fix would be to have `parse_min_expires()` returning false if there is no hdr or if it failed to parse.

bogdan-iancu:
@jes , see me alternative commit - if works for you, I will backport and close here.

jes:
@bogdan-iancu looks good to me!

It solves the segfault, and the logic looks fine for the `uac_registrant` case. I didn't check if anything else is relying on the old behaviour with `parse_min_expires()` returning 0 with `msg->min_expires == NULL`, but it seems unlikely and you're more likely to know than me anyway.

bogdan-iancu:
Thanks @jes for the confirmation, the backports are in place, closing here

PR #3166, opened 2023-08-24
cachedb_redis_dbase: Fix the type of port with unsigned short.

nphantom:
**Summary**

**Details**

I used the modparam to set the cachedb_url as redis:opensips://:xxxxxx:35454/, and then encountered an error: failed to open redis connection xxxx:35454 - Servname not supported for ai_socktype. Upon inspecting the code, I realized that the cluster_node and redis_con have inconsistent port types, causing 35454 to become a negative value.

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/cachedb_redis/cachedb_redis_dbase.h
+++ b/modules/cachedb_redis/cachedb_redis_dbase.h
@@ -37,12 +37,12 @@
 #endif
 
 typedef struct cluster_nodes {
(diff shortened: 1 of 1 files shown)


liviuchircu:
Good catch, @nphantom!  Your fix will make it into next week's round of minor stable releases, just in time.

PR #3187, opened 2023-09-12
`dialog`: Fix memory leak in `$dlg_val(...)`

benceszigeti:


bogdan-iancu:
Closing, the fix is already in place

PR #3203, opened 2023-09-21
drouting: allow DB_BIGINT for dr_rules.ruleid column

john08burke:
**Summary**
Currently the `drouting` module only has support for `DB_INT` type for the **dr_rules.ruleid** column. This PR adds support for the `DB_BIGINT` type as well.

**Details**
I ran into this when trying to integrate the `drouting` module with some custom `VIEW` based schema that was using `BIGINT` for the ID column. The resulting logs are below and the table fails to load.

> ERROR:drouting:dr_load_routing_info: column ruleid has a bad type [1], accepting only [0]

This should be a pretty safe one!

### DIFF
--- a/modules/drouting/dr_load.c
+++ b/modules/drouting/dr_load.c
@@ -641,7 +641,7 @@ rt_data_t* dr_load_routing_info(struct head_db *part,
 			for(i=0; i < RES_ROW_N(res); i++) {
 				row = RES_ROWS(res) + i;
 				/* RULE_ID column */
-				check_val( rule_id_drr_col, ROW_VALUES(row), DB_INT, 1, 0);
+				check_val2( rule_id_drr_col, ROW_VALUES(row), DB_INT, DB_BIGINT, 1, 0);
 				int_vals[INT_VALS_RULE_ID_DRR_COL] = VAL_INT (ROW_VALUES(row));
 				/* GROUP column */
 				check_val( group_drr_col, ROW_VALUES(row)+1, DB_STRING, 1, 1);


bogdan-iancu:
Shouldn't we extrapolate this to all tables with IDs? the `dr_gateways` and `dr_carriers` do also have a similar ID column.

bogdan-iancu:
@john08burke , your note points out a real issue actually. The DB_BIGINT is actually a `long long` , so a 64 bits data type. This may be incompatible with : 
* using `VAL_INT()` macro, so yes, we should either use `VAL_BIGINT()` (but it is an ugly trick, rely on the data alignment inside the union) or use a ternary operator to use the right macro
* the final holder for the rule ID is an integer :-/ (for Carriers and GWs we have string ID which is used only for log printing and it is not stored into internal structures), so it may be subject to an overflow (while reading 64b from DB)

So, at least, we need to upgrade the `rule->id` to `long long` (and also check the implications of that - not sure where this ID is used further in the code).
And ideally come up with a 100% correct approach on DB_INT / DB_BIGINT issue

modules/drouting/dr_load.c
```diff
@@ -368,9 +368,9 @@ rt_data_t* dr_load_routing_info(struct head_db *part,
...
 			/* DB ID column */
-			if ( VAL_TYPE( ROW_VALUES(row) ) == DB_INT ) {
-				/* if INT type, convert it to string */
-				check_val( id_drd_col, ROW_VALUES(row), DB_INT, 1, 0);
+			if ( VAL_TYPE( ROW_VALUES(row) ) == DB_INT || VAL_TYPE( ROW_VALUES(row) ) == DB_BIGINT ) {
+				/* if INT or BIGINT type, convert it to string */
+				check_val2( id_drd_col, ROW_VALUES(row), DB_INT, DB_BIGINT, 1, 0);
 				/* int2bstr returns a null terminated string */
 				str_vals[STR_VALS_ID_DRD_COL] =
 					int2bstr((unsigned long)VAL_INT(ROW_VALUES(row)),
```
bogdan-iancu, line 376: Here we need to read as VAL_INT or VAL_BIGINT, to ensure correctness.

modules/drouting/dr_load.c
```diff
@@ -522,9 +522,9 @@ rt_data_t* dr_load_routing_info(struct head_db *part,
...
 				/* DB ID column */
-				if ( VAL_TYPE( ROW_VALUES(row) ) == DB_INT ) {
-					/* if INT type, convert it to string */
-					check_val( id_drc_col, ROW_VALUES(row), DB_INT, 1, 0);
+				if ( VAL_TYPE( ROW_VALUES(row) ) == DB_INT || VAL_TYPE( ROW_VALUES(row) ) == DB_BIGINT ) {
+					/* if INT or BIGINT type, convert it to string */
+					check_val2( id_drc_col, ROW_VALUES(row), DB_INT, DB_BIGINT, 1, 0);
 					/* int2bstr returns a null terminated string */
 					str_vals[STR_VALS_ID_DRC_COL] =
 						int2bstr((unsigned long)VAL_INT(ROW_VALUES(row)),
```
bogdan-iancu, line 530: similar, read as VAL_INT or VAL_BIGINT.

modules/drouting/dr_load.c
```diff
@@ -641,7 +641,7 @@ rt_data_t* dr_load_routing_info(struct head_db *part,
 			for(i=0; i < RES_ROW_N(res); i++) {
 				row = RES_ROWS(res) + i;
 				/* RULE_ID column */
-				check_val( rule_id_drr_col, ROW_VALUES(row), DB_INT, 1, 0);
+				check_val2( rule_id_drr_col, ROW_VALUES(row), DB_INT, DB_BIGINT, 1, 0);
 				int_vals[INT_VALS_RULE_ID_DRR_COL] = VAL_INT (ROW_VALUES(row));
```
bogdan-iancu, line 645: similar, read as VAL_INT or VAL_BIGINT.

Also, the rule ID is exposed as AVP to script, so as an integer :-/ (for Carriers and GWs we have string ID which is used only for log printing and it is not stored into internal structures), so it may be subject to an overflow (while reading 64b from DB)

So, at least, we need to upgrade the rule->id to long long (and also check the implications of that - not sure where this ID is used further in the code). And the `rule_id_avp` to expose the ID as string, not int.

bogdan-iancu (changes requested):

john08burke:
Hey @bogdan-iancu, I'll take a look this week and try to address your comments!

Issue #3228, opened 2023-10-16
[FEATURE] Make the use of `Makefile.openssl` global to all modules


tcressonKeyyo:
**Is your feature request related to a problem? Please describe.**
It is not related to a problem per se but I ran into a little tricky thing when recompiling by hand `OpenSSL` and linking it to the `stir_shaken` module.

I modified (for my need) the `Makefile.openssl` to point exactly to the location of my libs and put the `rpath` accordingly. I thought it would be enough but it happens that the `stir_shaken` module has its own copy of the `Makefile.openssl` because it doesn't need `libssl` but only `libcrypto`.

It led the `stir_shaken` module to link with another version of `OpenSSL` which, in its turn, led the method `w_stir_auth` to fail on some `OpenSSL` call.

**Describe the solution you'd like**
I think it would be better if modules depending on `OpenSSL` always included `Makefile.openssl`. This way there would be no `OpenSSL` versions  when another one is needed.

**Implementation**
  - Component: It would modify the `Makefiles` in the following modules => `stir_shaken`, `cahedb_mongodb` and `cachedb_redis` (to include `Makefile.openssl`). It would also modify the `Makefile.openssl` to use a kind of flag to determine if only `libcrypto` is needed (like `USE_ONLY_CRYPTO` or something else).
  - Type: Compilation.

tcressonKeyyo:
I'm already trying to implement it and will submit the result of the implementation to you if the idea seems OK

razvancrainea:
Lets hold this back unil #3231 is sorted out/merged.

PR #3231, opened 2023-10-23
Use makefile openssl everywhere

tcressonKeyyo:
**Summary**
This PR is aiming to solve the feature request https://github.com/OpenSIPS/opensips/issues/3228

**Details**
One problem arose when I tried to compile my own version of `OpenSSL` for my needs and tried to use it for compiling `OpenSIPS` and its modules (in particular the `stir_shaken` module). I modified the file `Makefile.openssl` for my needs and it happend that the `stir_shaken` module had its own dependencies to `OpenSSL` (not including the file `Makefile.openssl`).

I found two other modules that did not use `Makefile.openssl` but I did not need to use them (but I'm still modifying them in order to use `Makefile.openssl`).

**Solution**
This PR enhances the `Makefile.openssl` to propose linking with either just `libcrypto` (which is standalone) or both `libcrypto` and `libssl` (`libssl` needs `libcrypto`) and make every module depending on `OpenSSL` include `Makefile.openssl`.

### DIFF
--- a/Makefile.openssl
+++ b/Makefile.openssl
@@ -4,23 +4,39 @@
 # Needed to build with openssl 3.x
 DEFS += -Wno-deprecated-declarations
 
+# normalize USE_ONLY_CRYPTO value
+ifeq ($(USE_ONLY_CRYPTO),)
+USE_ONLY_CRYPTO=false
+endif
+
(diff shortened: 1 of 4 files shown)


sobomax:
@tcressonKeyyo merged, thanks!

PR #3247, opened 2023-11-15
Add $true and $false.

sobomax:
**Summary**

This change adds $true and $false constants.

**Details**

Most of the real modern programming languages have special value for boolean "true" and "false" values. Those are useful for initializations (i.e. $var(foo) = $true), also to use as an argument to a function. In addition, it can be used to forcefully disable / enable code block in pre-processing and/or debugging.

For what's worth, the ChatGPT thinks those already exist:

```
You
what is the simplest expression that always evaluates to false in opensips?

ChatGPT
The simplest expression that always evaluates to false in OpenSIPS is $false. You can use it in an if statement like so:

if ($false) {
  # This block will never execute
}
```

**Solution**

$true and $false values have been added into core variables.

**Compatibility**

None expected.


sobomax:
P.S. I pushed some cleanup making variable name really const (and propagated str_const_init()) while I am here.

liviuchircu:
Hey, @sobomax!

Could you detail which problem you trying to solve using these new `$true` and `$false` variables that the built-in `true` (1) and `false` (0) _opensips.cfg_ constants cannot solve?

sobomax:
@liviuchircu sorry did not realize true / false exist. Searching the code did not bring any clues. Thanks!

PR #3255, opened 2023-11-24
Don't send an ACK to a 500 until *after* the callbacks

jes:
**Summary**

We had an issue where HEP traces would show the ACK to a 500 being sent out before the 500 was received.

**Details**

This is confusing because it makes it look like OpenSIPS is preemptively ACK'ing a 500 it hasn't yet received.

**Solution**

This PR delays the sending of the ACK until after the callbacks for the 500 have run, which allows the HEP for the 500 response to be transmitted before the ACK and its HEP are transmitted.

**Compatibility**

I don't imagine this breaks any other scenarios, but it is conceivable that it does. I don't know enough.

**Closing issues**

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -1581,15 +1581,6 @@ int reply_received( struct sip_msg  *p_msg )
 		reset_timer( &uac->request.fr_timer);
 	}
 
-	/* acknowledge negative INVITE replies (do it before detailed
-	 * on_reply processing, which may take very long, like if it
-	 * is attempted to establish a TCP connection to a fail-over dst */
-	if (is_invite(t) && ((msg_status >= 300) ||
-	(is_local(t) && !no_autoack(t) && msg_status >= 200) )) {
-		if (send_ack(p_msg, t, branch)!=0)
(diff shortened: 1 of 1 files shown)


jes:
No updates here.

jes:
No updates here.

bogdan-iancu:
Hi @jes , there is a very important comment, just on top of the code you moved:
````
/* acknowledge negative INVITE replies (do it before detailed
	 * on_reply processing, which may take very long, like if it
	 * is attempted to establish a TCP connection to a fail-over dst */
````
The idea here is to be able to send back the ACK (as confirmation for the negative reply) and stop any potential retransmission from the callee side - any extra processing here (like reply related callbacks) may delay the sending of the ACK and increase the chances of useless negative reply retransmissions.

Still, I agree with you on the bogus tracing behavior (like tracing first the ACK and then the reply) - it is a know issue and there is an outstanding bug report on that. Nevertheless, we do not have a solution here to make everybody happy :)

liviuchircu:
Enough time has passed here for testing the fix on `3.5` beta - now also backported to `3.4`.  Closing as "Fixed" until further notice.

jes:
Thanks!

PR #3262, opened 2023-12-01
topology_hiding: fix encoded Contact length calculation

jes:
**Summary**

OpenSIPS can segfault when topology_hiding's encoded Contact header is too big to fit in its calculated length.

**Details**

The calculated length should be accumulated in `total_len`, but instead the assignment is to `suffix_len`, presumably due to a copy and paste error.

This bug was introduced in https://github.com/OpenSIPS/opensips/commit/e23be5d19a4865083971d17dc719a4d4d3436a24 presumably as a copy-and-paste mistake.

**Solution**

My solution is to accumulate the length in `total_len` instead of in `suffix_len`.

**Compatibility**

Probably doesn't break other scenarios.

**Closing issues**


jes:
No updates here.

jes:
No updates here.

bogdan-iancu:
@jes , yes, you are totally right on this, good catch !

jes:
Cheers!

PR #3263, opened 2023-12-01
uac_registrant: compare Contact URI case-insensitively

jes:
**Summary**

We had an issue where the response Contact URI had ";transport=udp" changed to ";transport=UDP".

**Details**

The unequal casing of "UDP" meant that `uac_registrant` did not recognise the Contact header, and therefore ignored the `;expires=...` value on the header, leading to OpenSIPS failing to re-register the user at the correct interval.

**Solution**

This commit solves the problem by comparing the Contact URI case-insensitively.

**Compatibility**

I doubt this breaks any scenarios.

**Closing issues**

### DIFF
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -482,7 +482,7 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 		while (contact) {
 			/* Check for binding */
 			if (contact->uri.len==rec->contact_uri.len &&
-				strncmp(contact->uri.s,rec->contact_uri.s,contact->uri.len)==0){
+				strncasecmp(contact->uri.s,rec->contact_uri.s,contact->uri.len)==0){
 				if (contact->expires && contact->expires->body.len) {
 					if (str2int(&contact->expires->body, &exp)<0) {
 						LM_ERR("Unable to extract expires from [%.*s]"


vasilevalex:
Maybe we can continue and improve this one #2558 ?

modules/uac_registrant/registrant.c
```diff
@@ -482,7 +482,7 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 		while (contact) {
 			/* Check for binding */
 			if (contact->uri.len==rec->contact_uri.len &&
-				strncmp(contact->uri.s,rec->contact_uri.s,contact->uri.len)==0){
+				strncasecmp(contact->uri.s,rec->contact_uri.s,contact->uri.len)==0){
```
ovidiusas, line 485: This is not correct because if the username was altered, the URIs are not the same:
 `sip:alice@AtLanTa.CoM;Transport=udp`             (different usernames)
 `sip:ALICE@AtLanTa.CoM;Transport=UDP`

jes: Thanks, what would you think of adding a `strncmp_lc_transport` that takes a `pkg_malloc` copy of both strings, turns the `;transport=...` into lowercase, then `strcmp`s the copies with lowercased transport, and then frees the copies and returns the result of the `strcmp`?

ovidiusas: Take a look at `compare_uris()` from `parse_uri.c`.
It does the proper job if URIs are passed as parsed structures, but comparing raw URIs is bad :(
I pushed some fixes on the `compare_uris` branch:
https://github.com/OpenSIPS/opensips/tree/compare_uris
Please give it a try and let me know the results.

ovidiusas: Did you test the new branch in production and it's working ok for you?

jes: Not in production, no, only in a test setup. I registered a user with incorrect case in the `transport` param and it worked & did not cause any obvious issues.

jes:
No updates here.

jes:
No updates here.

ovidiusas:
This has be addressed but the following commits:
https://github.com/OpenSIPS/opensips/commit/54b5a353a75deacaf35e5ccbe8c75f55eece1366
https://github.com/OpenSIPS/opensips/commit/a796a4962fcaadd05bb7e0c7ad87c09df050c5ef

jes:
Oh! Great success, thank you!

jes:
Hi @ovidiusas  @bogdan-iancu - how come this change didn't make it into 3.4.4?

jes:
Reopening for visibility.

jes:
It was merged but not included in the release?

jes:
As per https://github.com/OpenSIPS/opensips/pull/3263#issuecomment-1937793957 - a fix was merged but didn't make it into the release, what is happening there?

jes:
Per https://www.opensips.org/pub/opensips/3.5.0-beta/ChangeLog it looks like this is going to make it into a release, closing.

PR #3264, opened 2023-12-03
usrloc: simplify processing of the refresh events

sobomax:
**Summary**

usrloc: simplify processing of the refresh events

**Details**

The code that generates refresh event tries to use sock_info* available in the contact structure to pass sock_str over. This requires allocating sock_info as well as setting up appropriate pointers. 

**Solution**

Do not abuse "struct socket_info" just to pass sock_str over. Instead, add str to the refresh event data and pass that alone. Not only it reduces amount of data to pass, but also makes code much easier to read and understand.

**Compatibility**

None

**Closing issues**

Nope

### DIFF
--- a/modules/usrloc/ul_evi.c
+++ b/modules/usrloc/ul_evi.c
@@ -498,9 +498,12 @@ void ul_raise_contact_event(event_id_t _e, const ucontact_t *_c)
 }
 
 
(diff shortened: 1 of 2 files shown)


liviuchircu:
LGTM @sobomax -- a very useful refactoring!

PR #3266, opened 2023-12-05
proto_tls: report error with correct variable

jes:
Assigning to `len` at this point has no effect, and comparing with the
code not far above, with the same comment, and the code under
`con_release` ending with `return rlen;` it is obvious that the
intention was to assign to `rlen` instead of `len`.


bogdan-iancu:
Thank you @jes !

PR #3270, opened 2023-12-11
uac_registrant: add "failure_retry_interval" modparam

jes:
**Summary**

This PR adds a modparam "failure_retry_interval" to the `uac_registrant` module to limit the frequency with which it will retry failed registrations.

**Details**

Currently, if a registration fails (for example because the credentials are wrong), OpenSIPS will vainly retry the registration every `timer_interval` seconds, causing excessive load both to OpenSIPS and to the remote server.

This PR allows the failure retry interval to be larger than the timer interval.

**Solution**

This PR adds an option "failure_retry_interval", and code to use it, so that failed registrations are not repeated until at leat `failure_retry_interval - timer_interval` has elapsed.

**Compatibility**

It should not break any scenarios, because the default value of 0 causes `uac_registrant` to retry the registration the next time the timer fires, matching the existing behaviour.

**Closing issues**

### DIFF
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -110,6 +110,7 @@ uac_auth_api_t uac_auth_api;
 
 unsigned int default_expires = 3600;
 unsigned int timer_interval = 100;
+unsigned int failure_retry_interval = 0;
 
 reg_table_t reg_htable = NULL;
 unsigned int reg_hsize = 1;
@@ -139,6 +140,7 @@ static const param_export_t params[]= {
 	{"hash_size",		INT_PARAM,			&reg_hsize},
 	{"default_expires",	INT_PARAM,			&default_expires},
 	{"timer_interval",	INT_PARAM,			&timer_interval},
+	{"failure_retry_interval", INT_PARAM,			&failure_retry_interval},
 	{"enable_clustering",	INT_PARAM,			&enable_clustering},
 	{"db_url",		STR_PARAM,			&db_url.s},
 	{"table_name",		STR_PARAM,			&reg_table_name.s},
@@ -417,7 +419,7 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 				if (!(rec->flags&REG_ENABLED)) {
 					/* succesfully unREGISTERED */
 					rec->state = NOT_REGISTERED_STATE;
-					rec->registration_timeout = 0;
+					rec->registration_timeout = now + failure_retry_interval - timer_interval;
 				} else {
 					/* registrant got enabled while waiting for a reply to
 					 * a previous unregister */
@@ -430,7 +432,7 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 						rec->last_register_sent = now;
 						rec->state = REGISTERING_STATE;
 					} else {
-						rec->registration_timeout = now + rec->expires - timer_interval;
+						rec->registration_timeout = now + failure_retry_interval - timer_interval;
 						rec->state = INTERNAL_ERROR_STATE;
 					}
 				}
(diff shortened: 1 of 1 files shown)


jes:
I noticed there are a few places that update `rec->state` to an error state without changing `rec->registration_timeout`.

I think this would have the effect of trying to re-register every `timer_interval` regardless of what `failure_retry_interval` is set to, maybe I should make it set `rec->registration_timeout` in every one of these as well.

jes:
No updates here.

jes:
No updates here.

bogdan-iancu (commented):
IMHO, the proposed changes are a bit too much. There is an already existing throttling  (for next registrations) if some error happens by setting the new timeout to `now + rec->expires - timer_interval` - meaning the next retry will happen with a new re-register cycle, right?
Indeed, the fix should be done when the timeout is to `0` for `WRONG_CREDENTIALS_STATE` too.
Nevertheless, when transiting to `NOT_REGISTERED_STATE`, I'm not sure what should be the correct behavior - to force a right away re-reg (as now, by setting to 0) ? or ? Maybe @ovidiusas  can bring some light here.
What is a key change in the PR is the test of the timeout value you added in `run_timer_check()` for the faulty states - that is a must, otherwise the timeout will be simply ignored in these states. So 90% of the fix is in this change.
I would say that by fixing the timeout test and the timeout set in `WRONG_CREDENTIALS_STATE` state, we should be good - I do not see a real value by adding the new `failure_retry_interval` param - the retry will be automatically done at the "expiration" time interval, per-record bases.

jes:
But if the user gets disabled, the registration will start failing but OpenSIPS will retry every timer_interval, spamming the server? `failure_retry_interval` means registrations that fail will not be tried again as soon as they normally would be.

ovidiusas:
It is unclear for me what exactly are we trying to fix here.
If an account has wrong credentials, then that account should be re-provisioned.
Also, an account can be disabled/enabled via the management interface.
One can simply inspect all registrations via mi and take action based on their status.
This looks more of a management issue then a protocol issue.

Also, third party registrations is an option: a separate instance of an opensips server can perform all the registrations, pointing the AORs to the main opensips: [third party registration](https://opensips.org/docs/modules/3.4.x/uac_registrant.html#param_third_party_registrant_column)

bogdan-iancu:
What I suggest here:

1. in `NOT_REGISTERED_STATE` we need to keep the 0 timeout, to force registering asap
2. in `WRONG_CREDENTIALS_STATE` we should NOT set 0 timeout, as this will continuously try to re-register, which is wrong. There is an error, some throttling is needed. This should be handled similar to ` REGISTRAR_ERROR_STATE` or `REGISTER_TIMEOUT_STATE`
3. for all error cases( ` REGISTRAR_ERROR_STATE`, `REGISTER_TIMEOUT_STATE`, `INTERNAL_ERROR_STATE`, `WRONG_CREDENTIALS_STATE`) we can add an optional delaying if needed ,by default we can keep it as now `now + rec->expires - timer_interval`, but we can do something like `now + (failure_retry_interval?failure_retry_interval:rec->expires) - timer_interval`. So, with `failure_retry_interval` default 0, we will have the original behavior to re-try after `rec->expires`
4. maybe adding a new MI function to force the re-registration of a record, by forcing the timeout to 0.

What do you think of this?

jes:
Yes, great idea, that sounds perfect!

I see the problem now that if you don't set `failure_retry_interval` in my initial code then it defaults to 0.

bogdan-iancu:
yes, if you do not set `failure_retry_interval`, it will default to 0, which will default the time to retry to `now + rec->expires - timer_interval` for `REGISTRAR_ERROR_STATE`, `REGISTER_TIMEOUT_STATE`, `INTERNAL_ERROR_STATE`, `WRONG_CREDENTIALS_STATE`  - see here we include the **WRONG_CREDENTIALS_STATE** which originally was 0 (immediate retry)

ovidiusas:
One enhancement: the failure retry value could be increased with each iteration, up to a certain value. On the first failure the extra delay is failure_retry_interval, on the second failure the delay could be double the failure_retry_interval and on the third it could be triple and continue to stay as is for subsequent failures.

Maybe in the future we should add events and some management functions. This will allow a tight control right from the script.

PR #3275, opened 2023-12-24
Ensure VERSIONTYPE is always defined

lemenkov:
**Summary**
If someone downloads tarball from GitHub then this archive won't have .git directoryor .gitrevision (autogenerated) file. Let's ensure we still have VERSIONTYPE macro set.

This also allows us to remove unnecessary guards in two places.

---
Also I'd like to say Merry Christmas and Happy New Year!

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -507,6 +507,8 @@ else
   VERSIONTYPE = $(shell [ -f "${TOP_SRCDIR}/.gitrevision" ] && echo "git")
   ifneq ($(VERSIONTYPE),)
     THISREVISION = $(shell cat ${TOP_SRCDIR}/.gitrevision)
+  else
+    VERSIONTYPE = unknown
   endif
   OLDREVISION = "$(THISREVISION)"
 endif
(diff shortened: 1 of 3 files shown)


liviuchircu (approved):
Happy New Year, @lemenkov!  I'd say this is a net improvement, from this:

```
Compiling mem/mem_dbg_hash.c
main.c:176:10: error: ‘VERSIONTYPE’ undeclared here (not in a function); did you mean ‘VERSION_TABLE’?
  176 |  .type = VERSIONTYPE,
      |          ^~~~~~~~~~~
      |          VERSION_TABLE
main.c:177:9: error: ‘THISREVISION’ undeclared here (not in a function)
  177 |  .rev = THISREVISION
      |         ^~~~~~~~~~~~
Compiling mem/mem.c
```

(where it seems that a source tree missing both the `.git/` and the `.gitrevision` gives a compile error -- the opensips.org tarballs are generated from a `git clone`, so they will at least include the `.gitrevision`)... to this:

```
$ ./opensips  -V
version: opensips 3.4.3 (x86_64/linux)
unknown revision: unknown
main.c compiled on 17:53:46 Jan  4 2024 with gcc 10.0.0
```

Which fixes the compilation even on the "worst" case.  So let's merge it!

PR #3277, opened 2023-12-29
dialog: add one more param to dlg_send_sequential mi function

smititelu:
Add one more "headers" parameter to dialog mi function: dlg_send_sequential

This will allow sending "REFER" method, in-dialog, and adding "Refer-To:" header to it.

Thanks and happy new year! :D

### DIFF
--- a/modules/dialog/dialog.c
+++ b/modules/dialog/dialog.c
@@ -433,6 +433,7 @@ static const mi_export_t mi_cmds[] = {
 		{mi_send_sequential_dlg, {"callid", "mode", "body", 0}},
 		{mi_send_sequential_dlg, {"callid", "method", "body", 0}},
 		{mi_send_sequential_dlg, {"callid", "method", "body", "mode", 0}},
+		{mi_send_sequential_dlg, {"callid", "method", "body", "mode", "headers", 0}},
 		{EMPTY_MI_RECIPE}}
 	},
 	{EMPTY_MI_EXPORT}
--- a/modules/dialog/dlg_req_within.c
+++ b/modules/dialog/dlg_req_within.c
@@ -796,7 +796,7 @@ error:
 }
 
 static mi_response_t *mi_send_sequential(struct dlg_cell *dlg, int sleg,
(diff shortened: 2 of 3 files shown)


modules/dialog/dialog.c
```diff
@@ -433,6 +433,7 @@ static const mi_export_t mi_cmds[] = {
 		{mi_send_sequential_dlg, {"callid", "mode", "body", 0}},
 		{mi_send_sequential_dlg, {"callid", "method", "body", 0}},
 		{mi_send_sequential_dlg, {"callid", "method", "body", "mode", 0}},
+		{mi_send_sequential_dlg, {"callid", "method", "mode", "headers", 0}},
```
razvancrainea, line 436: The way you declared it creates an overlapping for calling the MI command with positional parameters `callid+method+body+mode` has the same number of parameters as `callid+method+mode+headers` - this means that the function will always need to be called with named parameters, including the old version - this is not baclwards compatible.

In order to make it backwards compatible, I would add an "extended" recipe, i.e. adding the `headers` parameter after `callid+method+body+mode` - any reason why you did not do that? If there is some limitations (i.e. you cannot call the function with an empty body), then that is fine, but we need to be clear about the change between the versions in the migration page.

Nevertheless, I would not rule out the option to add a body and headers too.
I would suggest adding the "body" parameter as well, as we don't want to rule out the option to add both body and headers.

razvancrainea: Besides that, everything looks OK.

The only missing bit is the documentation - can you add the parameter in the documentation [xml](https://github.com/OpenSIPS/opensips/blob/master/modules/dialog/doc/dialog_admin.xml#L2563) as well please?

razvancrainea (changes requested):

modules/dialog/dialog.c
smititelu: Thanks for the feedback! I've updated PR accordingly.

razvancrainea (approved):

razvancrainea:
Thanks for your contribution!

PR #3279, opened 2024-01-03
STIR SHAKEN - Exported Events

Integration-IT:
**Summary**
Add new exported events E_STIR_AUTH, E_STIR_VERIFY, E_STIR_CHECK, E_STIR_CHECK_CERT.
1/4 PR to link E_STIR_CHECK events.

**Details**
Add new concept events.
Lazier to use and avoids repetition of lines of code inside exported functions.
event_gen_event need level message and mode function to raise an event.

Let suppose you would like raise an event in w_stir_check function linked to the E_STIR_CHECK event to inform it is not possible to find the Identity header.

Just do:
```
// static int event_gen_event(struct sip_msg *msg, int mode, int level)
event_gen_event(msg,2,2);
```
The first 2 mode will point to CHECK event.
The second 2 status will point to FAILED_TO_FIND_IDENTITY state.

**POC**
Core debug:
```
INFO:core:evi_publish_event: Registered event <E_STIR_AUTH(6)>
INFO:core:evi_publish_event: Registered event <E_STIR_VERIFY(7)>
INFO:core:evi_publish_event: Registered event <E_STIR_CHECK(8)>
INFO:core:evi_publish_event: Registered event <E_STIR_CHECK_CERT(9)>
```
opensips.cfg
```
event_route[E_STIR_CHECK] {
	xlog("SCRIPT:DBG: [E_STIR_CHECK] | We got an event from:$param(from), to:$param(to), call-id:$param(callid), state:$param(state)\n");
}
```
syslog
```
NOTICE:SCRIPT:DBG: [E_STIR_CHECK] | We got an event from:"alice"<sip:100@1.1.1.1>;tag=37663030303030313133633401363937303435383034, to:"bob"<sip:101@1.1.1.1>, call-id:887795433350235830738747, state:No Identity header found
```

**Compatibility**
```
version: opensips 3.5.0-dev (x86_64/linux)
```

**Closing issues**

### DIFF
--- a/modules/stir_shaken/stir_shaken.c
+++ b/modules/stir_shaken/stir_shaken.c
@@ -67,6 +67,7 @@
 #include "../../parser/parse_to.h"
 #include "../../parser/parse_uri.h"
 #include "../../parser/parse_param.h"
+#include "../../parser/parse_event.h"
 #include "../../data_lump.h"
 #include "../../lib/cJSON.h"
 #include "../../context.h"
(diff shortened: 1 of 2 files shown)


bogdan-iancu:
@Integration-IT , I somehow fail to see the value added by these events - the reason of the failure is also reported via return code (back to the script) - so the same code (for handling / reporting the failures) you may have in an event_route may be also in a sub-route you may call in case of failure. Or maybe I'm missing something here.

PR #3281, opened 2024-01-05
tls_openssl: only verify client certs if configured to require them

jes:
**Summary**
If you want opensips to verify server certs you have to enable `verify_cert`, but this also implicitly makes it verify *client* certs, when presented, even if you don't have `require_client_cert` enabled.

**Details**

There was previously no way to configure opensips to verify server certs without having it also verify client certs, even if you configured it to make client certs optional!

**Solution**

This PR makes `verify_cert` only apply in the case of opensips connecting out to a server, and `require_client_cert` only apply in the case of a client connecting into opensips.

**Compatibility**

The only scenario I can imagine it breaking is if you have `require_client_cert` but *not* `verify_cert`, in which case formerly it would allow any client cert (but require *some* client cert to be presented), but now it would verify the client certificate. I don't expect this to be a problem because what is the point in requiring a client cert if you don't verify it?

**Closing issues**

### DIFF
--- a/modules/tls_openssl/openssl_config.c
+++ b/modules/tls_openssl/openssl_config.c
@@ -202,15 +202,10 @@ static void get_ssl_ctx_verify_mode(struct tls_domain *d, int *verify_mode)
 		 *   SSL_VERIFY_PEER.
 		 */
 
-		if( d->verify_cert ) {
-			*verify_mode = SSL_VERIFY_PEER;
-			if( d->require_client_cert ) {
-				LM_INFO("client verification activated. Client "
-						"certificates are mandatory.\n");
-				*verify_mode |= SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
@jes , as I read the current code, you can do a `verify_cert` without `require_client_cert` - basically to verify whatever (if any)  cert is provided by the client (by itself), without asking. This makes sense IMHO.

On the other hand, a `require_client_cert` will not be honored without  `verify_cert` - if you require the cert, then verify it too! Which again seems OK to me.

So going back to your statement : 
````
There was previously no way to configure opensips to verify server certs without having it also verify client certs, even if you configured it to make client certs optional!
````
You can achieve this (for a server RLS domain) by simply setting the `verify_cert` only for that particular domain, right ?

jes:
Ah! Great idea. That works, thanks very much :).

PR #3306, opened 2024-02-16
Getsockopt warnings

jes:
**Summary**
We saw some nonsensical errors from `tls_openssl`, like:

    failed to retrieve SO_ERROR [server=...] (2) No such file or directory

I looked in `openssl_conn_ops.c` and found that the socket error is grabbed via `getsockopt()` of `SO_ERROR`, but I don't think `ENOENT` (No such file or directory) is applicable on a socket.

**Details**
The cause of the problem is that the error value isn't actually being successfully retrieved by `getsockopt()`, instead `getsockopt()` is failing for cause unknown, leaving its argument unchanged, and the argument `err` is also the value of the enclosing `switch`/`case` block, and we're in `case SSL_ERROR_WANT_READ`, and `SSL_ERROR_WANT_READ == ENOENT == 2`. Hence the nonsensical error message.

**Solution**
The solution is to check whether `getsockopt()` returned nonzero, and if so warn with the error that `getsockopt()` returned.

Obviously there is more to it than this (why is `getsockopt()` not working? what has gone wrong?), but logging a warning about the `getsockopt()` error seems strictly preferable to *ignoring* the error.

**Compatibility**
I can't imagine this breaks any scenarios as it doesn't change any logic, it just makes log output less confusing.

**Closing issues**

### DIFF
--- a/modules/tls_openssl/openssl_conn_ops.c
+++ b/modules/tls_openssl/openssl_conn_ops.c
@@ -756,7 +756,12 @@ again:
 #endif
 				{
 					err_len=sizeof(err);
-					getsockopt(fd, SOL_SOCKET, SO_ERROR, &err, &err_len);
+					if (getsockopt(fd, SOL_SOCKET, SO_ERROR, &err, &err_len) != 0) {
+						LM_WARN("getsockopt error: fd=%d [server=%s:%d]: (%d) %s\n", fd,
+								ip_addr2a(&con->rcv.src_ip), con->rcv.src_port,
+								errno, strerror(errno));
+						goto failure;
+					}
 					if ((err==0) && (poll_err==0))
 						continue; /* retry ssl connect */
 					if (err!=EINPROGRESS && err!=EALREADY){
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Hi @jes , I think there is a bit more here. Your patch is suppressing the log if the `getsockopt()` fails, but the following test on `err` should be also, right? As the `err` is set by `getsockopt()` which just failed, so the `err` value is not reliable. Maybe if `getsockopt()` failed, we should set `err` to indicate kind of error for the later test....

bogdan-iancu:
Right, right, probably was way too morning for me and I read that `if` upside down ;). So, yes, you log that warning if the `getsockopt()` fails. Nevertheless, my second point is still valid - does it make sense to test and rely on the `err` value in the late code, considering that `getsockopt()` failed ?

jes:
OK, I've force-pushed a new version which now does `goto failure` or `goto error` as appropriate.

jes:
It looks like the CI run failed because it couldn't download some PPA for Ubuntu.

bogdan-iancu:
thank you @jes

jes:
Cheers! Will this make it into the next minor release?

bogdan-iancu:
yes, for sure it will !

PR #3309, opened 2024-02-22
[db_postgres] fix the SSL control over the "use_tls" parameter

bogdan-iancu:
Fix how the `use_tls` controls the SSL usage when connecting to the server. The main issue here is that the SSL may be used even if the `use_tls` is not set (as the "prefer" mode is by default used).
Now we have:

- use_tls=0 (default) - the SSL support is disabled and there is no attempt to use it;
- use_tls=1 with "tls_domain" present in the DB URL - the SSL support is enabled, either "require", either "verify-ca", depending on the certificate settings;
- use_tls=1 with no "tls_domain" present in the DB URL - the SSL support is enabled in best effort mode (or "prefer"); if supported by the server, it will be used, otherwise it will fall back to non-SSL.

### DIFF
--- a/modules/db_postgres/doc/db_postgres_admin.xml
+++ b/modules/db_postgres/doc/db_postgres_admin.xml
@@ -119,6 +119,34 @@ modparam("db_postgres", "timeout", 2)
         <section id="param_use_tls" xreflabel="use_tls">
 		<title><varname>use_tls</varname> (integer)</title>
 		<para>
+		Parameter to control the way the SSL support is used when connecting
+		to the Postgres server, as follows:
+		</para>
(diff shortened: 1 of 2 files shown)


benceszigeti (approved):

liviuchircu (approved):
LGTM, there were multiple issues with the `use_tls` modparam that this patch fixes.  Not only does it fix the default value (which happens to be `prefer` in _libpq-dev_ and **not** `disable`), but it also fixes the case when we enable TLS without requiring server CA validation by explicitly setting the mode to `prefer` instead of leaving the mode _unset_ (yet again...).

PR #3311, opened 2024-02-23
Add sip2json pvar

vladpaiu:
**Summary**
Add sip2json pvar

**Details**
Needed for cases when you want to pass a generic SIP message to a SIP agnostic entity, but still want to provide some layer of SIP parsing before sending the full message further.

**Solution**
JSON with First line, Headers and Body

**Compatibility**
Backwards compatible


bogdan-iancu:
Thank you @vladpaiu . Here is some feedback:

"Must" changes:
* this function should not be part of the core, but rather some module, maybe `cfgutils` or `sipmsgops` (even better)
* documentation is needed, describing the json structure 

"Nice to have" change:
* for the body sections, it would be nicer to have an array of body parts, to also deal with the multi-part body. The array will have all the time at least one part (if there are no multi-parts) and each array element will consist of an element containing an optional `Content-Type` field and the mandatory body part itself.

If you consider converting this pvar into a function, you may also parametrize it - like an optional parameter with white/black filters to control the headers to be added into JSON.

Now, on second thoughts, I see a better option to have this functionality as function in sipmsgops module.

vladpaiu:
> Thank you @vladpaiu . Here is some feedback:
> 
> "Must" changes:
> 
> * this function should not be part of the core, but rather some module, maybe `cfgutils` or `sipmsgops` (even better)
> * documentation is needed, describing the json structure
> 
> "Nice to have" change:
> 
> * for the body sections, it would be nicer to have an array of body parts, to also deal with the multi-part body. The array will have all the time at least one part (if there are no multi-parts) and each array element will consist of an element containing an optional `Content-Type` field and the mandatory body part itself.
> 
> If you consider converting this pvar into a function, you may also parametrize it - like an optional parameter with white/black filters to control the headers to be added into JSON.
> 
> Now, on second thoughts, I see a better option to have this functionality as function in sipmsgops module.

thanks for the feedback, will make the necessary changes and get back to you here.

bogdan-iancu:
@vladpaiu, any chances to get the MUST changes (at least) by tomorrow, so we can include this into the 3.5 release on 9th ?

vladpaiu:
> @vladpaiu, any chances to get the MUST changes (at least) by tomorrow, so we can include this into the 3.5 release on 9th ?

moved the implementation to sipmsgops and added documentation

bogdan-iancu:
Thank you @vladpaiu  - could you add couple of words in the doc about the format of the resulting json? is it like one node per line? or ? Thanks !
O will merge as soon as the CI tests are validated.

PR #3312, opened 2024-02-23
Add get_header_value() function

vladpaiu:
**Summary**
Add get_header_value() function

**Details**
$hdr() does not support passing pvars to it, sometimes you need to fetch a custom header name, added a new function which iterates for you instead of script iteration through all headers.

**Compatibility**
Backwards-compatible

### DIFF
--- a/modules/sipmsgops/README
+++ b/modules/sipmsgops/README
@@ -20,40 +20,41 @@ sipmsgops Module
               1.3.5. insert_hf(txt, hdr)
               1.3.6. append_urihf(prefix, suffix)
               1.3.7. is_present_hf(hf_name)
-              1.3.8. append_time()
-              1.3.9. is_method(name)
-              1.3.10. remove_hf(hname)
-              1.3.11. remove_hf_re(hname_expr)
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Wouldn't be a simpler approach to add to `$hdr` support for variable name (similar to [what `$avp` have](https://github.com/OpenSIPS/opensips/blob/master/pvar.c#L3363)), rather than adding something in overlapping ? Not to mention that the `$hdr` supports indexes, something that this new functions doesn't.

vladpaiu:
> Wouldn't be a simpler approach to add to `$hdr` support for variable name (similar to [what `$avp` have](https://github.com/OpenSIPS/opensips/blob/master/pvar.c#L3363)), rather than adding something in overlapping ? Not to mention that the `$hdr` supports indexes, something that this new functions doesn't.

Oops, you're right, this is already valid in latest OpenSIPS : 

$var(hdr_name) = "From";
xlog("Header value is $hdr($var(hdr_name)) \n");

Closing PR.

PR #3313, opened 2024-02-23
Add generate_data_md5 param

vladpaiu:
**Summary**
Data loaded in drouting feedback

**Details**
If generate_data_md5 is set, OpenSIPS will compute an MD5 hash for each partition data, and export that MD5 in the reload_status MI command as well as in the reload status reports, for easy comparison between multiple OpenSIPS instances which should have the same dataset loaded.

**Solution**
For clusterer drouting usage, the platform admin can monitor that they are in sync in terms of MD5 hash of loaded data.

**Compatibility**
Backwards-compatible

### DIFF
--- a/modules/drouting/README
+++ b/modules/drouting/README
@@ -59,6 +59,7 @@ Dynamic Routing Module
               1.3.36. extra_prefix_chars (str)
               1.3.37. extra_id_chars (str)
               1.3.38. rule_tables_query (str)
+              1.3.39. generate_data_md5 (int)
 
         1.4. Exported Functions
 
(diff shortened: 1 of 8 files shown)


bogdan-iancu:
HI @vladpaiu , I like the of such checksum. Here is some feedback:
* IMHO, "checksum" will be a more suggestive name versus "md5" (too tech)
* please drop the README file from the PR, keep only the doc xml - the README's are automatically generated and committed, no need to do it by hand.
PS: I  really like the fact you integrated this with the status/report support ;)

vladpaiu:
> HI @vladpaiu , I like the of such checksum. Here is some feedback:
> 
> * IMHO, "checksum" will be a more suggestive name versus "md5" (too tech)
> * please drop the README file from the PR, keep only the doc xml - the README's are automatically generated and committed, no need to do it by hand.
>   PS: I  really like the fact you integrated this with the status/report support ;)

thanks for the feedback, changed the param name to generate_data_checksum, let me know if all's good.

bogdan-iancu (approved):
LGTM, thank you !!

PR #3316, opened 2024-02-26
Add MI fetching of keys based on glob

vladpaiu:
**Summary**
Fetch all cachedb_local keys from mem based on a glob regex

**Details**
New feature to fetch all data based on a global, useful for monitoring

**Solution**
Go through the local cache, match each key against the glob and return it in the reply key-value array


bogdan-iancu:
LGTM

bogdan-iancu (approved):

PR #3317, opened 2024-02-26
Rework contact generation, add validation & enforce contact_user even for  requests taking Contact from server_address

vladpaiu:
**Summary**
Rework contact generation, add validation & enforce contact_user even for  requests taking Contact from server_address

**Details**
Due to server_address potentially containing PVARs, validate the server_address before usage
Enforce contact_user modparam even for  requests that generate the Contact from server_address

Thanks to Voicenter for reporting it & helping troubleshoot.

### DIFF
--- a/modules/b2b_logic/logic.c
+++ b/modules/b2b_logic/logic.c
@@ -476,23 +476,43 @@ void b2b_mark_todel( b2bl_tuple_t* tuple)
 
 int b2b_get_local_contact(struct sip_msg *msg, str *from_uri, str *local_contact)
 {
-	struct sip_uri ct_uri;
+	struct sip_uri ct_uri,server_uri;
 	const struct socket_info *send_sock = msg ?
 		(msg->force_send_socket?msg->force_send_socket:msg->rcv.bind_address):NULL;
 
 	if (server_address.len) {
 		if (pv_printf_s(msg, server_address_pve, local_contact) != 0) {
 			LM_WARN("Failed to print format string from 'server_address'\n");
-
-			if (msg) {
-				get_local_contact(send_sock, NULL, local_contact);
-			} else {
-				LM_ERR("No current SIP message, "
-					"failed to build Contact from send socket\n");
-				return -1;
+			goto msg_contact;
+		} else {
+			/* validate what we have built */
+			if (parse_uri(local_contact->s, local_contact->len, &server_uri) < 0) {
+				LM_ERR("Not a valid server sip uri [%.*s]\n", local_contact->len, local_contact->s);
+				goto msg_contact;
 			}
+
+			/* we have expanded the server address, need to add username, if needed */
+			if (contact_user) {
+				send_sock = grep_sock_info( &server_uri.host, server_uri.port_no, server_uri.proto);
+				if (send_sock == NULL) {
+					LM_ERR("Failed to find send socket for server address [%.*s]\n", local_contact->len, local_contact->s);
+					goto msg_contact;
+				}
+
+				memset(&ct_uri, 0, sizeof(struct sip_uri));
+				if (parse_uri(from_uri->s, from_uri->len, &ct_uri) < 0) {
+					LM_ERR("Not a valid FROM sip uri [%.*s]\n", from_uri->len, from_uri->s);
+					goto done;
+				}
+
+				get_local_contact(send_sock, &ct_uri.user, local_contact);
+				goto done;
+			} else
+				/* no contact username needed, we have our valid Contact header based on the server_address */
+				goto done;
 		}
 	} else {
+msg_contact:
 		if (msg) {
 			memset(&ct_uri, 0, sizeof(struct sip_uri));
 			if (contact_user && parse_uri(from_uri->s, from_uri->len, &ct_uri) < 0) {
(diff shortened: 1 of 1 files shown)


bogdan-iancu (commented):
There are couple of things not right here : 
* parsing the generated URI (based on  `server_address_pve`) is an unnecessary burn - as script writer you need to be sure you configuration is correct and not to expect OpenSIPS to parse it (for each message) just to be sure you are right - eventually you can do a dummy test at startup, to see if the pve generates a valid URI
* the replacement of the username part (if `contact_user` is on) in the generated URI (based on  `server_address_pve`) does not look right - you do not actually replace the username part, but build a completely new URI based on the username and socket - so if my `server_address_pve` was pushing some URI params or anything else, it will be lost. Normally you should have here a sctrict username replacement into the `local_contact` parsed URI - this may be simply done with 3 memcpy's .

vladpaiu:
> There are couple of things not right here :
> 
> * parsing the generated URI (based on  `server_address_pve`) is an unnecessary burn - as script writer you need to be sure you configuration is correct and not to expect OpenSIPS to parse it (for each message) just to be sure you are right - eventually you can do a dummy test at startup, to see if the pve generates a valid URI

Don't agree with the above ... OpenSIPS should validate the server address upon building, in order to avoid sending broken messages  ( due to the script writer's fault, indeed ). Can't be validated at startup, since it might contain pvars which can be only evaluated in the SIP message context. 

> * the replacement of the username part (if `contact_user` is on) in the generated URI (based on  `server_address_pve`) does not look right - you do not actually replace the username part, but build a completely new URI based on the username and socket - so if my `server_address_pve` was pushing some URI params or anything else, it will be lost. Normally you should have here a sctrict username replacement into the `local_contact` parsed URI - this may be simply done with 3 memcpy's .

This is indeed an issue - my use case didn't have uri params so I lost track of this. Needs correcting.

bogdan-iancu:
> > There are couple of things not right here :
> > 
> > * parsing the generated URI (based on  `server_address_pve`) is an unnecessary burn - as script writer you need to be sure you configuration is correct and not to expect OpenSIPS to parse it (for each message) just to be sure you are right - eventually you can do a dummy test at startup, to see if the pve generates a valid URI
> 
> Don't agree with the above ... OpenSIPS should validate the server address upon building, in order to avoid sending broken messages ( due to the script writer's fault, indeed ). Can't be validated at startup, since it might contain pvars which can be only evaluated in the SIP message context.

yes, the contact must be correct, but opensips should not continuously check if the configured data is correct or not. Like we do not check if the `received` AVP (in registrar) is a valid info, and so one. It is the duty  of the script writer to be sure the configured data is correct.

> 
> > * the replacement of the username part (if `contact_user` is on) in the generated URI (based on  `server_address_pve`) does not look right - you do not actually replace the username part, but build a completely new URI based on the username and socket - so if my `server_address_pve` was pushing some URI params or anything else, it will be lost. Normally you should have here a sctrict username replacement into the `local_contact` parsed URI - this may be simply done with 3 memcpy's .
> 
> This is indeed an issue - my use case didn't have uri params so I lost track of this. Needs correcting.

LEt me know if you plan to change this, we have only 2 days left to the release

vladpaiu:
did not have time to prioritize this, since it's labeled as a bug, maybe we can hotfix & backport later.

modules/b2b_logic/logic.c
```diff
@@ -476,23 +476,43 @@ void b2b_mark_todel( b2bl_tuple_t* tuple)
...
-			if (msg) {
-				get_local_contact(send_sock, NULL, local_contact);
-			} else {
-				LM_ERR("No current SIP message, "
-					"failed to build Contact from send socket\n");
-				return -1;
+			goto msg_contact;
+		} else {
+			/* validate what we have built */
+			if (parse_uri(local_contact->s, local_contact->len, &server_uri) < 0) {
```
bogdan-iancu, line 489: parsing the generated URI (based on server_address_pve) is an unnecessary burn - as script writer you need to be sure you configuration is correct and not to expect OpenSIPS to parse it (for each message) just to be sure you are right - eventually you can do a dummy test at startup, to see if the pve generates a valid URI

modules/b2b_logic/logic.c
```diff
@@ -476,23 +476,43 @@ void b2b_mark_todel( b2bl_tuple_t* tuple)
...
+					goto msg_contact;
+				}
+
+				memset(&ct_uri, 0, sizeof(struct sip_uri));
+				if (parse_uri(from_uri->s, from_uri->len, &ct_uri) < 0) {
+					LM_ERR("Not a valid FROM sip uri [%.*s]\n", from_uri->len, from_uri->s);
+					goto done;
+				}
+
+				get_local_contact(send_sock, &ct_uri.user, local_contact);
```
bogdan-iancu, line 508: the replacement of the username part (if contact_user is on) in the generated URI (based on server_address_pve) does not look right - you do not actually replace the username part, but build a completely new URI based on the username and socket - so if my server_address_pve was pushing some URI params or anything else, it will be lost. Normally you should have here a sctrict username replacement into the local_contact parsed URI - this may be simply done with 3 memcpy's .

bogdan-iancu (changes requested):
I just took the comments I added almost 1 year ago and compiled them into a `review`

PR #3318, opened 2024-02-26
Add remove() filtering by bflags

vladpaiu:
**Summary**
Add remove() filtering by bflags

**Details**
Sometimes it's useful to filter the removal of contacts for an AOR based on the internally saved branch flags per contact.

**Solution**
New param with the bflags

**Compatibility**
Backwards-compatible

### DIFF
--- a/modules/registrar/README
+++ b/modules/registrar/README
@@ -53,7 +53,7 @@ registrar Module
               1.4.1. save(domain[, flags[, aor[, ownership_tag]]])
 
               1.4.2. remove(domain, AOR[, [contact][, [next_hop][,
-                      [sip_instance]]]])
+                      [sip_instance], [bflags]]]])
 
               1.4.3. remove_ip_port(IP,Port, domain, [AOR])
               1.4.4. lookup(domain [, flags [, aor]])
@@ -158,9 +158,9 @@ Chapter 1. Admin Guide
      * off - stores the value of the Path headers into usrloc
        without passing it back to the UAC in the reply.
      * lazy - stores the Path header and passes it back to the UAC
-       if Path-support is indicated by the “path” param in the
+       if Path-support is indicated by the "path" param in the
        Supported HF.
-     * strict - rejects the registration with “420 Bad Extension”
+     * strict - rejects the registration with "420 Bad Extension"
        if there's a Path header but no support for it is indicated
        by the UAC. Otherwise it's stored and passed back to the
        UAC.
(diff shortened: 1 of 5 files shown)


bogdan-iancu (commented):
Hi @vladpaiu , just to be sure we are on the same page, the `bflags` new parameter actually contains the string name of a SINGLE bflag, right (as per code) and not several? If so, shouldn't you align the docs and variable name to reflect the fact we have only one bflag?

modules/registrar/README
liviuchircu, line 1: The `README` textfile is auto-generated & committed by @opensips-github weekly, there is no need to pollute the commit with such kind of diffs (_even more so_ when the diff is locale-related, of all things...) :wink:

modules/registrar/doc/registrar_admin.xml
```diff
@@ -611,6 +611,12 @@ save("location", , , "vip");
 				"+sip.instance" value to be used for filtering purposes.
 			</para>
 		</listitem>
+		<listitem>
+			<para>
+				<emphasis>blfags (string, optional)</emphasis> - a
+				Branch flags to be used for filtering purposes.
```
liviuchircu, line 617: typos:

* blfags
* a Branch flags?

modules/registrar/save.c
```diff
@@ -950,6 +951,11 @@ int _remove(struct sip_msg *msg, void *udomain, str *aor_uri, str *match_ct,
 		goto out_unlock;
 	}
 
+	if (bflags_s) {
+		bflags = get_flag_id_by_name(FLAG_TYPE_BRANCH, bflags_s->s, bflags_s->len);
+		bflags = 1<<bflags;
+	}
```
liviuchircu, line 957: There is some potential for improvement here:

1. this repetitive flag parsing shouldn't be done at runtime (**N** times), rather during the fixup phase (**1** time).  And even if the flags were to arrive via a variable, the function interface will detect this case and correctly invoke the fixup _per function call_.
2. if the original intent was to allow more flags, then `flag_list_to_bitmask()` should have been used instead of `get_flag_id_by_name()`.  Going down the rabbit hole (credits @bogdan-iancu):  should the multiple flags be AND'ed or OR'ed?  As long as the docs clarify this behavior, anything is OK.  I see the current implementation opted for "OR".

liviuchircu (changes requested):

modules/registrar/save.c
bogdan-iancu: Addressed via 7fce69b4429f806e3f91186ed86ae2765a517704

PR #3327, opened 2024-03-07
Add option "reply_to_via" to OpenSIPS

carstenbock:
**Summary**
This PR brings back the "reply_to_via" functionality from the original SER project. We have seen in several network environments that replying to the source of an IP packet sometimes may not work. However, we are able to reply to the IP indicated in the Via-Header. This is especially the case for Kubernetes environments.
I am aware that it is more of a bug on the network layer than in the application, and everyone would prefer to fix the networking instead. However, this is not always possible, which has caused some controversy about this parameter in the past.
This parameter allows the core to reply to the IP/Port indicated in the Via-Header, instead of to the source-IP/Port seen of the IP Packet.

**Details**
This PR adds the parameter to the core and utilizes this in the SL and TM modules.

**Compatibility**
By default, the current behavior is kept as it is, thus it causes no impact to existing deployments.


bogdan-iancu (commented):
Hi @carstenbock , interesting option here, thanks you.
A first feedback - just options are best to be controllable per request and not globally - as we have the `force_rport()` function instead of a global param.  So we may consider migrating this to a function + msg flag (similar to `FL_FORCE_RPORT`). 
Also this feature is very close the the `rport` stuff, they  both control what is to be considered as destination for the replies. And being in the same area, I'm brainstorming here the possibility to merge these 2 features into a single function (with parameters) like `config_replying("FORCE_RPORT|REPLY_TO_VIA")` kind of stuff....

bogdan-iancu:
@carstenbock , please see 1ab78f83666b976628d6cba4b09d7ce8cbb68351 where I merged your PR with more VIA related functionalities

PR #3329, opened 2024-03-07
tm: Add functionality to execute a route, whenever TM creates a request or reply

carstenbock:
**Summary**
We needed to run certain actions, whenever TM creates a request or response. This PR adds the functionality to execute a specific route for requests or replies.

**Details**
This PR adds two new parameters to the TM-Module to define routes, which will be executed upon locally generated requests or replies.

**Solution**
It comes handy, if you for example are using OpenSIPS as a B2BUA and you want to trigger certain actions upon the termination of a session, e.g. when responding to a BYE request.

**Compatibility**
It does not affect any existing deployments, it just adds new functionality.

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -68,6 +68,7 @@
 #include "../../dprint.h"
 #include "../../config.h"
 #include "../../parser/parser_f.h"
+#include "../../parser/msg_parser.h"
 #include "../../ut.h"
 #include "../../timer.h"
 #include "../../error.h"
@@ -119,7 +120,7 @@ static struct script_route_ref *goto_on_reply = NULL;
 /* currently processed branch */
 extern int _tm_branch_index;
 
-
+static struct sip_msg dummy_msg;
 
 /* returns the picked branch */
 int t_get_picked_branch(void)
@@ -442,6 +443,21 @@ static int _reply_light( struct cell *trans, char* buf, unsigned int len,
 
 	if(trans->uas.request && trans->uas.request->flags&tcp_no_new_conn_rplflag)
 		tcp_no_new_conn = 1;
+
+	if(ref_script_route_is_valid(tm_local_reply)) {
+		LM_DBG("Found Local-Reply Route...\n");
+		LM_DBG("Message:\n-----------------\n%.*s\n--------------------\n", len, buf);
+		memset(&dummy_msg, 0, sizeof(struct sip_msg));
+		dummy_msg.buf = buf;
+		dummy_msg.len = len;
+
+		if (parse_msg(buf, len, &dummy_msg) == 0) {
+			LM_DBG("Parsed Message, executing Local-Reply Route with Message...\n");
+			run_top_route(sroutes->onreply[tm_local_reply->idx], &dummy_msg);
+		}
+		free_sip_msg(&dummy_msg);
+	}
+
 
 	if ( SEND_PR_BUFFER( rb, buf, len )==0 ) {
 		LM_DBG("reply sent out. buf=%p: %.9s..., "
--- a/modules/tm/uac.c
+++ b/modules/tm/uac.c
@@ -76,7 +76,7 @@ int pass_provisional_replies = 0;
 
 /* T holder for the last local transaction */
 struct cell** last_localT;
-
+static struct sip_msg dummy_msg;
 
 /*
  * Initialize UAC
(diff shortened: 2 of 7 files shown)


bogdan-iancu (commented):
Could you please cleanup the PR as you have here diff related to `reply_to_via` :)

Also, there is the `local_route {}` route which is triggered for any TM generated request. Is here any difference when comes to `local_request_route` ?

carstenbock:
Hi,

I will do the cleanup ASAP - I'm on holiday next week. The cleanup also
applies to my other PR's.

The difference to "local_route" is that this is executed after all
callbacks have been executed. For example, the "rtp_relay" module executes
some functions based on TM-Callbacks. The changes from the module are not
visible in "local_route". I needed the modified SDP so I could use that
information in Diameter-Rx-Requests for QoS.

Thanks,
Carsten


Am Di., 19. März 2024 um 18:49 Uhr schrieb Bogdan Andrei IANCU <
***@***.***>:

modules/tm/uac.c
```diff
@@ -553,6 +553,19 @@ int t_uac(str* method, str* headers, str* body, dlg_t* dialog,
...
 
+	if (ref_script_route_is_valid(tm_local_request)) {
+		LM_DBG("Found Local-Request Route...\n");
+		memset(&dummy_msg, 0, sizeof(struct sip_msg));
+		dummy_msg.buf = request->buffer.s;
+		dummy_msg.len = request->buffer.len;
+
+		if (parse_msg(request->buffer.s, request->buffer.len, &dummy_msg) == 0) {
+			LM_DBG("Parsed Message, executing Local-Reply Route with Message...\n");
+			run_top_route(sroutes->request[tm_local_request->idx], &dummy_msg);
```
bogdan-iancu, line 564: when comes to this new local_request, I see it is triggered right after the existing local_route. I see nothing in between to justify one more route. You mentioned some callbacks (being executed between the triggering of the old and new local request route), but I fail to  see any.....

modules/tm/t_reply.c
```diff
@@ -443,6 +444,21 @@ static int _reply_light( struct cell *trans, char* buf, unsigned int len,
...
+		memset(&dummy_msg, 0, sizeof(struct sip_msg));
+		dummy_msg.buf = buf;
+		dummy_msg.len = len;
+
+		if (parse_msg(buf, len, &dummy_msg) == 0) {
+			LM_DBG("Parsed Message, executing Local-Reply Route with Message...\n");
+			run_top_route(sroutes->onreply[tm_local_reply->idx], &dummy_msg);
+		}
+		free_sip_msg(&dummy_msg);
+	}
```
bogdan-iancu, line 459: ok, this is something new, to have a route run when opensips generates (not proxying). Still, will this be triggered for any locally generated reply, like even for the 100 Trying?
Secondly, more confirming actually, this route is NOT intended for changing anything on the reply, but only for inspection purposes, right ?
And...my I ask for the typical example of using such reply route? :)

modules/tm/uac.c
carstenbock: I need to double-check, but I added it since the request passed to the locally executed route did not include all the changes applied to the message (like content changed from the rtp_relay module). This is why I explicitly parsed the newly created buffer instead of using the other route.

bogdan-iancu: OK, I think I understand what is happening here. When the `local_route` is executed, the buffer (with all the changes so far) is constructed, but BEFORE re-parsing (to get a `sip_msg`), the TMCB_LOCAL_REQUEST_OUT callbacks are executed. So the the changes generated by TMCB_LOCAL_REQUEST_OUT are not seen in `local_route`.. and rtp_relay uses this TMCB_LOCAL_REQUEST_OUT callback.. I will check to see if we can change the order and fix this.

carstenbock:
We need to get the final response being sent out, so we can pass this to
the PCRF (in our case using an Diameter-Microservice):

onreply_route[tm_local_reply] {
if (is_method("INVITE") && $rb(application/sdp) &&
t_check_status("183|200")) {
$json(request) := "{}";
(rb(application/sdp));
$json(request/sdp_reply) = $rb(application/sdp);
$json(request/callback) = "http://192.168.0.51:8888/mi";
$var(send_ct) = "application/json; charset=utf-8";

        # Creating a resource using a RESTful service with an HTTP POST request
        $var(rc) =
rest_post("http://192.168.0.51:8080/qos/$(<request>fU)/$b2b_logic.key/$avp(src_ip)",
                    $json_compact(request), $var(send_ct),
$var(recv_body), $var(recv_ct), $var(rcode));
    }

}

Am Mo., 6. Mai 2024 um 16:06 Uhr schrieb Bogdan Andrei IANCU <
***@***.***>:

bogdan-iancu:
HI @carstenbock ,  so a quick status:
1) on the request part, the issue right now is that the TMCB_LOCAL_REQUEST_OUT callback used by rtp_relay is called after the local route, so you cannot "see" in the local route the changes introduced by rtp_relay
2) on the reply part, this seems to be more a b2b need, rather than TM... still the route can run from both B2B or TM. And to keep consistency, as for requests we have the local route in TM, yes, we can stick to TM with the replies also. Eventually add an extra modparam to tell the min reply code to trigger the reply route (like if you want to skip it for 100 or any provisional replies).

bogdan-iancu:
Merged in 1248f80b976c29d10857c4fb3ce129a1740bde69 and completed via fa93ebbe65ab3fabdfa58a1c34e6f13f48635368

Thank you @carstenbock

PR #3352, opened 2024-03-29
db_text: new parameter: buffer_size

ovidiusas:
**Summary**
New parameter for db_text: buffer_size.
This will allow db_text to handle larger text files.

### DIFF
--- a/modules/db_text/dbt_file.c
+++ b/modules/db_text/dbt_file.c
@@ -89,7 +89,8 @@ int dbt_check_mtime(const str *tbn, const str *dbn, time_t *mt)
 dbt_table_p dbt_load_file(const str *tbn, const str *dbn)
 {
 	FILE *fin=NULL;
-	char path[512], buf[4096];
+	char path[512];
+	char *buf;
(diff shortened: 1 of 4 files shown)


modules/db_text/dbt_file.c
```diff
@@ -125,6 +126,12 @@ dbt_table_p dbt_load_file(const str *tbn, const str *dbn)
 	if(!fin)
 		return NULL;
 
+	buf = pkg_malloc(buffer_size);
```
bogdan-iancu, line 129: @ovidiusas , wouldn't be worthy to have this `buf` as static var in the function, so we can allocate it upon first usage and keep re-use it for the whole runtime? Instead of keep allocating and freeing it after each load....

ovidiusas:
@bogdan-iancu Any updates here?

modules/db_text/dbt_file.c
bogdan-iancu: agreed, point taken

ovidiusas:
Done!

Issue #3340, opened 2024-03-12
[BUG] tracer: unable to trace if FQDN is used as an advertised address for a socket


ovidiusas:
**OpenSIPS version you are running**
Latest master and 3.4.

**Describe the bug**
If an FQDN is used as an advertised address for a socket, tracing will fail inside pipport2su() from tracer.c with error:
```
LM_ERR("host <%.*s> is not an IP\n",host_uri.len,host_uri.s);
```

A simple way to reproduce this:
```
socket=[my_proto]:[my_ip] as [my_fqdn]
```
Then perform tracing inside the config.
This is a side effect of commit a13e034.

@vladpaiu, can you please take a look at this issue?
When we are using the `adv_address` for tracing we should check if it's an IP or not (using `str2ip` and `str2ip6`) and only after that to use it for tracing. If it's not an IP, then we should either default to non advertised IP or resolve the FQDN and use the resolved IP for tracing.

ovidiusas:
Or maybe we should try to resolve the FQDN inside pipport2su() before throwing the "is not IP" error.

ovidiusas:
Please check pull request #3352
It has a fix for this issue.

bogdan-iancu:
@ovidiusas , could you generate a new PR with the proposed fix here ?

bogdan-iancu:
@ovidiusas , I guess the fix you push is a complete fix for the reported issue here, right ? Just to know if we should close here too

ovidiusas:
All is good now. The backport is also good!

PR #3364, opened 2024-04-11
tracer: set proper IP instead of FQDN for locally generated requests and outgoing replies

ovidiusas:
Partial fix related to #3340 (fix for locally generated requests)


ovidiusas:
@bogdan-iancu, any updates here?

bogdan-iancu:
thanks @ovidiusas

PR #3371, opened 2024-04-22
db/schema: fix dbtext dispatcher definition

ovidiusas:
**Summary**
When using db_text as a backend database for dispatcher, `priority` and `description` should be allowed `null/empty` values.

### DIFF
--- a/db/schema/dispatcher.xml
+++ b/db/schema/dispatcher.xml
@@ -90,7 +90,8 @@
         <name>attrs</name>
         <type>string</type>
         <size>128</size>
-        <default/>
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
@ovidiusas , I see the dispatcher code properly handles the NULL case for both those columns, so why the change only for db_text and not for all db type? something special about db_text? or something special for the rest of them?

ovidiusas:
You are right. My initial thought was to fix only dbtext without affecting all the other databases.
From a consistency POV, it's better to have all of them in synch.
Commit updated.

PR #3390, opened 2024-05-17
A new string transformation

### LINKED ISSUE #3236: [FEATURE] Parse Date Header in Core
**Given that the Stir/Shaken (module) can be timestamp (Date header) sensitive, having the ability to parse the timestamp independent of the SS mod would be useful**

I would really like to have a function to parse the Date header and it to be returned as a unix timestamp, which then in the script-level could be used to:
1. Detect calls which have an incorrect timestamp
2. Measure delta-drift to help identify typical latency and abnormal latency.

**Implementation**
 
  - Component: core 
  - Type: function
  - Name: parse_date 

The only other way to do this at present is to pass the Date header externally for it to then be processed. I suppose it may be possible to extract this information via regex and rebuild it, but this would be very messy

**Additional context**

lemenkov:
**Summary**
A new string transformation function influenced by a similar function from stir-shaken module. Closes #3236.


lemenkov:
[test.cfg.txt](https://github.com/OpenSIPS/opensips/files/15351983/test.cfg.txt)

A minimal config=file to check how this works.

transformations.c
```diff
@@ -457,6 +472,45 @@ int tr_eval_string(struct sip_msg *msg, tr_param_t *tp, int subtype,
...
+
+			if (!strptime(_tr_buffer, "%a, %d %b %Y %H:%M:%S GMT", &date_tm)) {
+				LM_ERR("Failed to parse Date header field\n");
+				goto error;
+			}
+
+			tz = getenv("TZ");
+			setenv("TZ", "", 1);
+			tzset();
+			snprintf(_tr_buffer, TR_BUFFER_SIZE, "%ld", mktime(&date_tm));
```
liviuchircu, line 495: Let's change the printing to `%lld`, with cast to `(long long)` please, for ARM32 build.  See commit 25107836 for the reason.

liviuchircu (changes requested):
Hi, @lemenkov!  The `{s.date2unix}` transformation seems like a nice addition, especially since the SIP `Date` header format is fixed, RFC-1123 like, so there is no need to worry about supporting other possible formats.

I've highlighted some minor pending changes, after which I am good to merge this.

transformations.c
```diff
@@ -457,6 +472,45 @@ int tr_eval_string(struct sip_msg *msg, tr_param_t *tp, int subtype,
...
+			if (tz)
+				setenv("TZ", tz, 1);
+			else
+				unsetenv("TZ");
+			tzset();
```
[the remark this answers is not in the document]
lemenkov: Done.

transformations.c
lemenkov: Done

liviuchircu:
Thanks for the update, @lemenkov!  Cheers!

PR #3403, opened 2024-05-31
Fix `perl_exec_simple` memory leak

### LINKED ISSUE #3402: [BUG] Memory leak in `perl_exec_simple`

```
version: opensips 3.4.5 (x86_64/linux)
git revision: ffdb1b4
main.c compiled on  with gcc 8
```

**Describe the bug**

When using the `perl_exec_simple(...)` function from the `perl` module, there is an os-malloc memory leak that accumulates over time.  Specifically, each time that function is called, the arguments which are passed to Perl are never fully garbage-collected by the Perl interpreter.  I've opened a bug report directly against Perl as well, but it's unlikely to be fixed in the short-term, so OpenSIPS ought to apply a workaround.

**To Reproduce**

Create a simple perl file named `perl_leak.pl` containing this:
```
sub perl_leak {  return 1; }
1;
```
And an opensips config like this:
```
mpath="/usr/lib/x86_64-linux-gnu/opensips/modules/"
loadmodule "sl.so"
loadmodule "signaling.so"
loadmodule "perl.so"
modparam("perl", "modpath", "/usr/lib/x86_64-linux-gnu/opensips/perl")
modparam("perl", "filename", "perl_leak.pl")
stderror_enabled=yes
syslog_enabled=no
log_level=-1
disable_core_dump=no
max_while_loops=2147483647
loadmodule "proto_udp.so"
socket=udp:127.0.0.1:9999

####### Test Logic ########

timer_route[foo, 1] {
  route("perl_leak");
}

route[perl_leak] {
  $var(i) = 0;
  while($var(i) < 250000) {
    # perl_exec("perl_leak", "foo:bar");
    perl_exec_simple("perl_leak", "foo:bar");
    $var(i) = $var(i) + 1;
  }
}
```
Launch it in the background like this:
```
opensips -f perl_leak.cfg -w $(pwd) &
```
And observe the memory usage growing (e.g., with `top`).

Be sure to `killall opensips` before your machine runs out of memory.

**Expected behavior**

I expect OpenSIPS to call the perl function and return without leaking memory.

**Relevant System Logs**

 - Operating System: `Debian 10`
 - OpenSIPS installation: `debs`
 - other relevant information: `libperl5.28:amd64` version `5.28.1-6+deb10u1`

**Additional context**

The underlying Perl bug is reported here: https://github.com/Perl/perl5/issues/22255

In that bug report, I also demonstrate a simple workaround which can be applied from the application point of view.

The entrypoint from OpenSIPS into the perl subsystem is here: https://github.com/OpenSIPS/opensips/blob/3.4.5/modules/perl/perlfunc.c#L84

The `perl_exec` function does not have this issue since it is specific to the `call_argv` mechanism, which is only used by `perl_exec_simple`

From my tests, it appears that `perl_exec_simple` is actually slightly slower than `perl_exec`!  I'm not sure what the cause is -- probably some internal call-stack handling differences in the perl APIs, but an alternate long-term OpenSIPS fix might be to implement both functions via `call_pv`.  Probably not worth the complexity in a backported bugfix, since the performance difference is very small, but in OpenSIPS 3.5 or 3.6 maybe `perlfunc.c` should be refactored.

p120ph37:
**Summary**

Fix a memory leak in the `perl` module that can accumulate OS `malloc`'d memory over time.  The leak is fairly slow, so unless your configuration makes very heavy use of perl functions, it may go unnoticed for quite a while, but it can eventually result in OS memory exhaustion and the OOM-killer.

**Details**

`perl_exec_simple` has a memory leak because `call_argv` doesn't garbage-collect its arguments properly.  This is technically a Perl bug, and is tracked here: https://github.com/Perl/perl5/issues/22255 But, until it is fixed in Perl, we should work around the issue from the OpenSIPS side.

See #3402 for more details on how to reproduce the bug in OpenSIPS and how to validate the fix.

**Solution**

This PR fixes the problem by applying the appropriate Perl garbage-collection context markers around the function call so that garbage collection of the arguments happens as control returns to the caller.

**Compatibility**

No functional change to OpenSIPS.

**Closing issues**

closes #3402

### DIFF
--- a/modules/perl/perlfunc.c
+++ b/modules/perl/perlfunc.c
@@ -81,7 +81,11 @@ int perl_exec_simple(struct sip_msg* _msg, str *_fnc_s, str *_param_s)
 	if (perl_checkfnc(fnc)) {
 		LM_DBG("running perl function \"%s\"\n", fnc);
 
+		ENTER;
+		SAVETMPS;
 		call_argv(fnc, flags, args);
+		FREETMPS;
+		LEAVE;
 		ret = 1;
 	} else {
 		LM_ERR("unknown function '%s' called.\n", fnc);


liviuchircu:
Hi @p120ph37!  I took a quick look at the `push_scope()`, `pop_scope()` internals, coupled w/ `Perl_savetmps()` and `Perl_free_tmps()` and they seem to be quite harmless - if adding this scope management fixes the leak for you, I am more than happy to merge & backport all the way down to `3.2`.  Cheers!

PR #3407, opened 2024-06-11
uac_registrant: make force_register run force_register instead of disable

### LINKED ISSUE #3395: [BUG] force_register acts like reg_disable
Sorry, I don't have time to do a proper PR right now, but I noticed the `reg_force_register` implementation accidentally calls the `run_mi_reg_disable` callback instead of `run_mi_reg_force_register` callback.

See https://github.com/OpenSIPS/opensips/blob/master/modules/uac_registrant/registrant.c#L1537

This means "reg_force_register" doesn't work and acts exactly the same as "reg_disable".

Btw "reg_force_register" is a very good and useful addition! (With the correct callback). Thank you.

jes:
**Summary**
I noticed the `reg_force_register` implementation accidentally calls the `run_mi_reg_disable` callback instead of `run_mi_reg_force_register` callback.

**Details**

This means "mi reg_force_register" acts the same as "mi reg_disable" - it disables the user instead of registering it.

**Solution**

My PR fixes the problem by correcting the callback that is invoked.

**Compatibility**

This will break any scenarios that depend on "mi reg_force_register" disabling the user instead of forcing a re-registration of the user.

**Closing issues**

Closes #3395.

### DIFF
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -1534,7 +1534,7 @@ static mi_response_t *mi_reg_force_register(const mi_params_t *params,
 
 	lock_get(&reg_htable[hash_code].lock);
 	rc = slinkedl_traverse(reg_htable[hash_code].p_list,
-					&run_mi_reg_disable, &coords, NULL);
+					&run_mi_reg_force_register, &coords, NULL);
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
Thanks @jes

jes:
Thanks!

PR #3410, opened 2024-06-13
RTPEngine - Last frame pos

NormB:
**Summary**

This PR will allow the script writer to optionally store the last-frame-pos when the rtpengine_stop_media() function is called.

RTPEngine returns the last-frame-pos with the last frame played from a previously executed play media command.

**Details**

This PR adds an optional parameter to the rtpengine_stop_media() function.

For example: rtpengine_stop_media("from-tag=$tt", ,$avp(last_frame_pos));

**Solution**

This PR uses the same logic from the rtpengine_play_media() function. This function stores data returned by RTPEngine into a "duration" variable.

**Compatibility**

There should be no compatibility issues because the existing rtpengine_stop_media() function does not handle any values returned by RTPEngine.

**Closing issues**


razvancrainea:
Thanks for your contribution!
Can you please move the parameter *after* the sock var, to be backwards compatible?

Issue #3411, opened 2024-06-14
[BUG] fraud_detection never release record for every user and every number


okhowang:
**OpenSIPS version you are running**

```
```

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Relevant System Logs**

**OS/environment information**
 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

**Additional context**

okhowang:
And matched_len may not be initialized when number has char except 0~9
It may cause unexpected behaviour.
https://github.com/OpenSIPS/opensips/blob/master/modules/fraud_detection/fraud_detection.c#L281

bogdan-iancu:
Could you detail a bit more on the leak you found? like what are the leaking structure and the case when the leak happens?

okhowang:
`fraud_detection` save every user's statistic in memory forever (util shutdown).
if there are 1,000,000 user in a system. it will has 1,000,000 record kept in memory.
even there only 1000 user call in one day.
maybe we should clear statistic every day/week.


I found this problem because, my service OOM.
and I notice `map_get` use a lot of mem. which is called by `fraud_detection`. and a lot of entries cost 64k memory.
it because uninitialized `matched_len` which make normal user's string become a huge random string.

uninitialized `matched_len` and unclearable statistic cause the OOM.

liviuchircu:
Thank you for the investigation work and your patch, @okhowang!  I applied a slightly better fix and closed #3413, could you pull it and check if the "leak" persists?  While I am fully aware that the fraud-detection stores `(user, prefix)` tuples, it is not clear as to what degree they will fill up the SHM pool.  Questions such as the following could help us get a better idea for SHM pool sizing in your case, or maybe even develop a "stat timer" feature for the module:

A) is the calling user pool finite (e.g. from `subscriber` table) or random (e.g. simply use `$fU`)?
B) does the majority of users typically dial random numbers frequently (calling center), or is it a residential service?

Cheers!

okhowang:
> I applied a slightly better fix and closed https://github.com/OpenSIPS/opensips/pull/3413, could you pull it and check if the "leak" persists?

thank you. But I'm still at 3.2 branch. I will upgrade to 3.4 LTS later. So, feedback may be delayed for weeks.
But I think it should work as expected.

> A) is the calling user pool finite (e.g. from subscriber table) or random (e.g. simply use $fU)?

It is finite. We use `subscriber` table currently.

[...]

PR #3413, opened 2024-06-15
fraud_detection: init matched_len

okhowang:
related https://github.com/OpenSIPS/opensips/issues/3411

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/fraud_detection/fraud_detection.c
+++ b/modules/fraud_detection/fraud_detection.c
@@ -278,7 +278,7 @@ static int check_fraud(struct sip_msg *msg, str *user, str *number, int *pid)
 
 	/* Find a rule */
 
-	unsigned int matched_len;
+	unsigned int matched_len = 0;
 	lock_start_read(frd_data_lock);
 	rt_info_t *rule = drb.match_number(*dr_head, *pid, number, &matched_len);
 


modules/fraud_detection/fraud_detection.c
```diff
@@ -278,7 +278,7 @@ static int check_fraud(struct sip_msg *msg, str *user, str *number, int *pid)
 
 	/* Find a rule */
 
-	unsigned int matched_len;
+	unsigned int matched_len = 0;
 	lock_start_read(frd_data_lock);
 	rt_info_t *rule = drb.match_number(*dr_head, *pid, number, &matched_len);
```
bogdan-iancu, line 283: the `match_number` function does not populate `matched_len` if error (no match) is returned. This is correct. But in this case `rule` will be NULL and the `check_fraud` will exit (see bellow). 
So I don't see the case where `matched_len` may be used in-init'ed.

okhowang: `match_number` call `find_rule_by_prefix_unsafe`.
in `find_rule_by_prefix_unsafe`, if `get_prefix` got NULL, it will try `check_rt` for prefixless rules again.
`matched_len` is not set by the branch. and we got `rule` in the branch.

PR #3420, opened 2024-06-21
Update b2b_logic_admin.xml example syntax

NormB:
Added parens around b2b_logic.entities to correct the examples.

**Summary**

This PR updates the examples which contain b2b_logic.entity so that they do not cause syntax errors when copy/pasted.

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/b2b_logic/doc/b2b_logic_admin.xml
+++ b/modules/b2b_logic/doc/b2b_logic_admin.xml
@@ -812,7 +812,7 @@ route[b2b_logic_request] {
 ...
 route[b2b_logic_reply] {
    ...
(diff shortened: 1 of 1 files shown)


NormB:
?

razvancrainea:
Not sure why you need to always specify `[0]` index - you mean it does not work without it?

NormB:
> Not sure why you need to always specify `[0]` index - you mean it does not work without it?

It's been a little while since this was opened but as I recall, a syntax error was thrown using the current example.

NormB:
Norm Brandinger
(c) 973-479-5595

PR #3424, opened 2024-07-01
fix missing headers if t_reply_body is used inside a failure route block

### LINKED ISSUE #3422: [BUG] append_to_reply not working when t_reply_with_body is used inside a failure_route block

```
[root@box ~]# opensips -V
version: opensips 3.4.5 (x86_64/linux)
git revision: 6beb906e5
main.c compiled on 01:49:20 Apr 19 2024 with gcc 8
```

**Describe the bug**
append_to_reply not working when t_reply_with_body is used inside a failure_route block. Only the body is present in the reply.

**To Reproduce**

```
failure_route[local_reply] {
    t_reply_with_body(404, "Not Found", "a b c"); 
    append_to_reply("Foo: bar\r\n");
```

**Expected behavior**

A sip reply with Foo: bar header and a body with content "a b c".

**Relevant System Logs**
no errors found in the opensip log.

 - Operating System: Almalinux 8
 - OpenSIPS installation: official rpm packages

A similar issue was [fixed](https://github.com/OpenSIPS/opensips/issues/1632) in the past for request route

Riccardo-78:
**Summary**

This PR fixes append_to_reply not working when t_reply_with_body is used inside a failure_route block

**Details**

see linked issue

**Solution**

added update_cloned_msg_from_msg to the failure_route case.

**Compatibility**

**Closing issues**

closes #3422

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -1865,6 +1865,7 @@ int w_t_reply_body(struct sip_msg* msg, unsigned int* code, str *text,
 				LM_BUG("no transaction found in Failure Route\n");
 				return -1;
 			}
+			update_cloned_msg_from_msg( t->uas.request, msg);
 			lock_replies = 0;
 			break;
 		case REQUEST_ROUTE:


Riccardo-78:
I was wondering if there’s anything holding back the merge of my pull request? Let me know if I can assist with anything!

bogdan-iancu:
Hi @Riccardo-78 - thanks for the PR here. Nevertheless, I'm not convinced this is the right fix. If we do `update_cloned_msg_from_msg` in the failure route, then all the changes you did in failure route will stick the transaction for *all* the future branches, not only for the current branch. Which is not the intended behavior (in failure route, your changes impact only the branches you create at that very moment, not all the future ones).
I think it is a deeper problem with the  `t_reply_with_body` function - as [the comment here](https://github.com/OpenSIPS/opensips/blob/master/modules/tm/t_reply.c#L1873), the problem is that the function takes as param the request from the transaction, not the current sip_msg provided from the script level - not sure why, I can check on this

PR #3425, opened 2024-07-01
Fix segfault in siprec srs failover

### LINKED ISSUE #3421: [BUG]  Using SIPREC module, while first SRS doesn't respond with 200ok, Opensips encounters with segmentation fault
```
version: opensips 3.6.0-dev (x86_64/linux)
git revision: 694349908
main.c compiled on 13:29:09 Jun 19 2024 with cc 11
```

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Relevant System Logs**

 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

**Additional context**
[opensips_test.zip](https://github.com/user-attachments/files/15950280/opensips_test.zip)

Fr-Soltanzadeh:
**Summary**

Using SIPREC module, while first SRS doesn't respond with 200ok, Opensips encounters with segmentation fault. This bug is solved. Furthermore, rtp data are not recorded in the second SRS because some of the session data are removed after failure of the first one.  This bug is fixed too.

**Details**

Explained in issue 3421

**Solution**

Explained in issue 3421

**Compatibility** 

It seems that this change will not break other scenarios. As srec_logic_destroy function is called at the end of the call, memory leakage is not a problem.

**Closing issues**

closes #3421

### DIFF
--- a/modules/siprec/siprec_body.c
+++ b/modules/siprec/siprec_body.c
@@ -394,7 +394,7 @@ int srs_build_body(struct src_sess *sess, str *sdp, str *body)
 	SIPREC_COPY_STR(sdp_content_type, &buf);
 	SIPREC_COPY(CRLF, &buf);
 
-	if (sdp)
+	if (sdp->s && sdp->len > 0)
 		SIPREC_COPY_STR((*sdp), &buf);
 
 	/* add second bondary */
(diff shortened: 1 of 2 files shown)


razvancrainea:
Thanks for your report and investigation! However, the fix is not complete - the main problem is indeed that the SDP is not present, because it had been released (wrongfully) in the previous attempt. However, the destroy should be done, as a new B2B entity should be created.
I will soon push a proper fix.

Issue #2853, opened 2022-06-09
[BUG] uac_registrant hard-coded to use expires path param instead of Expires header


sreeram-narayanan:
**OpenSIPS version you are running**

```
version: opensips 3.2.6 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: ce8cda891
main.c compiled on 19:27:07 May 31 2022 with gcc 6.3.0
```

**Describe the bug**

In uac_registrant, the module is hard-coded to use "expires" as a path param in the Contact header. This prevents us from setting "Expires" header.

```

	if (1) {
		/* adding exiration time as a parameter */
		memcpy(p, expires_param.s, expires_param.len);
		p += expires_param.len;
	} else {
		/* adding exiration time as a header */
		memcpy(p, CRLF, CRLF_LEN); p += CRLF_LEN;
		memcpy(p, expires_hdr.s, expires_hdr.len);
		p += expires_hdr.len;
	}

```

**To Reproduce**

uac_registrant always uses expires path param instead of Expires header.

**Expected behavior**

This can be made configurable by using a mod_param.

**Relevant System Logs**

Not relevant.

okhowang:
According https://www.rfc-editor.org/rfc/rfc2543#section-4.2.6
> A server MAY choose any duration for the registration lifetime.
   Registrations not refreshed after this amount of time SHOULD be
   silently discarded. Responses to a registration SHOULD include an
   Expires header ([Section 6.20](https://www.rfc-editor.org/rfc/rfc2543#section-6.20)) or expires Contact parameters ([Section](https://www.rfc-editor.org/rfc/rfc2543#section-6.13)
   [6.13](https://www.rfc-editor.org/rfc/rfc2543#section-6.13)), indicating the time at which the server will drop the
   registration. If none is present, one hour is assumed. Clients MAY
   request a registration lifetime by indicating the time in an Expires
   header in the request. A server SHOULD NOT use a higher lifetime than
[...]

PR #3429, opened 2024-07-05
uac_registrant: using Expires header if expires is absent in contact

okhowang:
**Summary**
using Expires header if expires is absent in contact

**Details**
in https://www.rfc-editor.org/rfc/rfc2543#section-4.2.6
both `Expires` header and `expires` parameter in contact can determine the duration for REGISTER.
but `uac_registrant` only use `expires` parameter in contact only.
it cause expires may work as expected if server response with only `Expires` header.

**Solution**
if `expires` in contact is not found, try to use `Expires` header in response.

**Compatibility**
there is no compatibility problem.

**Closing issues**
closes https://github.com/OpenSIPS/opensips/issues/2853

### DIFF
--- a/modules/uac_registrant/registrant.c
+++ b/modules/uac_registrant/registrant.c
@@ -35,6 +35,7 @@
 #include "../../parser/parse_authenticate.h"
 #include "../../parser/contact/parse_contact.h"
 #include "../../parser/parse_min_expires.h"
+#include "../../parser/parse_expires.h"
 #include "../uac_auth/uac_auth.h"
 #include "../../lib/digest_auth/digest_auth.h"
 #include "reg_records.h"
@@ -362,6 +363,7 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 	str msg_body;
 	int statuscode = 0;
 	unsigned int exp = 0;
+	exp_body_t *header_exp = NULL;
 	unsigned int bindings_counter = 0;
 	reg_record_t *rec = (reg_record_t*)e_data;
 	struct hdr_field *c_ptr, *head_contact;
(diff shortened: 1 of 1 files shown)


modules/uac_registrant/registrant.c
```diff
@@ -518,6 +520,13 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 				contact = contact->next;
 			}
 		}
+		// if there is no expires in contact, try parse expires from header
+		if (exp == 0 && msg->expires) {
+			header_exp = (exp_body_t*)msg->expires->parsed;
```
bogdan-iancu, line 525: Even if the `Expires` hdr was (if exists) found (by parse_headers + EOH), the payload/body of the hdr is not automatically parsed, so `int parse_expires(struct hdr_field* _h)` should be called here.

modules/uac_registrant/registrant.c
```diff
@@ -518,6 +520,16 @@ int run_reg_tm_cback(void *e_data, void *data, void *r_data)
 				contact = contact->next;
 			}
 		}
+		// if there is no expires in contact, try parse expires from header
+		if (exp == 0 && msg->expires) {
+			if (parse_expires(msg->expires) < 0) {
+				LM_ERR("failed to parse Expires header\n");
+			}
+			header_exp = (exp_body_t*)msg->expires->parsed;
+			if (header_exp && header_exp->valid) {
```
bogdan-iancu, line 529: LGTM, but you could improve here by avoiding the double testing and have all the ops with `header_exp` on the `else` branch of `if (parse_expires(msg->expires) < 0) {`

bogdan-iancu:
Thank you @okhowang

PR #3433, opened 2024-07-08
Bug fix compare 0 vs NULL

bhoeneis:
**Summary**
Bug fix compare 0 vs NULL


bogdan-iancu:
`0` and `NULL` are exactly the same for pointers in C, this is not a bug

PR #3438, opened 2024-07-22
tls_openssl: fix openssl_tls_async_connect

jes:
The `fd` value here is used with `poll()`/`select()` in the event that `SSL_connect()` gives an `SSL_ERROR_WANT_WRITE` error. So we need to be polling/selecting on the fd that is used by the underlying socket, which is the fd passed as an argument to the function, rather than `con->fd`.

So the fix is to let `fd` keep the value that was passed in the function argument rather than overwrite it with `con->fd`.

It seems to me that `handle_io()` updates `con->fd` to match the received fd number only when the socket wants reading and not when it wants writing. When the socket wants writing the correct fd to use is the one that is passed in the function argument, not the one that is in `con->fd`.

The reasons this mostly worked before are:

1.  on a box that is not heavily-loaded you often end up with the same worker setting up the TLS session as opened the socket in the first place, so the fd in `con->fd happens to be correct.
2.  if `SSL_connect()` happily connects straight away, without an `SSL_ERROR_WANT_WRITE`, then you never run into a code path that uses the value in `fd` anyway.

**Summary**
This fixes an issue we were seeing where TLS connections would sometimes fail, particularly on highly-loaded boxes.

**Details**
This code has been there since `openssl_tls_async_connect()` was first created ( https://github.com/OpenSIPS/opensips/commit/806221f8ede4b063885df1aa73e895a60b1724f5 ) so I'm surprised nobody else has noticed this problem. (Could the issue actually be something else? Should `handle_io()` be setting `con->fd` in the write path as well? But then why pass it as a function argument? I don't fully understand this).

Empirically this fix works.

**Compatibility**
I don't think this change will break any scenarios.

### DIFF
--- a/modules/tls_openssl/openssl_conn_ops.c
+++ b/modules/tls_openssl/openssl_conn_ops.c
@@ -657,7 +657,6 @@ int openssl_tls_async_connect(struct tcp_connection *con, int fd,
 	poll_err=0;
 	elapsed = 0;
 	to = timeout*1000;
-	fd = con->fd;
 
 #if defined(HAVE_SELECT) && defined(BLOCKING_USE_SELECT)
 	FD_ZERO(&orig_set);


jes:
No updates here.

liviuchircu:
Great detective work here, @jes!  Indeed, it looks like that's how the function was initially written, and the bug simply slipped unnoticed.  And, yes, it's not clear why [handle_io()](https://github.com/OpenSIPS/opensips/blob/master/net/net_tcp_proc.c#L191) carefully takes care of converting the `s` (socket) from remote process to local process equivalent, only to set the `con->fd` only on `IO_WATCH_READ`, but not `IO_WATCH_WRITE`.

However, the TCP [.write() API](https://github.com/OpenSIPS/opensips/blob/master/net/api_proto_net.h#L38) does include the `fd` explicitly as argument and implementations such as [tcp_async_write()](https://github.com/OpenSIPS/opensips/blob/master/net/tcp_common.c#L370) do make exclusive use of this socket when performing their writes, so I believe we have enough info to draw a conclusion here.

Let's just patch `openssl_tls_async_connect()` to also make use of the argument `fd` until proven contrary, especially since you've noticed visible improvements on the patch. The patch is not changing anything on the state of the `con`, so impact is minimal.  It only fixes a async-connect edge-case, whenever the Connect requires an extra `SSL_ERROR_WANT_WRITE` stage -- and there, we pass the correct `fd`.

Merging and backporting to 3.4.

liviuchircu (approved):
LGTM

PR #3444, opened 2024-08-19
Update stir_shaken.c - enforce stir_shaken_verify pvar generic error reason

Integration-IT:
- more explicit SET_VERIFY_ERR_VARS.
- stir_shaken_verify generic err_reason pvar is now more explicit to easily reflect the real reason.

- replace generic "Internal Server Error" by dedicated internal error like "Failed to parse headers", "Failed to parse identity header", "Failed to get current time", "Failed to get UNIX time from Date header", "Failed to load certificate", "Error validating certificate", "Error while verifying signature".


liviuchircu:
Good idea, @Integration-IT!  A direct improvement over the previous "500 Internal Server Error" reply on all error cases, which will help troubleshoot the AS or VS more efficiently.  I will also backport down to `3.4`.

PR #3445, opened 2024-08-20
backport the security patch of CVE-2023-28097

Crispy-fried-chicken:
Here is a vulnerability which is fixed in the master branch(https://github.com/OpenSIPS/opensips/commit/7cab422e2fc648f910abba34f3f0dbb3ae171ff5) but is not fixed in the branch of v3.0, maybe it should be backported?


bogdan-iancu:
@Crispy-fried-chicken , the 3.0 version is not longer maintained (EOL) ...since 2020, please see https://opensips.org/About/AvailableVersions. So no more fixes for it.
You should rather try to migrate to a new version.

Issue #1948, opened 2020-01-25
[CRASH] called on a not free-able lump when using async API


lemenkov:
Hello,
I'm still working on my module which is using async API and got another one crash. Unfortunately I could not create a small config reproducer so all I have is this backtrace (which looks very similar to #1683).

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fd78c8cf801 in __GI_abort () at abort.c:79
#2  0x0000559720552be6 in free_lump (lmp=0x7fd78a08aa80) at data_lump.c:441
#3  0x0000559720552ceb in free_lump_list (l=0x7fd78a08aa48) at data_lump.c:470
#4  0x00005597205531a8 in del_flaged_lumps (lump_list=0x7fd78946ecb8 <faked_req+2040>, flags=LUMPFLAG_BRANCH) at data_lump.c:574
#5  0x00007fd78924cbf8 in post_print_uac_request (request=0x7fd78946e4c0 <faked_req>, org_uri=0x7fffacabab80, org_dst=0x7fffacabab70, body_clone=0x7fd78baf4770) at t_fwd.c:254
#6  0x00007fd78924d617 in add_uac (t=0x7fd78a100350, request=0x7fd78946e4c0 <faked_req>, uri=0x7fffacabab80, next_hop=0x7fffacabab70, bflags=64, path=0x7fd78946e738 <faked_req+632>, proxy=0x7fd78baf5548) at t_fwd.c:472
#7  0x00007fd78924e502 in t_forward_nonack (t=0x7fd78a100350, p_msg=0x7fd78946e4c0 <faked_req>, proxy=0x0, reset_bcounter=1, locked=1) at t_fwd.c:746
#8  0x00007fd78923050f in w_t_relay (p_msg=0x7fd78946e4c0 <faked_req>, flags=0x0, proxy=0x0) at tm.c:1253
#9  0x000055972056cfcd in do_action (a=0x7fd78ba1f6e0, msg=0x7fd78946e4c0 <faked_req>) at action.c:812
#10 0x000055972056a2b8 in run_action_list (a=0x7fd78ba1f6e0, msg=0x7fd78946e4c0 <faked_req>) at action.c:157
#11 0x00005597205b16b5 in eval_elem (e=0x7fd78ba1f838, msg=0x7fd78946e4c0 <faked_req>, val=0x0) at route.c:588
[...]

PR #3447, opened 2024-08-22
Only update the TM UAS if we are in request context and not fake request

vladpaiu:
**Summary**
Fix crashing of OpenSIPS when script writer mistakenly runs async in branch route and/or event route

**Details**
In case of push notifications injecting branches, and if in branch_route the script writer does async() calls, OpenSIPS will update the UAS with information from the fake request ( which might contain new branch lumps and other changes ). In case there are multiple branches, post_print_uac_request will try to free up the new request and find old SHM branches and crash in 

        /* delete inserted branch lumps */
        del_flaged_lumps( &request->add_rm, LUMPFLAG_BRANCH);
        del_flaged_lumps( &request->body_lumps, LUMPFLAG_BRANCH);

**Solution**
Only update the UAS if we are running in request route and we are not having a fake request in hand.

**Compatibility**
Backwards compatible

### DIFF
--- a/modules/tm/async.c
+++ b/modules/tm/async.c
@@ -249,7 +249,7 @@ int t_handle_async(struct sip_msg *msg, struct action* a,
 	} else {
 		/* update the cloned UAS (from transaction)
 		 * with data from current msg */
-		if (t->uas.request)
+		if ((t->uas.request) && (route_type==REQUEST_ROUTE) && ((msg->msg_flags & FL_TM_FAKE_REQ) == 0))
 			update_cloned_msg_from_msg( t->uas.request, msg);
 	}
 


modules/tm/async.c
```diff
@@ -249,7 +249,7 @@ int t_handle_async(struct sip_msg *msg, struct action* a,
 	} else {
 		/* update the cloned UAS (from transaction)
 		 * with data from current msg */
-		if (t->uas.request)
+		if ((t->uas.request) && (route_type==REQUEST_ROUTE) && ((msg->msg_flags & FL_TM_FAKE_REQ) == 0))
```
ar45, line 252: Should this be just `if ((t->uas.request) && !(msg->msg_flags & FL_TM_FAKE_REQ))` ?

ar45:
Still active.

modules/tm/async.c
liviuchircu: While it _seems_ we could enable more scenarios by omitting the `(route_type==REQUEST_ROUTE)` filtering as you suggest (after all, we only need `!is_faked_req()` here) , the script's async calls are not designed (nor much tested) beyond the `request_route`.  This design choice can even be seen in the same function, a [few lines below](https://github.com/OpenSIPS/opensips/blob/master/modules/tm/async.c#L321), where any route apart from `REQUEST_ROUTE` is simply handled as sync, to be on the safe side.  The restriction is also documented on the opensips.org wiki, [here](https://www.opensips.org/Documentation/Script-Async-3-6#async_routes).

So, following the same principle, I suggest we keep the filter, to avoid potential issues with non-faked requests from non-request routes.  Maybe an `async()` in `local_route` or `event_route` falls in this category?!

liviuchircu (approved):
LGTM, scripting mistakes should not cause a segfault!

lemenkov:
Fixed #3447.

lemenkov:
Fixed #1948.

PR #3448, opened 2024-08-22
Fix comparision of SIP URIs, add GR and PN related params checks

vladpaiu:
**Summary**
In the case of multiple push notification devices registering same URI@HOST but with different PN related params, OpenSIPS was not comparing the URIs properly and overriding old registrations with new ones.

**Solution**
Do proper SIP URI comparison.

**Compatibility**
Backwards compatible.


liviuchircu:
Nice catch, @vladpaiu!  Interesting to note that the new params made it into `print_uri()`, but not into `compare_uris()`...

PR #3453, opened 2024-08-29
Telecom 10237

davidtrihy-genesys:
When using the topology_hiding module with transaction only, the topology_hiding_match function will never work when using `set_advertised_address` this is because when a request comes in and the match function is called it makes sure that the host in the request domain matches one of the sockets we are listening on, due to the nature of the change where we now use the local_trunk_identifier prepended to the DNS address byoc.use1.genesys.cloud we will not be advertising our address or putting socket aliases but rather changing the advertised address on a message by message basis and this causes sequentials to fail the topology_hiding_match, this bypasses this check and the topology_hiding_match now matches on the thinfo param only


davidtrihy-genesys:
I opened the PR against the wrong repo apologies

davidtrihy-genesys:
Closing

PR #3457, opened 2024-09-02
Fix HP_MALLOC build flag - DBG_MALLOC is not necessary for it

Shkiperon:
**Summary**
Fix building of OpenSIPS with HP_MALLOC

**Details**
I found that if I doesn't want turn on DBG_MALLOC flag, but I want do some experiments with HP_MALLOC when building OpenSIPS from source - I cannot do that, because in that case shm_hash_usage will be undefined.

**Solution**
Move definition of shm_hash_usage to the ifdef block of HP_MALLOC.

**Compatibility**
This change doesn't looks like breaking.

**Closing issues**
No any opened issue.

### DIFF
--- a/mem/shm_mem.h
+++ b/mem/shm_mem.h
@@ -301,6 +301,7 @@ inline static void shm_threshold_check(void)
 				lock_release(mem_lock); \
 		} while (0)
 	#endif
+	extern unsigned long long *shm_hash_usage;
 #else
 #define shm_lock()    lock_get(mem_lock)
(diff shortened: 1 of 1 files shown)


liviuchircu:
Good catch, @Shkiperon -- the `shm_hash_usage` global definitely tied to the `mem/hp_malloc*` codebase, rather than `DBG_MALLOC`, as per the current code.  Merging and backporting right away -- it also fixes the issue for me too.

PR #3458, opened 2024-09-04
Add support for subdomain checking in aliases and socket aliases

### LINKED ISSUE #1466: loose_route doesn't check to determine if Record-Route is myself
If Record-Route is set to a hostname that DNS resolves to an IP that OpenSIPs is configured to listen on, and that hostname is not listed as an alias, loose_route will route the transaction to itself causing the call to fail as the is_myself("$si") will return true when checking for loops in addition to other problems.

My issue is that I am using dynamic hostnames which are merely sub-domains pointing to my OpenSIPs server.  As a work around, I tried to set alias=*.mydomain.com, but OpenSIPs will fail to start.  

Next, I tried to use set_advertised_address and specify the $td, but my call still fails as it does not have the same affect as specifying the hostname in alias.  If I set the hostname in alias on startup, everything works fine.

If there a function that I can set alias dynamically as a work around?  A solution would be to allow alias values to be wildcards.

davidtrihy-genesys:
**Summary**
Adds support for subdomains in aliases and socket aliases

**Details**
New feature related to this issue https://github.com/OpenSIPS/opensips/issues/1466

When defining sockets or aliases we can mark them with subdomain so that when check_self is called it will match when there the host to check is subdomain of the alias, definition of the socket and alias as so

```
socket = udp:127.0.0.1:5060 as my.subdomain.com subdomain

or

alias = udp:my.subdomain.com:5060 subdomain
```

If either are set up this way then when check_self is called for example in the TM topology_hiding then if the request domain is something like `resource.my.subdomain.com` or `different.resource.my.subdomain.com` it will match the alias but also a deliberate design decision to allow `my.subdomain.com` to match too.

**Solution**
A flag is added, when the `subdomain|SUBDOMAIN` flag is on the socket or alias this allows it to match addresses that match that alias subdomain while preserving strict matching when the flag is not set

**Compatibility**
The flag is a new flag so existing config where the subdomain flag is not on the alias or socket will behave like it already does without this change, it's backwards compatible

**Closing issues**
`closes #1466`

### DIFF
--- a/modules/domain/README
+++ b/modules/domain/README
@@ -14,6 +14,7 @@ domain Module
               1.3.3. domain_table (string)
               1.3.4. domain_col (string)
               1.3.5. attrs_col (string)
+              1.3.6. accept_subdomain_col (string)
 
         1.4. Exported Functions
 
@@ -57,9 +58,10 @@ domain Module
    1.3. Setting domain_table parameter
    1.4. Setting domain_col parameter
    1.5. Setting attrs_col parameter
-   1.6. is_from_local usage
-   1.7. is_uri_host_local usage
-   1.8. is_domain_local usage
+   1.6. Setting accept_subdomain_col parameter
+   1.7. is_from_local usage
+   1.8. is_uri_host_local usage
+   1.9. is_domain_local usage
 
 Chapter 1. Admin Guide
 
@@ -139,6 +141,15 @@ modparam("domain", "domain_col", "domain_name")
    Example 1.5. Setting attrs_col parameter
 modparam("domain", "attrs_col", "attributes")
 
+1.3.6. accept_subdomain_col (string)
+
+   Name of column containing accept_subdomain value in domain table.
+
+   Default value is “accept_subdomain”.
+
+   Example 1.6. Setting accept_subdomain_col parameter
+modparam("domain", "accept_subdomain_col", "accept_subdomain")
+
 1.4. Exported Functions
 
 1.4.1. is_from_local([attrs_var])
@@ -151,7 +162,7 @@ modparam("domain", "attrs_col", "attributes")
 
    This function can be used from REQUEST_ROUTE.
 
-   Example 1.6. is_from_local usage
+   Example 1.7. is_from_local usage
 ...
 if (is_from_local()) {
         ...
@@ -177,7 +188,7 @@ if (is_from_local($var(attrs))) {
    This function can be used from REQUEST_ROUTE, FAILURE_ROUTE,
    BRANCH_ROUTE.
 
-   Example 1.7. is_uri_host_local usage
+   Example 1.8. is_uri_host_local usage
 ...
 if (is_uri_host_local()) {
         ...
@@ -209,7 +220,7 @@ if (is_uri_host_local($var(attrs))) {
    This function can be used from REQUEST_ROUTE, FAILURE_ROUTE,
    BRANCH_ROUTE.
 
-   Example 1.8. is_domain_local usage
+   Example 1.9. is_domain_local usage
 ...
 if (is_domain_local($rd)) {
         ...


davidtrihy-genesys:
Hi Bogdan, thanks for the feedback I'll work on integrating the changes you requested in the coming week

bogdan-iancu:
just a gentle ping here :)

bogdan-iancu:
and one more gentle ping here :)

davidtrihy-genesys:
@bogdan-iancu sorry I've been quite busy, the grammar changes and the fix suggested for ensuring we strictly match subdomains works. 

I also believe I have the domain changes working when you load the hash table into memory, the check for that is a lot easier as it can utilise the name_alias changes and provide the domain to check and the accept_subdomain parameter from the db instead of the socket aliases, still need to verify my changes work there but I think they should, just need to setup my env to add the new column.

The main reason why I haven't pushed these changes yet is I'm having trouble with the domain module when you do not load the table into memory (ie. db mode == 0) the query for this is a strict `SELECT * from domain where domain == 'super.mario.com'` via this piece of code `domain_dbf.query(db_handle, keys, 0, vals, cols, 1, 2, 0, &res)` my understanding here is that it does just do that strict select so there would need to be some refactor done to achieve this lookup.

One area would be to tokenise the domain so that we could potentially do multiple queries so domain `four.sub.domain.com` would be able to query for any subdomain of `sub.domain.com` or `domain.com` because the logic dictates that you should be able to match it against either if the `accept_subdomain` flag is set, this should be relatively easily achieved by doing a `SELECT * from domain where domain LIKE '%domain.com'` then just looping through the results and matching if the accept_subdomain is true.

Do you have any good example you can point me to achieve this `LIKE` query otherwise when db mode is 0 the lookup will be very difficult to achieve or non performant, one thing we could do is to do multiple db queries of `domain == 'four.sub.domain.com'` then `domain == 'sub.domain.com'` etc... but that is not an approach I particularly like.

If I could your thoughts I could have this finished quickly enough as I have the bulk of the changes just that scenario described and I need to test it

davidtrihy-genesys:
@bogdan-iancu I got the feature added to the domain module, I don't think there's anyway around needing to iterate through the domain to check each one against the host and accepting the subdomain.

I have tested it with both db mode on and off and with accept subdomain on and off and it works, also the new grammar you suggested is there too along with the fixes you mentioned.

There was a failure checking the rtpio I'm not sure if what I did affected that, let me know your thoughts on the PR

davidtrihy-genesys:
Just added in some comments to explain the algorithm @bogdan-iancu I understand this feature you have tagged for 3.6-dev which I presume means you would like it in before the conference on the 26th so let me know if you want to have a chat and we can try and fix any residual issues you may have with my PR or if you need clarification

modules/domain/domain.c
```diff
@@ -92,54 +93,60 @@ int domain_db_ver(str* name, int version)
 	return db_check_table_version(&domain_dbf, db_handle, name, version);
 }
 
-
-/*
- * Check if domain is local and store attributes in a pvar
- */
-int is_domain_local_pvar(struct sip_msg *msg, str* _host, pv_spec_t *pv)
+int db_table_lookup(struct sip_msg *msg, str *lookup_domain, str *host, pv_spec_t *pv)
```
davidtrihy-genesys, line 96: This is mostly just the existing `is_domain_local_pvar` refactored into a specific db lookup function with one minor change where it first gets the accept_subdomain column and does the match subdomain based on the supplied host rather than implictly trusting it matches based on getting a result back

name_alias.h
```diff
@@ -58,6 +60,29 @@ struct alias_function {
 
 extern struct alias_function* alias_fcts;
 
+static inline int match_domain(char* alias, int alias_len, char* host, int host_len, int accept_subdomain) {
```
davidtrihy-genesys, line 63: Broke this functionality out into a separate function so it could be exposed to the domain module for use there, this file is part of the core OpenSIPs so the domain module will always have access to this file regardless of the Makefile definitions

bogdan-iancu (approved):
Thank you @davidtrihy-genesys , it looks good - what is missing is the increment on the table version for "domain" (as the schema changed), but as we are in a hurry for 3.6 release, I will take care of that after merging ;)

PR #3474, opened 2024-09-18
Fix a corner case causing clusterer to segfault

tommybrecher:
Fix a corner case where if the local_node id is higher than the remote node_id and DNS resolution fails opensips would segfault.

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**


tommybrecher:
Rejected upstream by different patch

PR #3476, opened 2024-09-19
Fix `rtpengine_manage()` SDP existence check with custom body

ar45:
Correct the logic for checking the presence of SDP body when body var is provided. Previously, the condition was inverted, leading to incorrect handling of messages.

If possible to backport to 3.5.


ar45:
Still active

PR #3479, opened 2024-09-25
Add support for UUID version 7 generation.

ar45:
**Summary**
Add support for uuid v7 generation.

Credits https://github.com/nalgeon/uuidv7/blob/c400f1daa/src/uuidv7.c


liviuchircu:
Hey, @ar45!  Finally managed to look at this -- the support is nice, as it's directly embedded, optional and free of any shared library versioning restrictions.  Moreover, the author released their code as _"free and unencumbered software released into the public domain"_, so all is good on the licensing part when importing the whole function.

PR #3480, opened 2024-09-25
[maxfwd]: Fix bug `is_maxfwd_lt` function limit value

ar45:
**Summary**
`is_maxfwd_lt( val )` was using a random value, the ptr address.

**Solution**
Correctly use the int value.

**Compatibility**
Should be backported to all supported versions of opensips.


ar45:
Still active

ar45:
active

bogdan-iancu:
Great finding and fix @ar45  - thanks a lot and sorry for taking longer to merge this PR.

Issue #3486, opened 2024-10-03
[FEATURE]  Authentication for Cassandra / ScyllaDB


zooptwopointone:
**Is your feature request related to a problem? Please describe.**

I am working on getting Cache DB Working with our Scylla cluster.   I first implemented authentication into the current Cassandra module.    I will put in a Pull request for that.   But I am curious what I should do for the Scylla DB changes.  If you could guide me as I am not really a C programmer.    Basically the Scylla implementation/driver is exactly the same as Cassandra.  The only difference to the code is including the Scylla Lib vs the Cassandra lib.   1 other change was compiling on Debian I has to add this to the main Makefile's LDFLAGS   -ltinfo

So my question is for the scylla driver is there a better way to provide an option to choose between them dynamically via the MAKE options or something or is it better that I just copy the full cassandra module and submit it under scylla?

thanks.

**Describe the solution you'd like**
I have a PR I will submit, but just enabling username and password in the db_url to be passed to the Scylla/cassandra driver for authentication.

**Implementation**
Just implemented what was available in the lib to provide username and password and current functionality of Opensips

**Describe alternatives you've considered**
none

**Additional context**

I want to use Authentication when connecting to the database.

zooptwopointone:
Let me know if there is something I need to do for the  pull request that I submitted.

thanks.

zooptwopointone:
https://github.com/OpenSIPS/opensips/pull/3524  I had to create a new pull request as the other was automatically closed.

bogdan-iancu:
@zooptwopointone , thank you for the PR, it was merged. In regards to your dilemma on Scylla Lib vs the Cassandra lib, the approach here should be to add a shell variable in the module's Makefile to act as a switch between compiling/linking against the 2 libs.

PR #3487, opened 2024-10-03
Add authentication to Cassandra module

zooptwopointone:
**Summary**
This is to add authentication for the Cassandra module.  the URL already has the ability to provide a username and password, this just implements it back to the driver.    If you Database has authentication enabled, Opensips would not be able to connect to it.

**Details**
New Feature.
Need to use authentication when connecting to the database.
This is a very small change using existing logic in opensips code.
Uses the current db_url provided username:password structure for other DB types and makes it work for cassandra.  Documentation would need to be updated to let others know this is now possible to use.

**Solution**
Simply implemented the Cassandra API to pass username and password if provided.

**Compatibility**
It is unknown to me when Cassandra implemented authentication.  but I know it has been almost 8 years now.

**Closing issues**
I put in an issue, with some other questions related to this pull request.  And questions for how to best offer support for Scylla DB as it's required librariy is different than cassandra, but the API is the same.
https://github.com/OpenSIPS/opensips/issues/3486

### DIFF
--- a/modules/cachedb_cassandra/cachedb_cassandra_dbase.c
+++ b/modules/cachedb_cassandra/cachedb_cassandra_dbase.c
@@ -138,9 +138,12 @@ int cassandra_reopen(cassandra_con *cass_con)
 	return 0;
 }
 
-int cassandra_new_connection(cassandra_con *con, char *host, int port)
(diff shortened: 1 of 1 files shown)


zooptwopointone:
Is there something missing here?  do I need to do something else to get this asigned?

PR #3490, opened 2024-10-08
pua_reginfo: Fix memory leaks, fix crashes in edge-cases with large NOTIFY-Bodies

carstenbock:
**Summary**
- pua_reginfo: Use Stringbuffer for creating the body of a NOTIFY message
- pua_reginfo: Fixes some crashes in edge-cases + some memory leaks.
- lib/str_buffer.c/.h: A str_buffer for building strings without knowing the size.


carstenbock:
Will do some more changes, to make the checks happy and for feature-completeness.

Issue #3498, opened 2024-10-18
[CRASH] Accounting ctx free in free_cell context


vladpaiu:
**OpenSIPS version you are running**

```
3.4.5
```

**Crash Core Dump**
```
#0  0x00007f1371367de7 in dlg_ctx_get_ptr (dlg=0x70616d7074723d61, pos=0) at dlg_ctx.c:65
#1  0x00007f136e6994b3 in free_acc_ctx (ctx=0x7f1372733868) at acc_logic.c:175
#2  unref_acc_ctx (ctx=0x7f1372733868) at acc_logic.c:1219
#3  0x00007f1371fe617c in empty_tmcb_list (head=head@entry=0x7f1372734288) at t_hooks.c:53
#4  0x00007f1371fb81e2 in free_cell (dead_cell=0x7f1372734218) at h_table.c:127
#5  0x00007f137200b312 in wait_handler (wait_tl=0x7f1372734298) at timer.c:478
#6  timer_routine (ticks=<optimized out>, set=<optimized out>) at timer.c:1115
#7  0x000055ca5befe3a6 in handle_timer_job () at timer.c:1018
#8  0x000055ca5c061d60 in handle_io (idx=<optimized out>, event_type=<optimized out>, fm=<optimized out>) at net/net_udp.c:299
#9  io_wait_loop_epoll (repeat=<optimized out>, t=<optimized out>, h=<optimized out>) at net/../io_wait_loop.h:311
#10 0x000055ca5c0675b1 in udp_start_processes (chd_rank=chd_rank@entry=0x55ca5c1baff8 <chd_rank>, startup_done=startup_done@entry=0x0) at net/net_udp.c:528
#11 0x000055ca5bdcb430 in main_loop () at main.c:237

(gdb) f 1
#1  0x00007f136e6994b3 in free_acc_ctx (ctx=0x7f1372733868) at acc_logic.c:175
175	acc_logic.c: No such file or directory.
(gdb) p T
$1 = (struct cell *) 0x7f1372741818
(gdb) p T->dialog_ctx
$2 = (void *) 0x70616d7074723d61
```

Note that there exists a dangling T pointer ( probably pointing to an already de-allocated transaction )

**Describe the traffic that generated the bug**
Unknown

**To Reproduce**
Unknown

**Relevant System Logs**
None

**OS/environment information**
Debian 11.10, installed from official OpenSIPS repo.

**Additional context**
[...]

PR #3500, opened 2024-10-18
set_t while free_cell callbacks are running

vladpaiu:
**Summary**
Potential fix for https://github.com/OpenSIPS/opensips/issues/3498

**Details**
Potential fix for https://github.com/OpenSIPS/opensips/issues/3498

**Solution**
set_t while free_cell callbacks are running

**Compatibility**
Should be compatible

**Closing issues**

### DIFF
--- a/modules/tm/h_table.c
+++ b/modules/tm/h_table.c
@@ -120,6 +120,10 @@ void free_cell( struct cell* dead_cell )
 	struct sip_msg *rpl;
 	struct totag_elem *tt, *foo;
 	struct proxy_l *p;
+	struct cell *backup_t;
+
+	backup_t = get_t();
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
@vladpaiu , thanks for the work here. IMHO, in regards to the #3498 issue, the fix should not be setting a new T while running the destroy callbacks (there is no need for the value), but rather to be sure the T is cleaned up at the end of any previous processing. Some TM code is "leaking" the T value.... and fixing this leaking should be the proper approach.

bogdan-iancu:
The #3498 crash is more likely to be fixed by #3502 . The change here is not fixing anything IMHO, so let's skip it.

PR #3501, opened 2024-10-18
New rtp.io module (embedded RTPProxy)

sobomax:
**Summary**
This PR includes a new rtp.io module which allows to provide media relaying and other RTP manipulations without external media relay.

**Details**
Traditionally, OpenSIPS only handles signalling, offloading media operation to the external relay. However in some situations using external relay may be undesirable from system complexity point of view and others, so having an option to handle it internally may be useful.

**Solution**
What rtp.io module does is it starts up a RTP handling threads in the main OpenSIPS process and let rtpproxy module access those threads via a 1:1 socketpair.

The module requires RTPProxy 3.1 or higher, compiled with --enable-librtpproxy.

When rtpproxy module is loaded without arguments and rtp.io is loaded as well, the sockets exported by the rtp.io will be used automatically in the set 0. Otherwise those sockets can be used as part of other set(s) by using "rtp.io:auto" moniker.

Provided is also a CI job that builds rtp.io in a rtpproxy container for all supported architectures and tests it using voiptests package.

**Compatibility**
None

**Closing issues**
None

### DIFF
--- a/modules/rtp.io/rtp_io.c
+++ b/modules/rtp.io/rtp_io.c
@@ -1,0 +1,239 @@
+/*
+ * Copyright (C) 2023 Sippy Software, Inc.
+ *
+ * This file is part of opensips, a free SIP server.
+ *
+ * opensips is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ * opensips is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+ *
+ */
+
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include "../../sr_module.h"
+#include "../../mod_fix.h"
+#include "../../pt.h"
+#include "../../dprint.h"
+#include "../../timer.h"
+
+#include "rtp_io.h"
+#include "rtp_io_util.h"
+#include "rtp_io_params.h"
+#include "rtp_io_host.h"
+
+static int mod_init(void);
+static int child_init(int rank);
+static void mod_destroy(void);
+
+static const dep_export_t deps = {
+    { /* OpenSIPS module dependencies */
+        { MOD_TYPE_DEFAULT, "rtpproxy", DEP_SILENT|DEP_REVERSE },
+        { MOD_TYPE_NULL, NULL, 0 },
+    },
+};
+
+static int rtp_io_getchildsock(int);
+
+/*
+ * Exported functions
+ */
+static const cmd_export_t cmds[] = {
+    {"rtp_io_getchildsock", (cmd_function)rtp_io_getchildsock, {0}, 0},
+    {0}
+};
+
+/*
+ * Exported params
+ */
+static const param_export_t params[] = {
+    {"rtpproxy_args",       STR_PARAM|USE_FUNC_PARAM, (void*)rio_set_rtpp_args},
+    {0}
+};
+
+static const proc_export_t procs[] = {
+        {.name = "RTPProxy host process", .function = rtpproxy_host_process,
+         .no = 1, .flags = PROC_FLAG_HAS_IPC,
+         .post_fork_function = rtp_io_close_serv_socks},
+        {0}
+};
+
+struct module_exports exports= {
+    "rtp.io",        /* module's name */
+    MOD_TYPE_DEFAULT,/* class of this module */
+    MODULE_VERSION,
+    DEFAULT_DLFLAGS, /* dlopen flags */
+    0,		 /* load function */
+    &deps,           /* OpenSIPS module dependencies */
+    cmds,            /* exported functions */
+    0,               /* exported async functions */
+    params,      	 /* param exports */
+    0,       	 /* exported statistics */
+    0,         	 /* exported MI functions */
+    NULL,       	 /* exported pseudo-variables */
+    0,		 /* exported transformations */
+    procs,               /* extra processes */
+    0,        	 /* module pre-initialization function */
+    mod_init,        /* module initialization function */
+    0,               /* reply processing function */
+    mod_destroy,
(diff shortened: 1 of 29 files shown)


sobomax:
Need to work around issue with pushing to ghcr, so pipeline works in the PRs.

liviuchircu:
Hey @sobomax !  I tried out the build & startup phase, feedback as follows (from least to most problematic):

1) if `librtpproxy.a` is built without libsrtp installed, then you install it meanwhile (to fix `rtp.io.so` build warning about missing libsrtp), you get a silent startup exit due to RTPProxy's `rtpp_static_modules_lookup()` not finding the `dtls_gw` module.  Small problem at the end of the day, fully fixable through public builds and/or a proper developer `README.md`.
2) on my system, `libsrtp2-1` package was built without AES GCM ciphers, hence there was no `GCM` define, hence few symbols were missing, such as: `srtp_crypto_policy_set_aes_gcm_128_16_auth()`, which gave a `rtp.io` loader error on startup.  Also seems like a manageable problem.
3) OpenSIPS does **not** start anymore with a module which creates a dedicated proc, due to below assert.  So it still looks like the PR is in WIP mode here, and we need a solution to restore FIFO/HTTPD/etc. functionality:

```
    opensips: rtp_io.c:228: child_init: Assertion `rank >= 0' failed.
```
4) Last but not least, below is how the process/thread architecture looks like, after RTPProxy is booted through the `rtpp_main()` function, called at `mod_init()`.  We can see that it forks `12` threads in total, but they are part of the "Attendant" process.  Which means that if any of them segfaults, the whole attendant process is going down, unable to do "emergency shutdown" things such as dumping usrloc/dialog data to DB, etc.  I believe this also needs reworking, ideally we get the entire `rtpp_main()` to be called from a dedicated process, but I cannot tell the implications off the top of my head, as I haven't given it much though/testing -- maybe moving it there is simply not possible?!

opensips-rtp io

Cheers!

sobomax:
> Hey @sobomax ! I tried out the build & startup phase, feedback as follows (from least to most problematic):
> 
> 1. if `librtpproxy.a` is built without libsrtp installed, then you install it meanwhile (to fix `rtp.io.so` build warning about missing libsrtp), you get a silent startup exit due to RTPProxy's `rtpp_static_modules_lookup()` not finding the `dtls_gw` module.  Small problem at the end of the day, fully fixable through public builds and/or a proper developer `README.md`.

Well, edge cases handling is somewhat obviously not very well handled since the module at the moment has only one user (us) and we know how to cook it. :)

Just open a PR against respective README.md in question as user #2. :)

> 2. on my system, `libsrtp2-1` package was built without AES GCM ciphers, hence there was no `GCM` define, hence few symbols were missing, such as: `srtp_crypto_policy_set_aes_gcm_128_16_auth()`, which gave a `rtp.io` loader error on startup.  Also seems like a manageable problem.

Well, again, this is more of a call for an action against rtpproxy repository. I build against latest ubuntu and debian on like 6 different architectures (see the appropriate CI pipeline attached). If one of libraries installed on your system @liviuchircu deviates from that and it bothers you, then just open a PR in the https://github.com/sippy/rtpproxy and we'll handle that edge case gracefully. :)

> 3. OpenSIPS does **not** start anymore with a module which creates a dedicated proc, due to below assert.  So it still looks like the PR is in WIP mode here, and we need a solution to restore FIFO/HTTPD/etc. functionality:
> 
> ```
>     opensips: rtp_io.c:228: child_init: Assertion `rank >= 0' failed.
> ```

Fixed that, thanks for the reporting! We just never use any of the modules in question, so I just put it as a safeguard since I did not know what negative value might mean in that somewhat critical code path.

Regarding WIP mode, no. The code is somewhat experimental, yes, not not a work in progress. The PR is in fact comes with the full working CI pipeline (this is the first module actually and the first CI to actually run a call scenario against OpenSIPS in fact) that builds against latest ubuntu and debian releases across 6 architectures and runs against 2 of them (amd64 and riscv). 

> 4. Last but not least, below is how the process/thread architecture looks like, after RTPProxy is booted through the `rtpp_main()` function, called at `mod_init()`.  We can see that it forks `12` threads in total, but they are part of the "Attendant" process.  Which means that if any of them segfaults, the whole attendant process is going down, unable to do "emergency shutdown" things such as dumping usrloc/dialog data to DB, etc.  I believe this also needs reworking, ideally we get the entire `rtpp_main()` to be called from a dedicated process, but I cannot tell the implications off the top of my head, as I haven't given it much though/testing -- maybe moving it there is simply not possible?!

Well, just to say since it's not impeding any of the functionality this should be more as a feature request than anything else. The last time I saw rtpproxy crashing in production is like 5 years ago across like 300+ systems deployed. And I went to the great deal to make it reality with the glitching and fuzzing CI runs. But no, moving processing to a separate PID is not impossible and is not that difficult if you have time. So, again, good feature request.

Thanks for a review, regards!

-Max

modules/rtp.io/rtp_io.c
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+
+    int argc;
+    const char *const *argv = rtp_io_env_gen_argv(&rpi_descp->env, &argc);
+    if (argv == NULL)
+        goto e1;
+
+    OPT_RESTORE();
+    rpi_descp->rtpp_cfsp = rtpp_main(argc, argv);
```
liviuchircu, line 187: This `rtpp_main()` function call is the entrypoint into the `librtpproxy.a` embedded RTPProxy daemon.  The default logic proceeds to fork `12` worker threads which will be part of the "Attendant" OpenSIPS worker.

This logic needs to be moved as part of a `proc_export_t` extra process of the module, to prevent any scenario where a thread aborts and prevents OpenSIPS shutdown cleanup procedure.  I hate to be pedantic here, but not only there are `188` modules following this rule and we shouldn't start making exceptions, but also it is very likely that `rtp.io` will gain a good amount of traction on release, which further reinforces the need of a clean design from the start.  This is not a feature request as I see it, it is a design requirement.

Cheers,

sobomax: OK, the threads have been moved into a dedicated extra process, please check. Thanks!

sobomax:
PR updated to fix the noted deficiencies. I've also updated CI to remove 20.04 (being deprecated by the GitHub) and replace it with 24.04 where appropriate.

liviuchircu:
Great work, @sobomax!  Both of the big issues I noted above seem fixed now:

* you can freely load `mi_fifo.so` or `mi_http.so` functionality on top of `rtp.io.so` without breaking OpenSIPS startup
* indeed, the attendant process is no longer forking RTPProxy worker threads, which are nicely isolated as part of the `"RTPProxy host process"` OpenSIPS worker (see attached picture)

I for one am excited to see the new, self-contained RTPProxy module getting some usage and feedback from the community, it's time to merge it!!

image

PR #3502, opened 2024-10-23
fix: Cleanup T after early dialog dlg_end_dlg

vladpaiu:
**Summary**
Potential fix for https://github.com/OpenSIPS/opensips/issues/3498

**Details**
T is leaked when dialog looks up the transaction for early dialog hangup

**Solution**
Restore T after lookup ( hopefully to NULL )

**Compatibility**
Should be backwards compatible

**Closing issues**
https://github.com/OpenSIPS/opensips/issues/3498

### DIFF
--- a/modules/dialog/dlg_req_within.c
+++ b/modules/dialog/dlg_req_within.c
@@ -460,6 +460,7 @@ int dlg_end_dlg(struct dlg_cell *dlg, str *extra_hdrs, int send_byes)
 	str str_hdr = {NULL,0};
 	struct cell* t;
 	int res = -1;
+	struct cell* bkp_t;
 
 	if (!send_byes) {
 		dual_bye_event(dlg, NULL, 0);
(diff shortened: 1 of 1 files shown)


vladpaiu:
not stale

vladpaiu:
still valid

bogdan-iancu:
@vladpaiu , totally valid, both the report and the fix.  And I see a similar case with the same fix (already in place) in the [b2b_entities module](https://github.com/OpenSIPS/opensips/blob/master/modules/b2b_entities/dlg.c#L2460).
This `t_lookup_ident()` functions is used , via the TM API, only from dialog (your case) and from b2b_entities (case I found) modules only. It is also internally used by TM module - for some MI and inject scenarios.
My point is, in none of the case, there is no larger TM context needed, the scope of the transaction pointer is local, for couple of lines of code. Even more the, `t_lookup_ident()`  returns the found transaction as parameter .
So, bottom line - should the right fix be simply [NOT setting the T inside](https://github.com/OpenSIPS/opensips/blob/master/modules/tm/t_lookup.c#L1223) this `t_lookup_ident()` ??
By doing this, we solve in a simpler way the usage of the function via TM API (we do not need the fix here) and there is 0 impact of the TM internal usage of the function also.

vladpaiu:
not sure if that's good enough.. as maybe in the future in some cases we need the T to be set so subsequent callbacks done after lookup_ident find the proper T set...

bogdan-iancu:
That's the point - right now we do not have such a case; in all existing cases the transaction is actually returned and used via function param...and whatever transactional processing follows, it does not use the T...

bogdan-iancu:
I just pushed a fix, as per [my initial comment](https://github.com/OpenSIPS/opensips/pull/3502#issuecomment-2697802294) .

PR #3515, opened 2024-11-14
Feature: add pause recording function to module rtpengine

edman80:
**Summary**
RTPEngine Module: add exported function for `pause recording` operation. Pause recording tells the recording daemon to keep the recording file open so you don't end up with multiple for a single call.
**Details**
Feature to add pause recording functionality to RTPEngine module.
**Solution**
n/a

**Compatibility**
Backwards compatible.

**Closing issues**


NormB:
Successfully tested this update against the master branch.

PR #3532, opened 2024-12-02
rtpengine: Update fd when node reconnects

hateeyan:
When a node reconnects, we should update the fd to avoid using the old fd in subsequent polls.


hateeyan:
active

hateeyan:
active

PR #3533, opened 2024-12-03
Initial version of Janus module

vladpaiu:
**Summary**
Janus Websocket connector

**Details**
Janus Websocket connector

**Solution**
Janus Websocket connector

**Compatibility**
Backwards compatible

PR #3534, opened 2024-12-03
Add support for bigints in JSON ( store as string )

vladpaiu:
**Summary**
Add support for bigints in JSON

**Details**
In certain scenarios ( like #3533 ), we are processing JSONs which have bigints in them, so we need to properly parse those and export sub-fields in the opensips scripts.

**Solution**
Expose bigints in JSON as string to the opensips script.

**Compatibility**
Backwards compatible

### DIFF
--- a/modules/json/json.c
+++ b/modules/json/json.c
@@ -457,10 +457,16 @@ int pv_get_json_ext(struct sip_msg* msg,  pv_param_t* pvp, pv_value_t* val, int 
 
 	if( json_object_is_type(obj, json_type_int) )
 	{
-		val->rs.s = sint2str(json_object_get_int(obj), &val->rs.len);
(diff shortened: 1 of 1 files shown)


liviuchircu (changes requested):
Undefined variable: `int64_t int_value;`.  There is also a double `;;` typo in the middle.  LGTM otherwise.

liviuchircu:
Let's merge this, as it fixes a big limitation, at the end of the day.  Still, until we have 64bit integer support at _opensips.cfg_ level, script developers will have to deal with the fact that this solution lacks symmetry... i.e. if I re-serialize a freshly parsed **$json** variable back to a JSON string, the module will not be able to distinguish between a "OpenSIPS-quoted 64-bit integer" or a "string of digits longer than 10".

PR #3535, opened 2024-12-03
Initial version of the trie module

vladpaiu:
**Summary**
Initial version of the trie module

**Details**
Add a generic trie module implementation.

The drouting module is trying to do too many things at once. For the use cases where we just need maintain and query and internal trie ( and we do not have any gateways/carriers or routing involved ), the trie module provides a more memory efficient alternative.

Furthermore, due to the drouting complexity, we cannot easily update a rule in place and/or delete a prefix rule. The new trie module exposes new upsert&delete MI commands, thus getting rid of the limitation of having to reload the whole dataset in case minor changes need to be made to the trie.

**Solution**
New trie module.

**Compatibility**
Backwards compatible


bogdan-iancu:
Hi @vladpaiu , thank you for the contribution here. Just for the sake of the discussion , you know you can use the drouting module without the carriers and gateways part, and have the do_routing() to do only matching, nothing more. And IMHO this is a 100% overlapping with this new module....if I'm not missing anything.
But of course, all the time there are different ways to skin the cat. And having this as module will avoid any conflicts and direct overlapping - just more options.
Could you look into the failed checks from CI system and have them addressed please?

vladpaiu:
the reason for splitting this away from drouting was mostly memory usage related, while doing comparisons this is using close to 3x less mem. also, it was about adding/deleting rules without having to worry about all of the  drouting 'dependencies'.


in regards to the CI/CD failures, they are not related to this PR changes:

cachedb/cachedb_dict.c:481:31: error: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Werror,-Wformat]
  481 |                         i += sprintf(obj+i, "%ld", pair->val.val.i64);

bogdan-iancu:
@vladpaiu , please check the above CI issues, there are specific to the trie module this time :)

PR #3540, opened 2024-12-12
Support HTTP fragmentation

vladpaiu:
**Summary**
Sometimes, HTTP POST traffic can become fragmented, and the HTTP module does not support it.

**Details**
Enforce fragmentation at the interface layer somehow, and notice how fragmentation breaks httpd mod.

**Solution**
Accumulate HTTP incoming buffer until complete, and only then pass it to the processing layer.

**Compatibility**
Backwards compatible

### DIFF
--- a/modules/httpd/httpd_proc.c
+++ b/modules/httpd/httpd_proc.c
@@ -62,6 +62,11 @@ extern str tls_ciphers;
 extern struct httpd_cb *httpd_cb_list;
 static union sockaddr_union httpd_server_info;
 
+
+extern int receive_buf_size;
+extern char *httpd_receive_buff;
+extern int httpd_receive_buff_pos;
+
 static const str MI_HTTP_U_URL = str_init("<html><body>"
 "Unable to parse URL!</body></html>");
 static const str MI_HTTP_U_METHOD = str_init("<html><body>"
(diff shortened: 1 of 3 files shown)


modules/httpd/httpd_proc.c
```diff
@@ -132,6 +132,24 @@ int httpd_get_val(void *e_data, void *data, void *r_data)
 	return 0;
 }
 
+static char httpd_full_buff[16384];
```
bogdan-iancu, line 135: Why `16384` ? any special meaning? Maybe this should be a modparam, to be able to increase it? (of course, it will have to be pkg/system allocated at module startup.

modules/httpd/httpd_proc.c
```diff
@@ -132,6 +132,24 @@ int httpd_get_val(void *e_data, void *data, void *r_data)
...
+{
+	str_str_t *kv = (str_str_t*)e_data;
+	if (kv==NULL) {
+		LM_ERR("null data\n");
+	} else {
+		LM_DBG("data=[%p] [%p][%p] [%.*s]->[%.*s]\n",
+			kv, kv->key.s, kv->val.s,
+			kv->key.len, kv->key.s,
+			kv->val.len, kv->val.s);
+		memcpy(httpd_full_buff+httpd_full_buff_pos,kv->val.s,kv->val.len);
```
bogdan-iancu, line 147: Isn't here a need for a overflow test ?

bogdan-iancu (changes requested):

bogdan-iancu (approved):

bogdan-iancu:
Thank you @vladpaiu :heart:

PR #3550, opened 2025-01-03
Add onreply_route async capabilities

vladpaiu:
**Summary**
Add onreply_route async capabilities

**Solution**
Clone reply before running async functions & free afterwards

**Compatibility**
Backwards incompatible, previous async in onreply_route was run in sync mode.

### DIFF
--- a/modules/tm/async.c
+++ b/modules/tm/async.c
@@ -30,6 +30,7 @@
 #include "h_table.h"
 #include "t_lookup.h"
 #include "t_msgbuilder.h"
+#include "t_cancel.h"
 
 
 typedef struct _async_tm_ctx {
@@ -52,9 +53,15 @@ typedef struct _async_tm_ctx {
 	/* e2e ACK */
 	struct cell *e2eack_t;
 
+	/* reply related things */
+	struct sip_msg *reply;
+	int reply_length;
+	int branch;
+
 } async_tm_ctx;
 
 extern int return_code; /* from action.c, return code */
+extern int _tm_branch_index; /* TM branch index */
 
 
 
@@ -83,6 +90,13 @@ int t_resume_async(int fd, void *param, int was_timeout)
 	const struct socket_info* backup_si;
 	struct cell *t= ctx->t;
 	int route;
+	int msg_status;
+	int last_uac_status;
+	int branch;
+	int reply_status;
+	struct ua_client *reply_uac;
+	branch_bm_t cancel_bitmap;
+	utime_t timer;
 
 	if (valid_async_fd(fd))
 		LM_DBG("resuming on fd %d, transaction %p \n", fd, t);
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
@vladpaiu , this sounds awesome !!! Let me check it !

modules/tm/t_reply.c
```diff
@@ -1671,6 +1674,19 @@ int reply_received( struct sip_msg  *p_msg )
 		if (onreply_avp_mode)
 			/* restore original avp list */
 			set_avp_list( backup_list );
+
+		if (async_status > 0) {
+			/* async was started in the onreply route, no need to do anything more here */
+			/* mark that the UAC received replies */
+			uac->flags |= T_UAC_HAS_RECV_REPLY;
```
bogdan-iancu, line 1681: does it make any difference if we move the setting of the flag before running the onreply route, so we do not have to duplicate ?

modules/tm/async.c
```diff
@@ -319,8 +553,18 @@ int t_handle_async(struct sip_msg *msg, struct action* a,
 	}
 
 	if (route_type!=REQUEST_ROUTE) {
```
bogdan-iancu, line 555: Some
```
if ()
else if ()
else
```
may look better here :)

modules/tm/async.c
```diff
@@ -219,6 +217,242 @@ int t_resume_async(int fd, void *param, int was_timeout)
...
+	/* run the resume_route (some type as the original one) */
+	swap_route_type(route, ctx->route_type);
+	/* do not run any post script callback, we are a reply */
+	run_resume_route( ctx->resume_route, ctx->reply, 0);
+	set_route_type(route);
+
+	/* DO TM suspended actions to decide if we need to relay */
+	LOCK_REPLIES( t );
+
+	/* we fire a cancel on spot if (a) branch is marked "to be canceled" or (b)
```
bogdan-iancu, line 339: this chunk below is duplicated from the `reply_received()` (all the way to `restore_exit_reply` label). I haven;t looked into details, but maybe we should avoid code duplicate, especially in such a delicate section, on reply handing. Maybe some refactoring into some new inline function ?

modules/tm/async.c
```diff
@@ -219,6 +217,242 @@ int t_resume_async(int fd, void *param, int was_timeout)
...
+	}
+
+route:
+	if (async_status == ASYNC_DONE_CLOSE_FD && valid_async_fd(fd))
+		close(fd);
+
+	/* run the resume_route (some type as the original one) */
+	swap_route_type(route, ctx->route_type);
+	/* do not run any post script callback, we are a reply */
+	run_resume_route( ctx->resume_route, ctx->reply, 0);
```
bogdan-iancu, line 333: The initial onreply route allows you to run with the lock (see the `onreply_avp_mode` setting) or without lock. The resume route is not impacted (correct me if I'm wrong) by this setting, leading to concurrent access to the transactional list of AVPS.

modules/tm/async.c
```diff
@@ -219,6 +217,242 @@ int t_resume_async(int fd, void *param, int was_timeout)
...
+	/* enviroment setting */
+	current_processing_ctx = ctx->msg_ctx;
+	backup_t = get_t();
+	backup_e2eack_t = get_e2eack_t();
+	backup_cancelled_t = get_cancelled_t();
+
+	set_cancelled_t(ctx->cancelled_t);
+	set_e2eack_t(ctx->e2eack_t);
+	reset_kr();
+	set_kr(ctx->kr);
```
bogdan-iancu, line 254: I do not see this `ctx->kr` set anywhere (in the new code). Do we need this? KR is typical for request processing, not for replies.
Similar the e2eack and cancel stuff.

bogdan-iancu (changes requested):

modules/tm/async.c
vladpaiu: fixed in b2811a5f4c8403695acd7bdb92a7fe1f3df87881

resolving comment

modules/tm/async.c
vladpaiu: address in 7477f3f42b669f6c8dcfb0f0c1af73d1e84706ab , resolving comment

modules/tm/async.c
vladpaiu: addressed in 16ef67d27e1fabec487d65b41f1b502247f3093f , resolving comment

modules/tm/async.c
vladpaiu: good catch ! fixed under dd79818af2ee0c10d1b37c0f1a648888cb54e302, resolving comment

vladpaiu:
@bogdan-iancu addressed some of the comments, also fixed a mem leak under https://github.com/OpenSIPS/opensips/pull/3550/commits/6811373f7791e99749d7fb1cced36c75a6604e7f that I saw today while doing a bit of load testing to make sure all's good after recent changes.

modules/tm/t_reply.c
vladpaiu: 4 hours later it struck me that you were talking about T_UAC_HAS_RECV_REPLY :))

to be honest, I'm not sure if it matters or not, but to keep things consistent as they were before, I'm setting it after the reply route has been run ( with async or without async )

vladpaiu:
@bogdan-iancu  any other things to address before we can merge this ?

bogdan-iancu:
I'm on it :)

modules/tm/t_reply.c
bogdan-iancu: it makes sense, let's keep it like this for now

PR #3554, opened 2025-01-10
Add json_merge script function

vladpaiu:
**Summary**
Add json_merge script function

**Details**
When using provisioning things via JSON on a multi-layered approach it sometimes useful to merge the 'child' layer with the 'parent' layer for inheriting attributes that are not overwritten and for overwriting attributes that specifically come form the child ( patch ) layer.

**Solution**
Add json_merge script function

**Compatibility**
Backwards compatible

### DIFF
--- a/lib/cJSON.c
+++ b/lib/cJSON.c
@@ -2528,3 +2528,73 @@ void cJSON_Minify(char *json)
     /* and null-terminate. */
     *into = '\0';
 }
+
+cjbool cJSON_IsObject(const cJSON * const item)
+{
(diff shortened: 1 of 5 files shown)


vladpaiu:
all of the CI/CD failures are not related to this PR changes:

cachedb/cachedb_dict.c:481:31: error: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Werror,-Wformat]
  481 |                         i += sprintf(obj+i, "%ld", pair->val.val.i64);

PR #3556, opened 2025-01-14
Fix trie mod crash & clang compilation error

vladpaiu:
**Summary**
Fix trie mod crash & clang compilation error

**Details**
Fix trie mod crash & clang compilation error


bogdan-iancu:
Let's wait for the checks to complete :)

PR #3559, opened 2025-01-16
db_sqlite: new parameters for SQLite configuration

vasilevalex:
**Summary**
- Add new parameters to the module to configure the behavior of the database
- Use recommended by SQLite documentation way of processing locked state

**Details**
New parameters allowing tuning of SQLite database

**Solution**
Please review commits

**Compatibility**
The original behavior has been changed. If the database is locked, the process attempting to access it will no longer enter an infinite loop. Instead, it will return an error after the default timeout of 500 ms (this value can be modified as module parameter).


bogdan-iancu:
Thank you @vasilevalex

Issue #2551, opened 2021-06-11
[FEATURE] permissions: redesign address permissions regarding networks


Nestorfish:
**Is your feature request related to a problem? Please describe.**

We will soon have to insert quite a lot of IPv4 network ranges in `address`, so I took a look at the way OpenSIPS handles this, and more generally address permissions.

Here is what I found so far (the following applies to a partition):

## Storage

Individual hosts (mask 32 or 128 depending on address family) are stored in a hash-map, IP address begin the key.
The maximum number of buckets is set during compilation (`PERM_HASH_SIZE`, default 128, used in `perm_hash` macro as parameter to `core_hash`)
There is no particular ordering for elements matching this key, they are inserted in front of a chained list as they are read from database. 

Non-host network addresses are stored in a fixed-size array, whose size is set during compilation (the same `PERM_HASH_SIZE`, default 128).
They are ordered by `group_id`.

## Match

OpenSIPS first looks into the hosts hash-map.
It checks if the given `group_id` (if any) exists in any value, and aborts if not.
Then it iterates the chained list corresponding to the IP address.

If no matching entry is found above, it uses the networks array.
It first checks if there is any entry for the given `group_id` (if any), and aborts if not.
Then it looks for a matching entry.
(By the way, it forgets meantime the index where the `group_id` was found in the array, it could have avoided restarting from first position by remembering it)

## Remarks

 - Why does it first checks if the `group_id` exists? If it is absent, it only produces a LM_DBG statement along with a different return code (which is not documented for now I guess)
 - Networks entries are not scalable
 - Size of hashing function result could be set as a module parameter (like eg. `cache_collections` parameter of module `cachedb_local` which sets the number of bits). Not a must-have however
 - There is no fundamental difference between hosts and networks for this use case, they could all be considered as networks
 - This brings the last point: by looking first for individual hosts, it allows specific overrides of network entries. This behaviour can be generalized to networks: the match would be done on the longest mask entry found.

**Describe the solution you'd like**

Treat all entries the same way, namely:
No limitation for number of network entries
Match the smallest network first, being an host or not

**Implementation**

Proposal
---------

## Storage

 - Array of fixed size 129, the index being the mask of address entry (0 to 128).
 - Each entry of this array is a hash-map (as currently for hosts). During (re)load, we must ensure that the hashing is done on network address, ie. `ip` value after zeroing host bits.

## Match

1. Hash IP address
2. Set the initial index to the IP address family size (eg. 128 or 32)
[...]

PR #3560, opened 2025-01-20
permissions: Rewrite the in-memory storage backend

benceszigeti:
**Summary**  
Removed the hardcoded entry limit in the permission module’s in-memory subnet storage and optimized lookup performance.

**Details**  
Previously, exceeding 128 subnet entries prevented adding new subnets, requiring code modifications and custom builds to handle more. This update eliminates that limitation and enhances the scalability of in-memory storage for larger numbers of subnets and addresses.

**Solution**  
1. **Dynamic Hash Table**: Bucket count now grows dynamically, with code refactored for reusability.
2. **Prefix Tree**: Introduced a prefix tree for subnet storage, enabling efficient address lookup within subnet lists in a memory-friendly way.
3. **Group Partitioning**: Subnets and addresses are now partitioned by group ID in a hash table, removing the need for ordered storage and simplifying group lookups.
4. **Refactor**: Refactored the touched code parts to make the code more consistent.

**Compatibility**  
No breaking changes. The permission module interfaces (scripting, MI) remained unchanged: except, on the MI interface, when dumping subnets or addresses, the result is not in an ordered list.

**Closing Issues**  
https://github.com/OpenSIPS/opensips/issues/2551 (cc @Nestorfish)


c-leroy:
Hello @benceszigeti ,

I haven't had a look at your patch, but I have a question :
Does it work like IP routing (the finest network matches, ig. the with the largest network mask), or is it the first found that matches ?

benceszigeti:
@c-leroy
> I haven't had a look at your patch, but I have a question : Does it work like IP routing (the finest network matches, ig. the with the largest network mask), or is it the first found that matches ?

It returns the first match.

For `check_address` and `check_source_address`, we only need to know if any subnet matches, so we can return early (fastest) while iterating through the prefix tree.

For `get_source_group`, I can change the code to return the finest network match if there is a reason to do so. (Where we must also decide which group to return when the same match appears in multiple groups.)

benceszigeti:
If an address/subnet is defined in multiple groups, get_source_group may return a different group than it did previously. Previously, it returned the group with the lowest group ID where the match occurred (due to how the data was stored). Therefore, I am considering using a linked list for groups instead of a hash table, so we do not have to look up every group to find the smallest group ID. This change is necessary for backward compatibility.

benceszigeti:
Updated the PR: changes are now backward-compatible:
- Address/subnet MI dump returns an array ordered by group ID.
- `get_source_group` returns the lowest group ID with a match.

benceszigeti:
Thank you for the review, @liviuchircu!

I agree with your proposed changes, I have pushed the changes (accc2322e8d87fca11090c440a6c3e67fbdc91f7 -- `bucket grow factor`, 20c70e7c86c62498e0c4a010e80e3791d1e3b876 -- `reduce memory allocation`), and rebased the branch with master.

liviuchircu:
Changes looking good, @benceszigeti -- no more pending nitpicks here.  Merged.

PR #3563, opened 2025-01-21
implement hash table for tracking RTPEngine node allocations

### LINKED ISSUE #3036: [FEATURE] RTPEngine pool selection algo

When opensips is handling calls using SRTP/DTLS (in my situation in front of FreeSWITCH media servers) it is important that the proxy selected in the initial offer continues to be used throughout the call. In the event that a different media proxy is selected the DTLS renegotiation is triggered and this can disrupt and crash the back end media server (exact reasons unknown).

When we disable an instance in the pool this instance should continue to be used for any existing call sessions and only new sessions should be ignoring disabled hosts. There maybe a reason for multiple disabled modes to allow flexibility for anyone wishing to immediately disable a node from all activity.

**Implementation**

  - Component: `rtpengine` module
  - Type: function [select_rtpe_node](https://github.com/OpenSIPS/opensips/blob/master/modules/rtpengine/rtpengine.c#L3043)

Capturing the RTP socket selected after the initial offer, storing to the dialog and forcing all subsequent requests to this node. This is not going to work correctly as the node will be disabled and will likely cause those calls to fail.

**Additional context**

The current implementation is based on a crude hashing algorithm which assumes no change in the underlying pool, when a change is made all calculations will be affected. Including all the sockets in the pool (enabled and disabled) for every calculation and then checking the disabled flag before final selection (moving on to first enabled socket) will address this issue and impact only those which would have selected the disabled node during initial offer. Adding nodes to the pool will still trigger the issue however.

spacetourist:
**Summary**
This change allows RTP instances to be disabled safely without causing recalculation of allocations for existing dialogs, improving maintainability. The main driver is that I want to be able to disable instances to drain calls and perform maintenance on a 24 hour system (i.e. patching)

**Details**
The code was ported from Kamailio (https://github.com/kamailio/kamailio/pull/390) which offers this implementation, and then modifications were required to work correctly with the latest OpenSIPs codebase.

Note that this is a work in process I am submitting for initial review as I need to deliver this for my network, I'm not sure if you accept "Kamilio inspired" code but would appreciate a review and I can see this being useful to others. Documentation will be added as I continue work..

**Solution**
By implementing a hash table the module will now track all connected calls in a hash table of configurable size. This will be reviewed for all subsequent RTPEngine interactions and when found the existing allocation is preserved.

**Compatibility**
This should work as normal and the hash table has been made optional.

**Closing issues**
closes #3036


spacetourist:
I'm not sure how to clear the remaining build tests. Happy to look with some guidance

```
prefix_tree.o: file not recognized: file format not recognized
```

PR #3570, opened 2025-02-02
Libcouchbase3 for merge

lemenkov:
**Summary**
Make module `cachedb_couchbase` compatible with @couchbase's libcouchbase ver. 3.

**Details**
libcouchbase ver. 3 broke API and already 5 years old. Let's make this module compatible with this library. See also PR #2913 and #2389.

**Compatibility**
Module will require libcouchbase >= 3.0.0.

**Closing issues**
N/a.

### DIFF
--- a/modules/cachedb_couchbase/cachedb_couchbase_dbase.c
+++ b/modules/cachedb_couchbase/cachedb_couchbase_dbase.c
@@ -42,51 +42,64 @@ volatile lcb_STATUS op_error  = LCB_SUCCESS;
 
 static void couchbase_get_cb(lcb_INSTANCE* instance,
 		const void *cookie, lcb_STATUS error,
-		const lcb_get_resp_t *item)
+		const lcb_RESPGET *item)
 {
 	op_error = error;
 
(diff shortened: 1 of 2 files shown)


bogdan-iancu:
@lemenkov , thank you for this great work. As I do not have much experience with CouchDB, I have to 100% rely on your testing and consistency on this work.
BTW, I assume the new version is no longer compatible with pre ver. 3 CouchDB API, right? If so, is this something to worry?

lemenkov:
> @lemenkov , thank you for this great work. As I do not have much experience with CouchDB, I have to 100% rely on your testing and consistency on this work. BTW, I assume the new version is no longer compatible with pre ver. 3 CouchDB API, right? If so, is this something to worry?

Yes, unfortunately. It's not compatible anymore with pre-3.xy versions. I've tried to keep it compatible but the amount of work was too much at that time.

bogdan-iancu:
I see... any idea of how widely the pre-2.xy versions are still used by distros? as if not, we should consider acceptable the lost of compatibility with the old versions.

PR #3571, opened 2025-02-03
New TOS (per socket) options

nikbyte:
**Summary**  
This PR introduces two key changes:  
1. Allows setting the global TOS (Type of Service) parameter to `0` to disable the `IP_TOS` socket option.  
2. Adds a per-socket parameter `tos`, allowing TOS to be set on a per-socket basis. If this parameter is set, it overrides the global value.  

**Details**  
Previously, the global `TOS` parameter could not be set to `0`, meaning that disabling the `IP_TOS` socket option was not straightforward. This PR introduces the ability to explicitly set `TOS=0`, ensuring that no `IP_TOS` is applied.  

Additionally, this PR introduces a new per-socket `tos` parameter, which enables granular control over TOS settings on an individual socket level. If a socket-specific value is provided, it takes precedence over the global setting, allowing for more flexible traffic prioritization configurations.  

**Solution**  
- Modified the global TOS handling logic to allow `0` as a valid value.  
- Introduced a per-socket `tos` parameter in the socket configuration.  
- Ensured that when the per-socket `tos` value is set, it overrides the global TOS setting.  
- Maintained backward compatibility by preserving previous behaviors unless explicitly overridden.  

**Compatibility**  
- This change maintains backward compatibility as existing configurations remain unchanged unless explicitly modified.  
- The ability to set `TOS=0` does not affect existing deployments but provides more flexibility.  
- The per-socket `tos` parameter is an additional feature that does not interfere with existing configurations.

### DIFF
--- a/cfg.y
+++ b/cfg.y
@@ -180,6 +180,7 @@ extern char *finame;
 struct listen_param {
 	enum si_flags flags;
 	int workers;
+	int tos;
 	struct socket_id *socket;
 	char *tag;
 	char *auto_scaling_profile;
@@ -458,6 +459,7 @@ extern int cfg_parse_only_routes;
 %token SLASH
 %token AS
 %token USE_WORKERS
+%token SOCK_TOS
 %token USE_AUTO_SCALING_PROFILE
 %token MAX
 %token MIN
@@ -732,6 +734,9 @@ socket_def_param: ANYCAST { IFOR();
 					}
 				| USE_WORKERS NUMBER { IFOR();
 					p_tmp.workers=$2;
+					}
+				| SOCK_TOS NUMBER { IFOR();
+					p_tmp.tos=$2;
 					}
 				| AS listen_id_def { IFOR();
 					p_tmp.socket = $2;
@@ -2746,6 +2751,7 @@ static void fill_socket_id(struct listen_param *param, struct socket_id *s)
 	while (s) {
 		s->flags |= param->flags;
 		s->workers = param->workers;
+		s->tos = param->tos;
 		s->auto_scaling_profile = param->auto_scaling_profile;
 		s->tag = param->tag;
 		if (param->socket) {
--- a/cfg.y
+++ b/cfg.y
@@ -1540,7 +1540,7 @@ assign_stm: LOGLEVEL EQUAL snumber { IFOR();
 		  }
 		| MCAST_TTL EQUAL error { yyerror("number expected as tos"); }
 		| TOS EQUAL NUMBER { IFOR(); tos = $3;
-							if (tos<=0)
+							if (tos<0)
 								yyerror("invalid tos value");
 		 }
 		| TOS EQUAL ID { IFOR();
--- a/net/net_tcp.c
+++ b/net/net_tcp.c
@@ -234,7 +234,7 @@ static inline void set_sock_reuseport(int s)
 /*! \brief Set all socket/fd options:  disable nagle, tos lowdelay,
  * non-blocking
  * \return -1 on error */
-int tcp_init_sock_opt(int s, const struct tcp_conn_profile *prof, enum si_flags socketflags)
+int tcp_init_sock_opt(int s, const struct tcp_conn_profile *prof, enum si_flags socketflags, int sock_tos)
 {
 	int flags;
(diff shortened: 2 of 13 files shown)


net/net_tcp.c
```diff
@@ -248,9 +248,12 @@ int tcp_init_sock_opt(int s, const struct tcp_conn_profile *prof, enum si_flags
 #endif
 	/* tos*/
 	optval = tos;
-	if (setsockopt(s, IPPROTO_IP, IP_TOS, (void*)&optval,sizeof(optval)) ==-1){
-		LM_WARN("setsockopt tos: %s\n",	strerror(errno));
-		/* continue since this is not critical */
+	if (sock_tos > 0) optval = sock_tos;
```
bogdan-iancu, line 251: Could you do here a:
```
optval = (sock_tos > 0) ? sock_tos : tos ;
```
It's more clear and compact :)

net/net_tcp.c
```diff
@@ -376,10 +379,13 @@ int tcp_init_listener(struct socket_info *si)
 #endif
 	/* tos */
 	optval = tos;
-	if (setsockopt(si->socket, IPPROTO_IP, IP_TOS, (void*)&optval,
-	sizeof(optval)) ==-1){
-		LM_WARN("setsockopt tos: %s\n", strerror(errno));
-		/* continue since this is not critical */
+	if (si->tos > 0) optval = si->tos;
```
bogdan-iancu, line 382: same change here

cfg.y
```diff
@@ -1540,7 +1545,7 @@ assign_stm: LOGLEVEL EQUAL snumber { IFOR();
 		  }
 		| MCAST_TTL EQUAL error { yyerror("number expected as tos"); }
 		| TOS EQUAL NUMBER { IFOR(); tos = $3;
-							if (tos<=0)
+							if (tos<0)
```
bogdan-iancu, line 1548: Why trashing the 0 value? Not sure if 0 is a legitimate value for TOS, but the current code allows you to set it.

net/net_udp.c
```diff
@@ -199,15 +199,18 @@ int udp_init_listener(struct socket_info *si, int status_flags)
...
-	if (addr->s.sa_family==AF_INET6){
-		if (setsockopt(si->socket,  IPPROTO_IPV6, IPV6_TCLASS, (void*)&optval, sizeof(optval)) ==-1){
-			LM_WARN("setsockopt tos for IPV6: %s\n", strerror(errno));
-			/* continue since this is not critical */
-		}
-	} else {
-		if (setsockopt(si->socket, IPPROTO_IP, IP_TOS, (void*)&optval, sizeof(optval)) ==-1){
-			LM_WARN("setsockopt tos: %s\n", strerror(errno));
-			/* continue since this is not critical */
+	if (si->tos > 0) optval = si->tos;
```
bogdan-iancu, line 202: same change here too

cfg.y
nikbyte: Thank you for reviewing. 0 here means "not set TOS".

bogdan-iancu: you say `setsockopt()` fails with a 0 TOS value ?

nikbyte: By default, OpenSIPS sets the TOS value globally on all interfaces, and this behavior cannot be disabled.

This patch allows a TOS value of 0 to prevent TOS from being set at all.
In other words, if the global option tos=0 is configured, OpenSIPS will not apply TOS to the socket.

This change helps avoid validation issues when parsing the OpenSIPS script.

nikbyte: Yes, exactly. By default, OpenSIPS sets TOS, which is not always the desired behavior. Setting TOS=0 disables it. We should have the option to avoid setting TOS both globally and on specific sockets.

net/net_tcp.c
nikbyte: fixed, thank you

net/net_udp.c
nikbyte: fixed, thank you

bogdan-iancu:
Thank you @nikbyte , it looks good to me, I will merge and update the 3.6 docs

PR #3575, opened 2025-02-09
Fix avpops memory leak in avp_db_load / avp_db_delete / avp_db_store (for 3.4)

nikbyte:
**Summary**
This pull request fixes a memory leak in the avpops module (function fixup_db_avp) where a dynamically allocated string (sp->u.s.s) was not released in certain error paths or after normal fixup usage.

**Details**
When parsing script parameters for avp_db_load() / avp_db_delete() / avp_db_store(), if the first parameter was a plain string (not prefixed by '$'), the code allocated memory for sp->u.s.s. However, it never got properly freed:
1. In the error path (goto err_free:), only sp was freed, leaving sp->u.s.s allocated.
2. In the success path, there was no custom fixup_free_* function to release sp->u.s.s.

Repeated calls thus caused incremental memory leaks.

**Solution**
1. Introduced a new free function fixup_free_avp_source() that properly frees both sp->u.s.s (if allocated) and the sp structure.
2. Updated the module to use this free function instead of the standard fixup_free_pkg().

By doing so, all allocated memory is correctly released, eliminating the leak.

**Compatibility**
No breaking changes or behavioral differences other than fixing the memory usage issue. No impacts on SIP interoperability or script compatibility.

### DIFF
--- a/modules/avpops/avpops.c
+++ b/modules/avpops/avpops.c
@@ -106,6 +106,7 @@ static int fixup_free_check_avp_p2(void** param);
 static int fixup_free_avp_subst_p1(void** param);
 static int fixup_free_avp_subst_p2(void** param);
 static int fixup_free_pvname_list(void** param);
+static int fixup_free_avp_source(void** param);
 static int fixup_free_avp_dbparam(void** param);
 static int fixup_avp_shuffle_name(void** param);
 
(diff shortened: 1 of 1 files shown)


nikbyte:
Unit tests can't be passed because of:
```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`
```

bogdan-iancu:
Good catch @nikbyte , thank you !

PR #3590, opened 2025-02-25
Fix dbid password special chars

edman80:
**Summary**
Allow additional special characters in database passwords.

**Details**
We found that certain characters in database passwords would cause the state machine to terminate reading of the password and move into the next state. Specifically forward slashes. This commit aims to allow more special characters. The only character which would terminate password reading is @. To support @ symbol, further refactor would be needed to escape literal characters.
**Solution**
Added additional ST_PASSWORD state.

**Compatibility**
It does not seem to break normal scenarios and we have built many tests into this PR to show that.

**Closing issues**
n/a


liviuchircu:
LGTM, good work!  Backwards-compatible addition, backed up by tests, all tests passing.  Merging to `master`.

PR #3594, opened 2025-03-06
Fix sql_query required parameters

NormB:
**Summary**

The sql_query() function definition is incorrect as it causes a failure when the query and two optional parameters are supplied. This PR fixes the problem.

**Details**

The documented function call is:

sql_query(query, [res_col_avps], [db_id])

When specifying all three parameters, OpenSIPS fails to start because the parsing logic detects an incorrect function call.

**Solution**

This PR updates the function definition so that if the one required and two optional are specified, the parsing is successful.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/sqlops/sqlops.c
+++ b/modules/sqlops/sqlops.c
@@ -164,6 +164,7 @@ static const cmd_export_t cmds[] = {
 		ALL_ROUTES},
 
 	{"sql_query", (cmd_function)w_sql_query, {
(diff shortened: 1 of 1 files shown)


edman80:
Looks like this may have regressed in #9062d8e

liviuchircu:
Good catch @NormB @edman80 ... it looks like some kind of mouse-slip / line-delete operation which got included in that commit by mistake.  Will also backport to 3.5 if needed.

PR #3598, opened 2025-03-11
Preserving the flags of the topology_hiding in non dialog mode in thi…

davidtrihy-genesys:
**Summary**
Encodes the topology hiding flags into the thinfo param in the Contact header to be used for sequential requests

**Details**
SImply encodes the flags directly into the thinfo param, only TOPOH_KEEP_USER is used for transaction only topology hiding but this solution allows for any flags to be used moving forward, have not tested using sipsert but my own internal tooling and this solution works

**Solution**
Encodes the flags into the thinfo param and retrieves them on sequential request

**Compatibility**
Will change the behaviour of any configuration which assumes on sequential requests using non dialog topology hiding that user will not be preserved, this will preserve the user on sequentials

### DIFF
--- a/modules/topology_hiding/topo_hiding_logic.c
+++ b/modules/topology_hiding/topo_hiding_logic.c
@@ -23,6 +23,9 @@
 * -------
 *  2015-02-17  initial version (Vlad Paiu)
 */
+#define MAX_INT_LEN 11 /* 2^32: 10 chars + 1 char sign */
+
+#include <stdio.h>
 
 #include "topo_hiding_logic.h"
 
@@ -1551,12 +1554,14 @@ error:
 
 /* We encode the RR headers, the actual Contact and the socket str for this leg */
 /* Via headers will be restored using the TM module, no need to save anything for them */
-static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int *suffix_len)
+static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int *suffix_len, int flags)
 {
-	short rr_len,ct_len,addr_len,enc_len;
+	char flags_char[MAX_INT_LEN + 2];
+	short rr_len,ct_len,addr_len,flags_len,enc_len;
 	char *suffix_plain,*suffix_enc,*p,*s;
 	str rr_set = {NULL, 0};
 	str contact;
+	str flags_str;
 	int i,total_len;
 	struct sip_uri ctu;
 	struct th_ct_params* el;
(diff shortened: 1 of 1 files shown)


modules/topology_hiding/topo_hiding_logic.c
```diff
@@ -1666,6 +1675,10 @@ static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int
 		memcpy(p,contact.s,contact.len);
 		p+= contact.len;
 	}
+	memcpy(p,&flags_len,sizeof(short));
```
razvancrainea, line 1678: Shouldn't we add the length size (`sizeof(short)` ) in the `local_len`?

modules/topology_hiding/topo_hiding_logic.c
```diff
@@ -1596,8 +1601,12 @@ static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int
 		ct_len = (short)contact.len;
 	}
 
+	flags_len = (short)sprintf(flags_char, "%d", flags);
```
razvancrainea, line 1604: The practice is to use [int2str](https://github.com/OpenSIPS/opensips/blob/master/ut.h#L280) function to convert ints to string.

modules/topology_hiding/topo_hiding_logic.c
```diff
@@ -2100,7 +2114,15 @@ static int topo_no_dlg_seq_handling(struct sip_msg *msg,str *info)
 		free_rr(&head);
 	pkg_free(dec_buf);
 
-	if (topo_no_dlg_encode_contact(msg,0,NULL) < 0) {
+	if (flags_buf.len && flags_buf.s) {
+		if (sscanf(flags_buf.s, "%d", &flags) < 0) {
```
razvancrainea, line 2118: similarly, the practice is to use the [int2str](https://github.com/OpenSIPS/opensips/blob/master/ut.h#L720) function for this.

modules/topology_hiding/topo_hiding_logic.c
```diff
@@ -23,6 +23,9 @@
 * -------
 *  2015-02-17  initial version (Vlad Paiu)
 */
+#define MAX_INT_LEN 11 /* 2^32: 10 chars + 1 char sign */
```
razvancrainea, line 26: This is not needed if `int2str` is being used.

razvancrainea (changes requested):
Looks good, however I believe there is a bug in counting the length. The other two are optional changes, but nice to have to preserve the code practice.

modules/topology_hiding/topo_hiding_logic.c
davidtrihy-genesys: Added on line 1574

davidtrihy-genesys:
@razvancrainea I've applied all the comments you made and I actually fixed another limitation on replies within a dialog

modules/topology_hiding/topo_hiding_logic.c
```diff
@@ -924,6 +925,11 @@ static void th_no_dlg_onreply_within(struct cell* t, int type, struct tmcb_param
 	_th_no_dlg_onreply(t,type,param,0,0);
 }
 
+static void th_no_dlg_user_onreply_within(struct cell* t, int type, struct tmcb_params *param)
```
davidtrihy-genesys, line 928: There's probably a better way to do this by passing flags as part of the callback but given that there is only flag that is viable for no dialog topology hiding then I think this suffices, if more flags become the norm in this module then this can be revised but for now I think it works fine as it is and follows the same pattern as initial requests

PR #3599, opened 2025-03-11
Ensure proper memory deallocation in case of parsing errors

### LINKED ISSUE #3589: [BUG] Memory Leak when receiving malformed SDP lines
### Description

When OpenSIPS receives an INVITE with SDP lines incorrectly separated by `\n` only (instead of the standard `\r\n`) and then uses sdp.line transformation, it can end up concatenating two lines. For example:

```
c=IN IP4 8.8.8.8
t=0 0
```

In a well-formed SDP, `c=` and `t=` lines would be separated by `\r\n`. However, here they're effectively joined into one token that OpenSIPS sees as:

```
c=IN IP4 8.8.8.8
t=0 0
```

From the configuration side, a snippet like:

```opensips
$avp(sdp_ip) := $(rb{sdp.line,c,0}); 
avp_subst("$avp(sdp_ip)", "/^c=IN IP[46] (.*)$$/\1/");
```

will produce a concatenated line rather than just the `c=` line. Subsequently, if we insert that resulting string into a SIP header, e.g.,

```opensips
append_hf("X-SDPIP: $avp(sdp_ip):$avp(sdp_port)\r\n");
```

we end up with a new message like:

```
X-HCV-OrigSDP: 8.8.8.8
t=0 0:21022
```

in the outgoing INVITE. Now OpenSIPS itself, upon receiving (or re-processing) this generated INVITE, sees `t=0 0` as a short header name `t:` (similar to `To:`) and tries to parse it. This leads to a parsing error and triggers a **memory leak** where the parser-allocated memory for this invalid header does not get correctly freed.

### Steps to Reproduce

1. Send an INVITE to OpenSIPS with a c-line ending by `\n` instead of `\r\n`.
2. Let OpenSIPS parse that SDP and store it in an AVP.
3. Use sdp.line transformation to manipulate the combined line, then inject it back into an outgoing header (e.g., via `append_hf()`).
4. OpenSIPS regenerates an INVITE containing lines like:

   ```
   X-HCV-OrigSDP: 8.8.8.8
   t=0 0:21022
   ```

5. Now OpenSIPS re-parses its own generated INVITE. The partial line `t=0 0:21022` is detected as an invalid short `To:` header, and that parse path causes a memory leak.

### Analysis of the Problem

- The root cause is a malformed SDP line that merges `c=` and `t=` with only a `\n`.
- OpenSIPS fails to split them as separate lines in the script, and they get written back into a new INVITE.
- During the subsequent parse attempt, the line `t=0 0:21022` is misread as a SIP header. The parser attempts to allocate memory to parse it, fails, and does not free what it allocated (i.e., a parse-time memory leak).
- This behavior eventually can degrade OpenSIPS, especially under high call rates or repeated occurrences of malformed lines.

### Proposed Fix / Patch

A patch has been developed that ensures memory allocated for malformed or partially parsed headers is always freed, preventing leaks even if the line is invalid. The patch does **not** change how OpenSIPS treats non-standard SDP lines logically; it only addresses the memory leak that occurs during the parse error.

```
diff --git a/parser/msg_parser.c b/parser/msg_parser.c
index ad84ac611..b02473e28 100644
--- a/parser/msg_parser.c
+++ b/parser/msg_parser.c
@@ -840,6 +840,9 @@ int parse_msg(char* buf, unsigned int len, struct sip_msg* msg)
   return 0;

 error:
+  // Free the SIP message if it has already been created to prevent memory leaks.
+  free_sip_msg(msg);
+
   /* more debugging, msg->orig is/should be null terminated*/
   LM_ERR("message=<%.*s>\n", (int)len, ZSW(buf));
   return -1;
```

### Additional Notes

- While RFCs dictate that `\r\n` must be used to separate lines, some upstream systems may produce only `\n`. This is not RFC-compliant, but real-world scenarios often violate specs.
- The snippet using sdp.line transformation is correct for normal well-formed inputs but leads to unpredictable results with malformed lines.

### Next Steps

- Merge the patch that fixes the memory leak in the parse procedure, ensuring that any allocated parse buffers are correctly freed.
- Optionally, consider implementing an additional check in sdp_parse() to detect and reject such SDPs, preventing malformed SIP requests/replies from being processed.

nikbyte:
**Summary**
This PR addresses a memory leak issue in OpenSIPS that occurs when message parsing fails under certain conditions. Currently, in some failure scenarios within parse_msg(), allocated memory is not properly freed, leading to potential leaks. This patch ensures proper memory deallocation in case of parsing errors.

**Details**
The parse_msg() function is responsible for processing incoming SIP messages. However, its behavior regarding memory management on failure is inconsistent across different parts of the OpenSIPS codebase—sometimes the SIP message structure is freed, while in other cases, it is not handled explicitly.

This can result in memory being allocated but never released when parsing errors occur, leading to a slow memory leak. Over time, under high traffic conditions, this could contribute to increased memory usage and degraded system performance.

**Solution**
This PR ensures that parse_msg() consistently frees allocated memory when a parsing failure occurs. Specifically, it:
	•	Adds explicit memory deallocation inside parse_msg() in case of failure.
	•	Prevents potential memory leaks caused by incomplete or erroneous message parsing.

While this fix does not change OpenSIPS’ parsing logic, it improves robustness by ensuring memory is properly handled in failure cases.

**Compatibility**
This patch does not introduce any functional changes to OpenSIPS’ behavior beyond improved memory management. It only ensures that resources are properly freed in failure scenarios. However, some minor adjustments might be needed in other parts of the codebase to align with the improved cleanup logic and avoid redundant free() calls.

**Closing issues**
closes #3589 partially

### DIFF
--- a/parser/msg_parser.c
+++ b/parser/msg_parser.c
@@ -872,7 +872,7 @@ int parse_msg(char* buf, unsigned int len, struct sip_msg* msg)
 	}
 	msg->unparsed=tmp;
 	/*find first Via: */
-	if (parse_headers(msg, flags, 0)==-1) goto error;
+	if (parse_headers(msg, flags, 0)==-1) goto parse_error;
 
 #ifdef EXTRA_DEBUG
 	/* dump parsed data */
@@ -923,6 +923,11 @@ int parse_msg(char* buf, unsigned int len, struct sip_msg* msg)
 #endif
 
 	return 0;
+
+parse_error:
+	/* Free the SIP message if it has already been created to prevent memory leaks */
+	free_sip_msg(msg);
+	memset(msg, 0, sizeof(*msg));
 
 error:
 	/* more debugging, msg->orig is/should be null terminated*/


parser/msg_parser.c
```diff
@@ -925,6 +925,9 @@ int parse_msg(char* buf, unsigned int len, struct sip_msg* msg)
 	return 0;
 
 error:
+	/* Free the SIP message if it has already been created to prevent memory leaks */
```
bogdan-iancu, line 928: tiny improvement, more on clarity than in performance. As only `parse_headers` may induce a potential leak (line 875 here), I would do there a `goto error2` and here, on top of the existing `error` label I would add the new `error2` doing the `free_sip_msg()` and continuing with the `error` sequence.

nikbyte: Thank you for the review. I fixed in updated commit.

bogdan-iancu:
@nikbyte , thanks for the update. We still have the problem with the double free, please see [my comment here](https://github.com/OpenSIPS/opensips/pull/3599#pullrequestreview-2678534045)

nikbyte:
Ok, I've got. Let me review it too. Thanks.

nikbyte:
I have reviewed all the places that call parse_msg(), but in some cases the logic still isn’t entirely clear. To avoid breaking anything, I propose simply adding memset(msg, 0, sizeof(*msg)) after the free_sip_msg(msg) call. This way, we reduce the risk of a double-free without altering the existing logic elsewhere.

bogdan-iancu:
Thank you @nikbyte , I think we are good to go here :)

PR #3600, opened 2025-03-12
Add min_ttl parameter to dns_cache

leosmith36:
**Summary**
This PR introduces a `min_ttl` parameter to the `dns_cache` module, allowing a minimum TTL to be enforced for cached DNS records.

**Details**
DNS records with very short TTLs can cause excessive re-queries to DNS servers, leading to increased load on DNS infrastructure, unnecessary latency, and degraded performance due to the blocking nature of DNS lookups. To mitigate these issues, it is beneficial to enforce a minimum TTL for cached records.

**Solution**
This PR enables applications using `dns_cache` to define a `min_ttl` parameter, specifying the minimum number of seconds a DNS record should be cached. If a DNS response provides a TTL lower than `min_ttl`, the record will instead be cached for `min_ttl` seconds.

### DIFF
--- a/modules/dns_cache/doc/dns_cache_admin.xml
+++ b/modules/dns_cache/doc/dns_cache_admin.xml
@@ -83,7 +83,7 @@ modparam("dns_cache", "blacklist_timeout",7200) # 2 hours
 		<title>Set <varname>min_ttl</varname> parameter</title>
 		<programlisting format="linespecific">
 ...
-modparam("min_ttl", "min_ttl",300) # 5 minutes
+modparam("dns_cache", "min_ttl",300) # 5 minutes
 ...
 		</programlisting>
 		</example>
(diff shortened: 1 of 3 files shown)


bogdan-iancu (changes requested):
Thanks @leosmith36 , I see this a useful addition. Please check my comments and update the PR, so I can go for the merge

modules/dns_cache/doc/dns_cache_admin.xml
```diff
@@ -70,6 +70,25 @@ modparam("dns_cache", "blacklist_timeout",7200) # 2 hours
...
+			The minimum number of seconds that a DNS record will be kept in
+			cache. If the TTL received in the DNS answer is lower than this
+			value, the record will be cached for min_ttl seconds.
+		</para>
+		
+		<example>
+		<title>Set <varname>min_ttl</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("min_ttl", "min_ttl",300) # 5 minutes
```
bogdan-iancu, line 86: Just a small typo here, in the name of the module. It should be 
```
modparam("dns_cache", "min_ttl",300)
```
Also, a good practice is to mention if it is optional or not and if YES, what is he default value of this new param - `0` meaning no min_ttl forcing.

bogdan-iancu (approved):
Good to go !

bogdan-iancu:
Thank you @leosmith36

PR #3607, opened 2025-03-24
Add Support for SIP-Digest-Authentication

LarryLaffer-dev:
**Summary**
Currently, the auth_aka module only supports AKAv1 or AKAv2 authentication schemes. This patch allows the use of standard MD5 authentication as well - so basically any standard SIP-Client can register with credentials received from the HSS.

**Details**
- Add MD5 as supported algorhitms
- Build Authorize Header depending on AKA or MD5 algorhithms

**Compatibility**
- Tested with PyHSS and Twinkle and Zoiper as SIP-Clients


LarryLaffer-dev:
Two reasons:
1) Right now, we just needed MD5 for legacy SIP-Clients.
2) I did not look into if we need to change something else in the Authorization header (likely not, but not checked)

But good point, it could be rather easily extended that way.

PR #3608, opened 2025-03-24
Fixes to PUA and related Usrloc-changes

LarryLaffer-dev:
**Summary**
This PR fixes some crashes identified in the pua_usrloc module. In order for this module to provide the relevant information, Usrloc needs to store additional information (Parameters from Contact), which is now incorporated as well.

**Details**
Since the body may become large, the PR introduces a variable String-Buffer to facilitate larger results. Since the result should include the params of the Contact-Header, the according Params are retrieved (and stored) from usrloc.

**Solution**
I did not fully test the clustering and DB-operations of data yet.

**Compatibility**
This PR changes the schema of the location DB (added location column), so it is not backward compatible.

### DIFF
--- a/db/schema/location.xml
+++ b/db/schema/location.xml
@@ -9,7 +9,7 @@
 
 <table id="location" xmlns:db="http://docbook.org/ns/docbook">
     <name>location</name>
-    <version>1013</version>
+    <version>1014</version>
     <type db="mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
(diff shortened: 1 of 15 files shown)


bogdan-iancu:
@LarryLaffer-dev , thanks for your contribution. In order to be properly evaluated, you will need to break it into smaller pieces, which can be digested by itself. Like please open a separate PR for each pua_usrloc crash, we can understand the issue and the fix (without mixes from other crashes). As for the extensions, similar, please create a dedicated PR for each of them. Many thanks!

PR #3609, opened 2025-03-24
Make the Server-URI configurable (+ docs)

LarryLaffer-dev:
**Summary**
Minor change: Make the Server-URI configurable (+ docs)

**Details**
The Server-URI was using the Realm. The HSS can store the information, so an improper URI was returned in some cases.

**Solution**
Make the Server-URI configurable.


razvancrainea:
This looks good, but I'd rather make it a bit more flexible. For example, if not explicitly set, build it from the realm, by adding the `sip:` and making it an URI.

PR #3616, opened 2025-03-30
Add multi-lock allocator

vladpaiu:
**Summary**
For memory intensive OpenSIPS scripts ( heavy AVP, dialog vars, ACC usage, etc ), most of the CPU is wasted in contention on the global shm lock. 

The PR adds a new memory allocator, which at the core is a F_MALLOC clone, with the only difference that the SHM memory pool is split between TOTAL_F_PARALLEL_POOLS ( each pool with it's own SHM memory chunk and own lock ).

**Details**
At memory allocation, a random chunk will be picked. At memory freeing/reallocation, the same memory chunk will be used.

**Solution**
We are aiming to relieve contention at the global shm_lock by splitting it across multiple locks & separating the memory chunks entirely. 

The new allocator makes no sense for PKG memory. Also, the memory pool is assumed to be sufficiently big so that it can be safely split into TOTAL_F_PARALLEL_POOLS chunks

**Compatibility**
Should be backwards compatible. If OpenSIPS is started without any memory allocators selected, the default allocator ( F_MALLOC ) will be used. Explicitly selected the new allocator can be done via 

```
/opensips -f etc/opensips.cfg -a F_PARALLEL_MALLOC -k F_MALLOC -s F_PARALLEL_MALLOC -m 1024 -w
```


ovidiusas:
Hello @vladpaiu! This looks interesting! Thank you for the contribution!
Can you share some metrics on the performance improvements?

vladpaiu:
> Hello @vladpaiu! This looks interesting! Thank you for the contribution! Can you share some metrics on the performance improvements?

hey Ovidiu, I will be discussing this in the OpenSIPS summit, for now I can say the difference is significant, depending on the memory usage pattern of the OpenSIPS script. 

To load test this, I am using a script with 32 UDP worker that uses TM & Dialog & Topology hiding & ACC, which is just setting 30 different AVPs at INVITE time, 20 different $dlg_vals at invite time, then storing all those 50 values as $acc_extra ( logging fully disabled, traffic is on the lo interface, so it's raw CPU usage )

Testing with SIPP ( default UAC & UAS scenarios) on my laptop with 2000 calls per second, the F_MALLOC cannot keep up ( load: stats go to 100, UDP buffer gets full, packets get lost, SIPP calls start to drop ).

When using F_PARALLEL_MALLOC, 2000 calls per second get processed just fine, with "load:load": 10

bogdan-iancu:
Hey @vladpaiu , have you tested this new allocator against the existing HP_MALLOC - this is also a  multi-locks manager, trying to deal with the same problem as in your case.

vladpaiu:
> Hey @vladpaiu , have you tested this new allocator against the existing HP_MALLOC - this is also a multi-locks manager, trying to deal with the same problem as in your case.

I did, for my testing ( with a cold start ) HP_MALLOC is performing a bit worse compared to F_MALLOC as all allocated memory chunks are similar in size.

bogdan-iancu:
Thank you @vladpaiu 
@liviuchircu , it is all yours for review and merging.

liviuchircu (approved):
Fantastic work, @vladpaiu !  Found some time to look at this, and I have to say it really looks clean!

There is full `Makefile.conf.template` integration, similar to the other allocators, allowing flexibility at build time.  Also, `opensips -V` feature-flags support, as well as command-line options support.  I did notice a lack of cleanup on shutdown of the blocks, but I don't think it's a big deal.  Also, GPL headers missing from some (all?) of the new files, but that's quick to fix.

Out of curiosity, should we take a look at the `rand()` function overhead and perhaps consider changing it with a simple round-robin global?  Hopefully, I'm not prematurely optimizing (how many malloc() calls are there on each SIP message after all?  This is pre-SHM lock phase we're talking about, so no contention).

vladpaiu:
hey @liviuchircu thanks for looking into this.

added GPL headers ( https://github.com/OpenSIPS/opensips/pull/3616/commits/2d4926c40b3cf6ed64280c0a9f8b133306234bcd ) and fixed a bug related to HP_MALLOC usage ( https://github.com/OpenSIPS/opensips/pull/3616/commits/b07c227595a87f89abe0b46b33889991a64cc007 ) 

I used rand() on purpose here to ensure some sort of normal distribution, that is done outside of any locks so yeah think we are good for an initial version.

vladpaiu:
@liviuchircu fixed pipeline checks.

PR #3617, opened 2025-04-01
Add support to send a command to an RTPEngine specified by the sock_var

NormB:
**Summary**

Many rtpengine_* functions accept a parameter that will be populated with the socket of the RTPEngine chosen for the particular operation.  For example, the rtpengine_offer() function accepts a "sock_var" parameter. The "sock_var" variable is populated correctly. The ability to use the "sock_var" in subsequent rtpengine_* function calls has not been available.  This PR adds the ability to have the "sock_var" used to determine the specific RTPEngine that the operation is to be directed towards.

**Details**

When using a pool of RTPEngines (and/or a pool of OpenSIPS servers) and enabling services such as early media along with bridging multiple call legs it is possible that an OpenSIPS server may not be able to use the call-id to locate which RTPEngine the legs of a call are on. Most of the time, the call-id lookups that rtpengine.c use work well. In some advanced scenarios, the lookup will send the RTPEngine command to an incorrect RTPEngine sever.

This PR provides an ability for the script writer to store the chosen RTPEngine in the "sock_var" variable for an rtpengine_play_media() command and then use the "sock_var" in the rtpengine_stop_media() command.

The format of the data stored in "sock_var" is: proto:ip:port so it is possible for the script writer to create and use a "sock_var" without having it first stored by one of the rtpengine_* functions. Clearly, the script writer needs to have some mechanism to select the correct RTPEngine server.

**Solution**

The rtpengine.c module is designed in a modular way with subroutines for each rtpengine_* call. All of these subroutines eventually call rtpe_function_call() where the specific RTPEngine node is determined. This PR adds additional logic to rtpe_function_call() to support using the "sock_var" if it has been specified and is non-empty.

In addition, an important improvement has been made to the rtpe_test() subroutine. During the development of this PR a memory corruption / leak was identified. Two different binary encoding (bencode) routines where called, one for requests and the other for replies. Each of these routines allocates space on the bencode_buffer. The problem is that each of these routines stored the resulting allocated memory address in the same variable.  Creation of a second variable so that requests and replies are stored separately resolved this issue.

**Compatibility**

There is no change to any logic when storing the chosen RTPEngine in a specified "sock_var".

If a script specifies a "sock_var" and that variable is non-NULL, then it will be used to determine the RTPEngine server to direct the command to.

If the script writer wants to specify a "sock_var" parameter, but does not want it to be used then the "sock_var" variable should not be populated. In this case, it will be populated when one of the rtpengine_* functions has chosen an RTPEngine server.

**Closing issues**

### DIFF
--- a/modules/rtpengine/rtpengine.c
+++ b/modules/rtpengine/rtpengine.c
@@ -2836,27 +2836,22 @@ static bencode_item_t *rtpe_function_call(bencode_buffer_t *bencbuf, struct sip_
 	if (rtpe_function_call_prepare(bencbuf, msg, op, &ng_flags, flags_str, body_in, extra_dict,&err) < 0)
 		goto error;
 
-	/*
-	 * If the Set is not specified as a parameter into this routine
-	 * try to determine. Using the specified setid_avp module parameter,
-	 * or from the default default_rtpe_set().
-	 */
+	/*** set can be passed into this routine. If it is NULL, then attempt to determine it. ***/
 	if (!set) {
-		set_rtpengine_set_from_avp(msg);
 		if ((set=rtpe_ctx_set_get())==NULL)
 			set = *default_rtpe_set;
+	}
+
+	/*** If the spvar "sock_var" has been specified, parse it into a (socket_val) STR variable ***/
+	if (spvar) {
+		memset(&socket_val, 0, sizeof(pv_value_t));
+		pv_get_spec_value(msg, spvar, &socket_val);
 	}
 
 	failed_node = NULL;
 
 	RTPE_START_READ();
 	do {
-		/* If the spvar "sock_var" has been specified, parse it into a (socket_val) STR variable */
-		if (spvar) {
-			memset(&socket_val, 0, sizeof(pv_value_t));
-			pv_get_spec_value(msg, spvar, &socket_val);
-		}
-
 		/*
 		 * Many rtpengine_* commands allow a "sock_var" to be specified by the script writer.
 		 * The "sock_var" will be populated by the rtpengine_* command with the RTPEngine that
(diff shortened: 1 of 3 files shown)


NormB:
@razvancrainea @bogdan-iancu This update offers the possibility of using a script to cause the proxy to direct RTPEngine commands to any RTPEngine that has previously been defined and is operational. This should be used with the same caution as any other function that the proxy provides.

Norm

@liviuchircu

modules/rtpengine/rtpengine.c
```diff
@@ -2827,30 +2827,61 @@ static bencode_item_t *rtpe_function_call(bencode_buffer_t *bencbuf, struct sip_
...
+			set = *default_rtpe_set;
+	}
 
 	failed_node = NULL;
 
 	RTPE_START_READ();
 	do {
-		if (snode && snode->s) {
+		/* If the spvar "sock_var" has been specified, parse it into a (socket_val) STR variable */
+		if (spvar) {
```
razvancrainea, line 2855: don't see a reason to check this under loop - simply "resolve" the socket before starting the loop and assign it to `snode`

modules/rtpengine/rtpengine.c
```diff
@@ -2827,30 +2827,61 @@ static bencode_item_t *rtpe_function_call(bencode_buffer_t *bencbuf, struct sip_
...
-	/*** send it out ***/
-	if (!set && (set=rtpe_ctx_set_get())==NULL )
-		set = *default_rtpe_set;
+	/*
+	 * If the Set is not specified as a parameter into this routine
+	 * try to determine. Using the specified setid_avp module parameter,
+	 * or from the default default_rtpe_set().
+	 */
+	if (!set) {
+		set_rtpengine_set_from_avp(msg);
```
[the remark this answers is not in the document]
NormB: Removed the call to set_rtpengine_set_from_avp()

modules/rtpengine/rtpengine.c
NormB: Moved the conversion from the socket char to str outside of the loop.

NormB:
Successful testing after requested updates were implemented.

Issue #3539, opened 2024-12-10
[BUG] Memory leak in rtp_relay


okhowang:
**OpenSIPS version you are running**

```
version: opensips 3.4.7 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 79ebc9d
main.c compiled on  with gcc 12
```

**Describe the bug**

I migrate route script from `rtpengine` to `rtp_relay`, and shm memory increase continuously

**To Reproduce**

**Expected behavior**

**Relevant System Logs**

output of `opensips-cli -x mi mem_shm_dump -1`
```
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [../../modules/tm/../../rw_locking.h: lock_init_rw, line 45]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [../../rw_locking.h: lock_init_rw, line 45]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [../tm/../../rw_locking.h: lock_init_rw, line 45]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [address.c: init_address_part, line 398]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [address.c: init_address_part, line 410]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [dlg_timer.c: init_dlg_ping_timer, line 174]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [dlg_timer.c: init_dlg_reinvite_ping_timer, line 204]
2024-12-10 14:16:09 Dec 10 14:16:09 [22] ERROR:          8 : 1 x [dprint.c: init_log_level, line 920]
[...]

PR #3623, opened 2025-04-14
Fix: rtp_relay memory leak

Kicey:
**Summary**

To fix rtp_relay memory leak metioned in https://github.com/OpenSIPS/opensips/issues/3539.

**Details**

The rtp_relay_ctx is not released when the opensips receive uas response with status >= 300. In which case the delete request to rtpengine/rtpproxy is sent and session is freed. But the rtp_relay_ctx pointer is not put into dialog. So when dialog destory, it will not release the corresponding rtp_relay_ctx.

**Solution**

Retrive the dialog and put the ctx pointer in. So that when destroy dialog, the rtp_relay_ctx will be released.

**Compatibility**

The change don't break other scenarios. It work well on opensips 3.4.11 .

### DIFF
--- a/modules/rtp_relay/rtp_relay_ctx.c
+++ b/modules/rtp_relay/rtp_relay_ctx.c
@@ -1879,6 +1879,13 @@ static void rtp_relay_ctx_initial_cb(struct cell* t, int type, struct tmcb_param
 					rtp_relay_ctx_leg_reply(ctx, p->rpl, t, sess, RTP_RELAY_CALLEE);
 					break;
 				case 1:
+					struct dlg_cell *dlg;
+					dlg = rtp_relay_dlg.get_dlg();
+					if (!dlg) {
+						LM_ERR("could not find dialog!\n");
+					} else{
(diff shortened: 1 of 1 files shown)


razvancrainea:
Thanks for the report and finding this corner case!
I believe your fix is more or less OK, but I don't see the point setting it on failure - why not setting it all the way from the start? not sure what are the implications though, I will have to do some more investigation to figure it out.

Kicey:
> Thanks for the report and finding this corner case! I believe your fix is more or less OK, but I don't see the point setting it on failure - why not setting it all the way from the start? not sure what are the implications though, I will have to do some more investigation to figure it out.

I think this may because the rtp_relay exports the rtp_relay_ctx as pseudo-variables, so it doesn't put the pointer until all fields are ready (such as to_tag). I'm not sure if it will cause another problem if we setting it from the start. I'm goint to try setting it from the  very start.

razvancrainea:
I double checked the code, and if the context is not moved in the dialog, it remains in the transaction, and gets removed when the transaction is deleted. Are you sure there is actually a leak? Can you send some debug logs so we can track this down?

Issue #3622, opened 2025-04-11
[BUG] [TCP_main] trying to del fd 153 with flags 2 1


kadabusha:
**OpenSIPS version you are running**

```
$ opensips34 -V
version: opensips34 3.4.11 (x86_64/freebsd)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, DBG_MALLOC, USE_UMUTEX
MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, select, kqueue.
git revision: d16321cbf
main.c compiled on 08:47:13 Apr 11 2025 with cc 14.0.5
```
Config file used - just for the case:
[opensips34.txt](https://github.com/user-attachments/files/19712586/opensips34.txt)

**Describe the bug**
For the case HEP is enabled with TCP protocol, and the collector does not process the incoming traffic properly,
the export gets into TCP_main bug, causing logs to flood pretty quick.

**To Reproduce**
Collector was set to 
```
abc.com:123;version=3;transport=tcp
```
That's just fuzzy address - no listener is there. After few minutes the messages like below get into logs, and the log file bloats, getting few GBs in a dozen of minutes.

**Expected behaviour**
Case collector is not responding, stop the export after some timeout. Maybe retry after another time passed?

**Relevant System Logs**
```
Apr 11 10:14:09 jenv_1 /usr/local/sbin/opensips34[32533]: ERROR:core:tcp_async_add_chunk: We have reached the limit of max async postponed chunks 32
Apr 11 10:14:09 jenv_1 /usr/local/sbin/opensips34[32533]: ERROR:proto_hep:hep_tcp_send: Failed to add another write chunk to 0x82ea4fab8
Apr 11 10:14:09 jenv_1 /usr/local/sbin/opensips34[32533]: ERROR:proto_hep:msg_send: send() to 18.239.18.36:123 for proto hep_tcp/10 failed
Apr 11 10:14:09 jenv_1 /usr/local/sbin/opensips34[32533]: ERROR:proto_hep:send_hep_message: Cannot send hep message!
Apr 11 10:14:09 jenv_1 /usr/local/sbin/opensips34[32533]: ERROR:tracer:send_trace_proto_duplicate: failed to forward message to destination!
[...]

PR #3627, opened 2025-04-15
Add max_retries and cooldown timer to send_hep

rita7lopes:
**Summary**
Whenever we use TCP as the transport to send the hep data to either Homer or Hepic, and the connection cannot be established because the endpoint is unreachable or down, opensips keeps in a loop trying to send the data. 
Since it fails to send, it writes the data to the disk and keeps trying, eventually using all the workers available and leading opensips to crash.
Sending hep is not the central point of Openips, so it should never crash because the hep destination is unavailable or unreachable.
This PR aims to fix that.

**Details**
This is a bug fix that currently impacts anyone using hep over TCP. All endpoints have maintenance, and network instability happens from time to time. 
This PR prevents Opensips from crashing because of the hep module.

**Solution**
Add a cool-down timer as a variable (allow for customisation) to delay the next connection re-establish attempt after the max_retries (another variable that allows customisation) is hit. 

**Compatibility**
This is a change-specific and isolated in hep module - it's 100% retro-compatible with the other versions.

### DIFF
--- a/modules/proto_hep/README
+++ b/modules/proto_hep/README
@@ -246,7 +246,34 @@ modparam("proto_hep", "hep_async_max_postponed_chunks", 16)
 modparam("proto_hep", "hep_capture_id", 234)
 ...
 
-1.3.10. hep_async_local_connect_timeout (integer)
+1.3.10. hep_retry_cooldown (integer)
+
+   This parameter defines how many seconds OpenSIPS should wait 
+   before retrying a TCP connection to the HEP destination after 
+   reaching the maximum number of failed attempts set by 
+   hep_max_retries. Limitation: 16-bit integer.
+
+   Default value is "3600".
+
+   Example 1.10. Set hep_retry_cooldown parameter
+...
+modparam("proto_hep", "hep_retry_cooldown", 60)
+...
+
+1.3.11. hep_max_retries (integer)
+
+   This parameter defines the maximum number of attempts 
+   OpenSIPS will make to establish a TCP connection with 
+   the HEP destination. Limitation: 16-bit integer.
+
+   Default value is "5".
+
+   Example 1.11. Set hep_max_retries parameter
+...
+modparam("proto_hep", "hep_max_retries", 10)
+...
+
+1.3.12. hep_async_local_connect_timeout (integer)
 
    If hep_async is enabled, this specifies the number of
    milliseconds that a connect will be tried in blocking mode
@@ -256,12 +283,12 @@ modparam("proto_hep", "hep_capture_id", 234)
 
    Default value is 100 ms.
 
-   Example 1.10. Set hep_async_local_connect_timeout parameter
+   Example 1.12. Set hep_async_local_connect_timeout parameter
 ...
 modparam("proto_hep", "hep_async_local_connect_timeout", 200)
 ...
 
-1.3.11. hep_async_local_write_timeout (integer)
+1.3.13. hep_async_local_write_timeout (integer)
 
    If hep_async is enabled, this specifies the number of
    milliseconds that a write op will be tried in blocking mode
@@ -271,7 +298,7 @@ modparam("proto_hep", "hep_async_local_connect_timeout", 200)
 
    Default value is 10 ms.
 
-   Example 1.11. Set hep_async_local_write_timeout parameter
+   Example 1.13. Set hep_async_local_write_timeout parameter
 ...
 modparam("proto_hep", "hep_async_local_write_timeout", 100)
 ...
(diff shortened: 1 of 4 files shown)


liviuchircu (approved):
Great work, @rita7lopes!  There are many ways we can make this "HEP cooldown" support more practical and more sysadmin-friendly.  To name a few problems which come to mind:

* if a sysadmin brings up the Homer server, they will be unable to get the HEP tracing flowing again without waiting 1 hour, assuming they avoid restarting the OpenSIPS, during traffic hours.  There is no MI interaction to help here.
* if I am tracing to multiple HEP destinations, triggering the HEP cooldown on a single destination will impact all of them, by disabling all HEP sending at `proto_hep` global module level

Still, there is a huge difference between impacting the SIP service and impacting the SIP tracing service -- if one is down, you get customer complaint calls, if the other is down, it has a very good chance of not causing any impact at all.

So I fully approve merging this initial "safety net" support as-is and backporting it.  We can _refine_ the operational details later, as we get more feedback.

sobomax:
Hi, @rita7lopes @liviuchircu we've tested the patch quickly, it reduces the issue somewhat however we still see a lot of those `BUG - [TCP_main] trying to del ...` messages in the syslog. I suspect it does not close the connection properly on a failure to connect or something.

See also Issue #3622 by @kadabusha

```
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:send_trace_proto_duplicate: failed to forward message to destination!
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:save_siptrace: Failed to duplicate with hep to <abc.com:123>
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:proto_hep:send_hep_message: HEP send suppressed: too many failures (8), in cooldown (59
 seconds left)
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:send_trace_proto_duplicate: failed to forward message to destination!
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:save_siptrace: Failed to duplicate with hep to <abc.com:123>
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:proto_hep:send_hep_message: HEP send suppressed: too many failures (8), in cooldown (59
 seconds left)
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:send_trace_proto_duplicate: failed to forward message to destination!
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:save_siptrace: Failed to duplicate with hep to <abc.com:123>
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:proto_hep:send_hep_message: HEP send suppressed: too many failures (8), in cooldown (59
 seconds left)
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:send_trace_proto_duplicate: failed to forward message to destination!
Apr 21 03:00:53 jenv_1 /usr/local/sbin/opensips34[6755]: ERROR:tracer:save_siptrace: Failed to duplicate with hep to <abc.com:123>
Apr 21 03:01:02 jenv_1 /usr/local/sbin/opensips34[6780]: ERROR:core:io_watch_del: BUG - [TCP_main] trying to del fd 157 with flags 2 1
Apr 21 03:01:02 jenv_1 syslogd: last message repeated 6175 times
```

liviuchircu:
Thanks for the update @sobomax !  Any specific settings/behavior on that dummy TCP server that you are using, so we can get a reproduce here ?  Or maybe on the UAC side, in terms of CPS perhaps?  I tried with both a `nc -l` TCP server and with a small Python script which only does `.accept()` in a loop, and while I could reproduce the "max async postponed chunks" errors, the TCP Main fd bug state never occurred.  Thank you!

sobomax:
Oh, there is no "server". This is just a dymmy address that you can just use in your script: `abc.com:123;version=3;transport=tcp`

liviuchircu:
Hello!  There is a hidden side-effect to the HEP cooldown that we discovered, which is why I didn't merge this sooner:  if just a few processes trigger the cooldown, but other don't (`heplify-server` comes back online), then you have 1 full hour of _incomplete_ HEP traces.  Imagine random UDP workers processing SIP normally, but only _some_ have their HEP on cooldown.  Leading to random SIP packets being missing from the Homer visualizations, which is a somewhat strange and unwanted effect.

BUT, still, thanks to this PR, service is ONLINE and customers won't open tickets about missing ACK/BYE from Homer.  Case closed for now.  Merged.

PS: @sobomax if you could help me on reproducing that "ERROR:core:io_watch_del: BUG - [TCP_main] trying to del fd 157 with flags 2 1" spam (perhaps a Docker image, or _opensips.cfg_ script and sipp scenario), that would help a ton!  Thanks !

PR #3632, opened 2025-04-22
Fix: siprec segfault when copy delete

Kicey:
**Summary**

To fix issue https://github.com/OpenSIPS/opensips/issues/3602.
In short, the siprec unreference the dialog recorded too early. After releasing the memory of the rtp_relay_ctx in dialog, check of the dialog (by pointer) is not valid. Do failover and end the recording(copy delete) both access the dialog (access call id, rtp relay context, etc). So, I try to unreference the dialog when the recording dialog ends.

**Details**

The code following will unreference the dialog when the transaction end,
https://github.com/OpenSIPS/opensips/blob/b17f160696a5cb6c31f8d08de98b5fd9ccf579d4/modules/siprec/siprec.c#L320-L321
And if the recorded dialog end before the srs respond (before siprec get the b2b notify), the `srec_b2b_notify`'s processing of the dialog will be uncertain. So, the segment fault is likely to raise here (when invoke rtp_relay `copy_delete` or rebuild the invite body).

To simplify the code, I invoke the `src_unref_session` in the `srec_logic_destroy` function and use the `keep_sdp` flag. If this is inappropriate, please let me know.

**Solution**

Make siprec unref the dialog when the recording end.

**Compatibility**

The fix works well with 3.4.11.

**Closing issues**

### DIFF
--- a/modules/siprec/siprec.c
+++ b/modules/siprec/siprec.c
@@ -199,14 +199,6 @@ static void mod_destroy(void)
 {
 }
 
-static void tm_src_unref_session(void *p)
-{
-	struct src_sess *ss = (struct src_sess *)p;
-	srec_dlg.dlg_unref(ss->ctx->dlg, 1); /* release the dialog */
-	srec_hlog(ss, SREC_UNREF, "start recording unref");
-	SIPREC_UNREF(ss);
-}
-
 /*
  * function that simply prints the parameters passed
  */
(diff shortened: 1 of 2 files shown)


razvancrainea:
This should have been sorted out by 3c120dbb. If it is still crashing for you, please reopen with further details.

Issue #2811, opened 2022-04-15
[FEATURE] Support for a REDIS cluster


Lt-Flash:
**Is your feature request related to a problem? Please describe.**
It's both a feature request and a problem we've experienced. We were using a single instance of REDIS server without any issues for a long time. Recently we have deployed a REDIS cluster consisting of three MASTER nodes and three SLAVE nodes and run into an issue. REDIS protocol shards data between nodes in a cluster so when you're connected to M1 node (for example) but KEY you're looking for is on another node REDIS will respond with (for example):
```
MOVED 16281 192.168.0.14:6379
```

And as per REDIS documentation:
```
The error includes the hash slot of the key (3999) and the endpoint:port of the instance that can serve the query. The client needs to reissue the query to the specified node's endpoint address and port. The endpoint can be either an IP address, a hostname, or it can be empty (e.g. -MOVED 3999 :6380). An empty endpoint indicates that the server node has an an unknown endpoint, and the client should send the next request to the same endpoint as the current request but with the provided port.

Note that even if the client waits a long time before reissuing the query, and in the meantime the cluster configuration changed, the destination node will reply again with a MOVED error if the hash slot 3999 is now served by another node. The same happens if the contacted node had no updated information.
```

This behaviour disallows us to use REDIS cluster even with HAProxy or some other load balancer in front of it as the proxy can't analyze actual REDIS protocol and re-connect client to another REDIS node.

**Describe the solution you'd like**
Update OpenSIPS REDIS cache module to support recent version of REDIS cluster.

**Implementation**
  - Component: cachedb_redis
  - Type: List of REDIS nodes to connect to initially (it can be any master node)
  - Name: Up to developers

**Describe alternatives you've considered**
The only alternative is to continue using a single REDIS instance with a probable VRRP backup and some synchronization between them.

Thank you!

Lt-Flash:
Another idea - maybe it would be good to have all REDIS servers defined in a modparam or in DB so OpenSIPS could pre-open connections to all REDIS master/slave servers and as soon as it receives 'MOVED' response it can just use already open connection. If the server in 'MOVED' reply not listed then new connection needs to be open.

Lt-Flash:
Just a reference to Kamalio's module, it seem to support 'MOVED' operations:

https://www.kamailio.org/docs/modules/devel/modules/ndb_redis.html

ishvaram:
[...]

PR #3639, opened 2025-05-02
Add Redis MOVED Redirection support

NormB:
**Summary**

This update adds Redis MOVED Redirection support.

**Details**

When Redis is operating as a cluster, it will sometimes respond to a request with a MOVED reply. The contents of the MOVED reply contain a slot hash, endpoint and port that the request should be redirected toward. This is more fully described here: https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec/

This update partially supports issue: https://github.com/OpenSIPS/opensips/issues/2811

**Solution**

A request to Redis that is responded to with a MOVED will cause the redisReply to have an error state. When this error was received, the list of existing connections are attempted without any consideration to the endpoint returned in the redisReply.

A second issue was uncovered when issuing Redis requests with keys that contain hashtags which are parameters enclosed in braces {}. When hashtags are used, the slot hash is NOT computed using the entire key, but instead, the slot hash is determined using the contents of the first {hashtag}. The routine used to compare a slot hash against the range of slots in each Redis node does not take this into consideration.

This PR adds two updates to the error handling logic because a MOVED Redirection is returned as an error by Redis.

The first update parses the MOVED reply into it's components: slot, endpoint and port.
The second update uses the data from the MOVED reply components to perform a lookup for a node that matches the slot (range), endpoint and port.  If a match is found, the node is returned to the error handler and the request is retried.

**Compatibility**

This update is only triggered when an error redisReply is received with a MOVED Redirection. Existing errors are handled the way they were before so this update should not cause any problems with non-MOVED errors.

**Closing issues**

Updates #2811


NormB:
The comparison of the MOVED data does use the "slot" number in the reply.  This may be a little too aggressive in the compare logic. Redirecting the original request to the MOVED endpoint:port without taking the slot into consideration would be more lenient.

This update, as noted is one step toward providing cluster support. It is not a complete cluster implementation. For example:
When a node is added / removed from the cluster, the cached list of nodes must be updated which is outside of the scope of this change. The edge case when the node has an unknown endpoint, is also outside the scope of this change as it related to a cluster node changing state (or more accurately, changing port).

Below is the related documentation:

https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec/#moved-redirection

MOVED Redirection

A Redis client is free to send queries to every node in the cluster, including replica nodes. The node will analyze the query, and if it is acceptable (that is, only a single key is mentioned in the query, or the multiple keys mentioned are all to the same hash slot) it will lookup what node is responsible for the hash slot where the key or keys belong.

If the hash slot is served by the node, the query is simply processed, otherwise the node will check its internal hash slot to node map, and will reply to the client with a MOVED error, like in the following example:

GET x
-MOVED 3999 127.0.0.1:6381

The error includes the hash slot of the key (3999) and the endpoint:port of the instance that can serve the query. The client needs to reissue the query to the specified node's endpoint address and port.

An empty endpoint indicates that the server node has an unknown endpoint, and the client should send the next request to the same endpoint as the current request but with the provided port.

modules/cachedb_redis/cachedb_redis_utils.c
```diff
@@ -114,12 +114,14 @@ cluster_node *get_redis_connection_by_endpoint(redis_con *con, redis_moved *redi
 		return con->nodes;
 	} else {
 		for (it=con->nodes;it;it=it->next) {
-            if (match_prefix(redis_info->endpoint.s, redis_info->endpoint.len, it->ip, strlen(it->ip))) {
+			if (match_prefix(redis_info->endpoint.s, redis_info->endpoint.len, it->ip, strlen(it->ip))) {
 				if (it->port == redis_info->port) {
-					if (it->start_slot <= redis_info->slot && it->end_slot >= redis_info->slot) {
+					// Removed slot comparison as it may be a little too aggressive of a match
+					// Code is still here in the event that it needs to be added back in
+					//if (it->start_slot <= redis_info->slot && it->end_slot >= redis_info->slot) {
```
liviuchircu, line 121: Agree on removing the slot comparison for now, until the cluster "topology scanning" support is extended to be also performed at runtime, perhaps a one-time operation after each MOVED reply, ideally.

But isn't there still an issue with your MOVED implementation whenever we add a +1 node to the Redis Cluster?  All of a sudden, some of the OpenSIPS queries get a MOVED redirection to the new cluster node (the keys are balanced across all nodes).  But this new function you added, `get_redis_connection_by_endpoint()` tries to match the new endpoint against the existing list of nodes, which is the incomplete one (before the +1 node was added).  Won't the lookup fail?  Please correct me if I'm wrong.

liviuchircu (approved):

liviuchircu:
Merged, as step #1 towards full MOVED support.  Step #2 is related to the Redis Cluster configuration (including hash slots), and perhaps launching a `CLUSTER NODES` topology scanning command (+ redis node list rebuild) on each new MOVED command received.

PR #3653, opened 2025-05-23
domain: accept BIGINT type for accept_subdomain column

ovidiusas:
**Summary**
accept BIGINT type for accept_subdomain column

**Details**
If using a `VIEW `(instead of a regular `TABLE`) and populating the `accept_subdomain `column with a default value, in MySQL we have the option to convert to `UNSIGNED `and this will create a `BIGINT`.

Something like:
```
CREATE VIEW domain (
    id,
    domain,
    attrs,
    accept_subdomain,
    last_modified
) AS SELECT
    ...
   CAST(0 AS UNSIGNED),
   ...
FROM my_table;
```
Allowing `accept_subdomain `to be parsed as a BIGINT solves the issue.


ovidiusas:
Additional fix for domain table: 3 -> 4

PR #3654, opened 2025-05-26
Resolve the issue of incomplete transmission of large SIP messages over TLS protocol

36952362:
**Summary**
This PR fixes the issue of incomplete transmission of large SIP messages over TLS

**Details**
When OpenSIPS communicates with a remote UA over the TLS protocol, if the size of the SIP message large than `16,384 bytes`, OpenSIPS fails to send the entire SIP message.

if the DEBUG log level (`log_level=4`) is enabled, the logs show that OpenSIPS sends only `16,384 bytes` in the first write attempt. The remaining data is stored in the TCP connection’s `tcp_async_chunk`. However, OpenSIPS does `not` release the `tcp connection` back to the `TCP main` process, which would normally add it back to the `reactor` to listen again for the `write event`.

---log snippet---

`[26] DBG:proto_tls:proto_tls_send: sending via fd 4...
[26] DBG:tls_openssl:openssl_tls_update_fd: New fd is 4
[26] DBG:tls_openssl:openssl_tls_write: write was successful (16384 bytes)
[26] ERROR:proto_tls:tls_write_on_socket: buff len:20430, write len:16384, 0x7fa82fc9c450
[26] DBG:proto_tls:proto_tls_send: after write: c=0x7fa82fc9c450 n=0 fd=4`

**Solution**
If the SIP message cannot be fully sent in the first attempt, the `tcp_connection` must be released back to the `TCP main` process so that it can be added to the reactor to listen for the write event again.

---log snippet---
`//First write attempt. which [26] is one process of TCP workers
[26] DBG:proto_tls:proto_tls_send: sending via fd 4...
[26] DBG:tls_openssl:openssl_tls_update_fd: New fd is 4
[26] DBG:tls_openssl:openssl_tls_write: write was successful (16384 bytes)
[26] ERROR:proto_tls:tls_write_on_socket: buff len:20430, write len:16384, 0x7fa82fc9c450
[26] DBG:proto_tls:proto_tls_send: after write: c=0x7fa82fc9c450 n=0 fd=4

//Release tcp connection to tcp main process with state: ASYNC_WRITE_GENW(5).
[26] DBG:core:tcpconn_release:  releasing con 0x7fa82fc9c450, state 5, fd=4, id=105154990
[26] DBG:core:tcpconn_release:  extra_data 0x7fa82fca3248

//TCP main process added the fd to reactor with write. which [31]: the TCP main process
[31] DBG:core:handle_worker: read response= 7fa82fc9c450, 5, fd -1 from 13 (26)
[31] DBG:core:io_watch_add: [TCP_main] io_watch_add op (122 on 4) (0xa37720, 122, 19, 0x7fa82fc9c450,2), fd_no=25/1048576

//TCP main process received write event and dispatcher to [26] worker process
[31] DBG:core:handle_tcpconn_ev: connection 0x7fa82fc9c450 fd 122 is now writable
[31] DBG:core:send2worker: to tcp worker 0 (0/13) load 0, 0x7fa82fc9c450/122 rw 2

//Write remain data
[26] DBG:core:handle_io: We have received conn 0x7fa82fc9c450 with rw 2 on fd 140
[26] DBG:core:handle_io: Received con for async write 0x7fa82fc9c450 ref = 2
[26] ERROR:proto_tls:tls_async_write: Trying to send 4046 bytes from chunk 0x7fa82fcd6a80 in conn 0x7fa82fc9c450 - 159 159
[26] DBG:tls_openssl:openssl_tls_write: write was successful (4046 bytes)
[26] DBG:core:tcp_async_update_write: We have finished writing all our async chunks in 0x7fa82fc9c450
[26] DBG:core:tcpconn_release:  releasing con 0x7fa82fc9c450, state 6, fd=4, id=105154990`

**Compatibility**
This is a change-specific and isolated in proto_tls module - it's 100% retro-compatible with the other versions.

**Closing issues**
N/A

### DIFF
--- a/modules/proto_tls/proto_tls.c
+++ b/modules/proto_tls/proto_tls.c
@@ -618,7 +618,7 @@ send_it:
 	send_sock->last_real_ports->local = c->rcv.dst_port;
 	send_sock->last_real_ports->remote = c->rcv.src_port;
 
-	tcp_conn_release(c, 0);
+	tcp_conn_release(c, c->async && c->async->pending?1:0);
 	return rlen;
 con_release:
 	sh_log(c->hist, TCP_SEND2MAIN, "send 1, (%d)", c->refcnt);


bogdan-iancu:
Hi @36952362 , you have a fair point here. Taking this a step further, maybe the test of ASYNC (to have the fd added for WRITE in reactor) should be mask and automatically done in the `tcp_conn_release()` (instead of having the second param)

modules/proto_tls/proto_tls.c
```diff
@@ -618,7 +618,7 @@ static int proto_tls_send(const struct socket_info* send_sock,
 	send_sock->last_real_ports->local = c->rcv.dst_port;
 	send_sock->last_real_ports->remote = c->rcv.src_port;
 
-	tcp_conn_release(c, 0);
+	tcp_conn_release(c, c->async && c->async->pending?1:0);
```
bogdan-iancu, line 621: OK, let's keep it local, here. Nevertheless, to avoid any clarity or synchronization issue (btw, the access to `->async->pending` should be under lock), I would go for the more simpler approach from the [proto_tcp](https://github.com/OpenSIPS/opensips/blob/master/net/proto_tcp/proto_tcp.c#L588), with the test on the amount of written data. Could you :
* rework the test here
* remove the `tcp_connections_no` fix (to keep it in  #3664)

bogdan-iancu (changes requested):

bogdan-iancu:
@36952362, maybe I failed with my explanations :). Yes, using the `->async` fields requires locking, but this is not the best way to go. I was rather suggesting to use the same kind of testing as in the proto_tcp module (which requires no locking):
```
tcp_conn_release(c, (n<len)?1:0/*pending data in async mode?*/ );
```
by comparing the overall `len` with the actually written data `n` .
As the fix is simple here, let me know if you want to update the PR or I should simply push a fix for that line.

bogdan-iancu:
Thank you @36952362 , PR merged and backported to all 3.4+ versions

PR #3655, opened 2025-05-28
[aaa_diameter] Ignore unknown AVPS in Diameter Replies

LarryLaffer-dev:
**Summary**
Minor change in the "aaa_diameter" module: Simply ignore unknown AVP's in Diameter replies - currently it causes OpenSIPS to crash.

**Details**
Simply go to the "skip" section, if AVP was not found.

**Compatibility**
Non breaking change.


liviuchircu (approved):
Hello, @LarryLaffer-dev, this seems like a much-needed change, in order to smoothen operations and make Diameter interactions work more often, instead of dropping with failure at the very first "unknown" AVP...

Your change actually applies to both Requests and Responses, which is even better.  I also double-checked and there should be an ERROR log for each unknown AVP being now skipped, so the developer has enough info to troubleshoot & fix.

Merging and backporting to all stables.  Great job!

PR #3664, opened 2025-06-04
Resolve the issue of inaccurate tcp_connections_no

36952362:
**Summary**
When OpenSIPS acts as a TCP/TLS client to create connections, the `tcp_connections_no` counter gradually decreases and may even become negative.

**Details**
When other modules, such as `tm`, need to send a message to a target address and no existing TCP/TLS connection is available, the TCP worker process responsible for sending the message will attempt to establish a new TLS connection and create a `tcp_connection` object. It also increments the `tcp_connections_no` counter; however, this increment only affects the `tcp_connections_no` variable local to the TCP worker process (as this variable is process-specific). Then, the worker notifies the TCP main process via a command (either `ASYNC_CONNECT` or `CONN_NEW`).

```
proto_tls_send
  --> tcp_async_connect
  		--> tcp_conn_create
  		      --> tcpconn_new
  		      		--> tcp_connections_no++;
  		      --> tcp_conn_send
```

However, when the TCP main process receives the `ASYNC_CONNECT` or `CONN_NEW` command, it does not increment its own `tcp_connections_no` counter. Later, when the connection is destroyed (in `tcpconn_destroy`) or expires due to its lifetime (in `tcpconn_lifetime`), the counter is decremented.

As a result, the `tcp_connections_no` counter gradually decreases and may even become negative.
create_new_connection

**Solution**
when the TCP main process receives the `ASYNC_CONNECT` or `CONN_NEW` command, it also increment its own `tcp_connections_no` counter

**Compatibility**
N/A

**Closing issues**
N/A

### DIFF
--- a/net/net_tcp.c
+++ b/net/net_tcp.c
@@ -889,7 +889,7 @@ static inline void tcpconn_ref(struct tcp_connection* c)
 
 static struct tcp_connection* tcpconn_new(int sock, const union sockaddr_union* su,
                     const struct socket_info* si, const struct tcp_conn_profile *prof,
-                    int state, int flags)
+                    int state, int flags, int in_main_proc)
 {
 	struct tcp_connection *c;
 	union sockaddr_union local_su;
@@ -956,8 +956,8 @@ static struct tcp_connection* tcpconn_new(int sock, const union sockaddr_union* 
 			goto error;
 		}
 	}
-
-	tcp_connections_no++;
+	if(in_main_proc)
+		tcp_connections_no++;
 	return c;
 
 error:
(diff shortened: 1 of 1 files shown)


36952362:
Thanks for your comments @bogdan-iancu .I had updated the PR according to your suggestion. add one more param(`in_main_proc`) to indicate this new tcp connection will be created in TCP MAIN proc or not.  Please help to double check the logic again. thanks 

Agree with you. According to the comment of `tcp_connections_no`, it means "_current number of open connections_", but it only checks for new in-coming connections, it does not accurately.

bogdan-iancu:
Thank you @36952362 , merge done and backported to all 3.4+ versions

PR #3673, opened 2025-06-19
fix: Disable TLS tickets for 1.2 and 1.3 in order to support reconnects after OpenSIPS restart

vladpaiu:
**Summary**
Fix clients opening new SSL connections ( TLS, WSS, etc ) after an OpenSIPS restart

**Details**
We are not managing TLS 1.2 or 1.3 tickets server side
so in case a client will want to reconnect after an OpenSIPS
restart, it will provide a pre-shared ticket and this re-connection
 will just fail, so take the lazy way out and disable tickets 
https://datatracker.ietf.org/doc/html/rfc8446#section-2.2

**Solution**
Disable tickets support for TLS 1.2 and 1.3

**Compatibility**
Backwards compatible

### DIFF
--- a/modules/tls_wolfssl/wolfssl_conn_ops.c
+++ b/modules/tls_wolfssl/wolfssl_conn_ops.c
@@ -201,6 +201,14 @@ int _wolfssl_tls_conn_init(struct tcp_connection* c,
 		return -1;
 	}
 
+	/* we are no managing TLS 1.2 or 1.3 tickets server side
+	 * so in case a client will want to reconnect after an OpenSIPS
+	 * restart, it will provide a pre-shared ticket and this re-connection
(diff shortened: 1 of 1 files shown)


vladpaiu:
> I am going to merge this as a quick fix, but ideally we should sort out why tickets are not working - did you manage to troubleshoot why they were not working in the first place?

wolfssl was somtimes very 'misterious' as to why the failures was happening : 

```
tls_wolfssl:_wolfssl_tls_accept: TLS accept error: -1, unknown error number
```

other times it was being more explicit : 

```
ERROR:tls_wolfssl:_wolfssl_tls_accept: TLS accept error: -399, Bad session ticket message Size Error
```

which basically lead me down the path of disabling tickets entirely.

PR #3674, opened 2025-06-19
rtp.io: hook the OpenSIPS notification socket automatically

sobomax:
**Summary**

rtp.io: hook the OpenSIPS notification socket automatically.

**Details**

In the current version of the rtp.io, the notification socket still needs to be configured manually in both rtp.io and rtpproxy modules. 

**Solution**

This patch is expected to fix that by allocating extra socketpair and passing that to the librtpproxy and rtpproxy module.

**Compatibility**

Existing configuration should work as is.


sobomax:
This has been merged to the master already. Thanks @razvancrainea

PR #3675, opened 2025-06-24
core: split DEP_REVERSE_INIT into _MINIT and _CINIT parts

sobomax:
**Summary**

Add `DEP_REVERSE_CINIT` and `DEP_REVERSE_MINIT` dependency flags allowing to reverse `mod_init()` and `child_init()` call order for module A and module B independently.

Make DEP_REVERSE_INIT be (DEP_REVERSE_CINIT|DEP_REVERSE_MINIT).

**Details**

This is needed when for whatever architectural reason certain pair of modules (i.e. `rtp.io` and `rtpproxy`) want to use different order for `mod_init()` and `child_init()`. In case of the `rtp.io` vs `rtpproxy`, we want `mod_init()` for the `rtp.io` first so that it can allocate sockets and initialize `rtpproxy` client config. Yet, we want `child_init()` for the `rtpproxy` module to be called first, so that we can set flag when we are in the notification process.

**Solution**

`DEP_REVERSE_INIT` became `DEP_REVERSE_MINIT`, added `DEP_REVERSE_CINIT` and define compound `DEP_REVERSE_INIT` be `(DEP_REVERSE_CINIT|DEP_REVERSE_MINIT)`.

**Compatibility**

Should not be any.

### DIFF
--- a/sr_module.c
+++ b/sr_module.c
@@ -675,7 +675,7 @@ static int init_mod_child( struct sr_module* m, int rank, char *type,
 	if (!skip_others && init_mod_child(m->next, rank, type, 0) != 0)
 		return -1;
 
-	for (dep = m->sr_deps_init; dep; dep = dep->next)
+	for (dep = m->sr_deps_cinit; dep; dep = dep->next)
 		if (!dep->mod->init_child_done)
 			if (init_mod_child(dep->mod, rank, type, 1) != 0)
 				return -1;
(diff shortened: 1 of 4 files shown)


liviuchircu (approved):
Very cool refactoring & simplification @sobomax!  The new "reverse child init" addition looks safe and backwards-compatible too.  Merging and backporting to 3.6 branch.

PR #3687, opened 2025-07-18
fix: when deciding if we need to propagate a reply, skip branches scheduled for cancellation but with no replies yet

vladpaiu:
**Summary**
Current bug is : 

- we have branch1 sent to a destination which is dead/down/not replying
- we have branch2 replying with a code that the script decides should terminate the transaction ( ie. if ($rs == 486) t_cancel_branch("o")

The bug is that the 486 does not get propagated on the spot - it gets saved internally and relayed only when fr_timeout elapses.

**Details**
Ignore branches scheduled for CANCELation via t_cancel_branch, which don't have any replies on them yet.

**Solution**
Ignore branches scheduled for CANCELation via t_cancel_branch, which don't have any replies on them yet.

**Compatibility**
Should be backwards compatible - I hope.

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -833,6 +833,11 @@ static inline int t_pick_branch( struct cell *t, int *res_code, int *do_cancel)
 			continue;
 		/* skip 'empty branches' */
 		if (!t->uac[b].request.buffer.s) continue;
+
+		/* skip branches with no replies yet, but which were marked to
+		 * be cancelled via t_cancel_branch */
+		if (t->uac[b].flags & T_UAC_TO_CANCEL_FLAG) continue;
+
 		/* there is still an unfinished UAC transaction; wait now! */
 		if ( t->uac[b].last_received<200 ) {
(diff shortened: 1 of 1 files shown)


bogdan-iancu (commented):
@vladpaiu , this is not a bug at all. The RFC3261 says you have to forward back to caller a negative final reply only after completing all the branches of the transaction. Now, a branch which haven't received any reply, even if marked for canceling, it is still a open, ongoing branch, you cannot simply ignore it. It is possible for that branch to generate a 200 OK at a point (as a response to your INVITE), so you cannot cancel it anymore.
Also, for the sake of the sanity of the TM internal logic, branches should be properly terminated upon sending a reply back to caller. I mean that unresponsive branch should be marked as completed, like with a 408 timeout at least. So the transactional data is consistent.
In your case, why not setting a low (1-2 secs) fr_timer to avoid waiting too long for such unresponsive branches? 

Anyhow, I do not see this as a bug at all. In the best way maybe as a feature request in changing the TM behavior.

bogdan-iancu:
ping @vladpaiu

PR #3688, opened 2025-07-18
run reply async resume callback under reply lock, as it might make changes to AVPs

vladpaiu:
**Summary**
Currently, the reply async resume callback is run without holding the replies lock, and if that callback tries to populate AVPs as part of the resume script function, it might lead to concurrent access to the transaction AVP list, if another reply ends up being processed in another process.

**Details**
Run the async resumt callback under reply lock

**Solution**
Run the async resumt callback under reply lock

**Compatibility**
Backwards compatible


bogdan-iancu (approved):
All looks good, thank you

PR #3693, opened 2025-07-25
Fix various function or variable type mismatch

lemenkov:
**Summary**
Fix a few definition/declaration type mismatches

**Details**
This PR fixes various GCC warnings about type mismatch.

**Solution**
Ensured that function/variable declaration and their definition are consistent.

**Compatibility**
This improves compatibility with various ANSI C compilers I believe.

**Closing issues**
n/a.


bogdan-iancu:
Great catches @lemenkov !

PR #3731, opened 2025-10-02
sql_cacher: add support for reads of sql_cached_value while table is being reloaded

davidtrihy-genesys:
**Summary**
When calling `$sql_cached_value` during a reload initiated by MI it locks the whole table and that read tries to acquire a lock and can only acquire it once the write has finished, this can block SIP processing unnecessarily as the cachedb implementations typically lock on their own values, eg. cachedb_local locks on each hash which corresponds to a row, this PR adds support to read during a full reload and only lock the row being read, if a reload is in progress you can read the previous version value during the reload if it is returned

**Details and Solution**
Uses stdatomic to set a boolean flag once the write lock is acquired, this can then be tested during reads and be more permissive with values returned as long as the version is only current version - 1 otherwise it returns null

**Compatibility**
Changes the default behaviour to now be more permissive but adds a new param to `cache_table` called `full_caching_lock_scope` which can be set to `row` or `table`, defaults to `row` and setting it to `table` provides support for the current way it works

### DIFF
--- a/modules/sql_cacher/sql_cacher.c
+++ b/modules/sql_cacher/sql_cacher.c
@@ -198,7 +198,7 @@ struct module_exports exports = {
 
 static int parse_cache_entry(unsigned int type, void *val)
 {
-	cache_entry_t *new_entry;
+	cache_entry_t *new_entry = NULL;
 	char *p1, *p2, *tmp, *c_tmp1, *c_tmp2;
 	int col_idx;
 	int rc = -1;


modules/sql_cacher/sql_cacher.c
```diff
@@ -157,14 +199,15 @@ struct module_exports exports = {
 static int parse_cache_entry(unsigned int type, void *val)
 {
 	cache_entry_t *new_entry;
+	cache_lock_t *new_lock;
```
davidtrihy-genesys, line 202: Need to delete this

davidtrihy-genesys:
@liviuchircu rather than pushing commits to fix any compiler warnings on all the platforms, I'll let you review first and then I can address any additional ones that pop up

atomic.h
```diff
@@ -72,6 +73,7 @@ typedef _Atomic(unsigned long) atomic_t;
  * not some alias that contains the same information.
  */
 typedef struct { volatile unsigned int counter; } atomic_t;
+typedef atomic_t atomic_bool_t;
```
davidtrihy-genesys, line 76: Just tested this typedef for using the assembly atomic operations defined for x86_64 and it works fine too, so we have support if someone is still using an unreasonably old version of GCC to compile this

davidtrihy-genesys:
The multi_arch build failed for an unrelated reason to my changes but I can't kick it off again so I consider this PR fully ready for review

PR #3757, opened 2025-11-14
Reduce verbosity of libcurl and send errors to stderr

vladpaiu:
**Summary**
Reduce rest_client logging verbosity and send errors to stderr

**Details**
When running OpenSIPS in a container and logging to stdout, libcurl is way too verbose and produces alot of noise that pollutes the logs

**Solution**
Reduce verbosity and send errors to stderr

**Compatibility**
Backwards compatible


liviuchircu:
Hi @vladpaiu!  This makes sense, most likely it looks like a left-over development setting, but since OpenSIPS closes the STDOUT file descriptor by default (unless you enable `log_stdout = true`), the libcurl verbose logging will go unnoticed.

Backporting down to `3.4`.

PR #3761, opened 2025-11-20
Fix memory leaks in stir_shaken module

pfournier:
**Summary**

The stir-shaken module would leak memory any time a signature was added to a reply instead of the request or returned in a variable. This commit makes sure the Identity header lump does not get duplicated and freed only once. Also fixed possible memory leaks in case of issues while adding a date header to a request without one.

**Details**

When implementing a STIR/SHAKEN signature server using OpenSIPS, whenever OpenSIPS was returning the SHAKEN signature in a `302` response, memory would be leaked causing a lack of `PKG` memory after some time. Increasing the `P_MEMORY` would only extend the time before opensips processes would run out of memory and be unable to process requests.

**Solution**

This patches adds the `LUMP_RPL_NODUP` flag to the `add_lump_rpl()` call to prevent that function from duplicating the header. When the lumps are freed, the header created in the `build_identity_hf()` function is now freed too.

**Compatibility**

Does not break anything.

**Closing issues**

### DIFF
--- a/modules/stir_shaken/stir_shaken.c
+++ b/modules/stir_shaken/stir_shaken.c
@@ -503,11 +503,13 @@ static int add_date_hf(struct sip_msg *msg, time_t *date_ts)
 	anchor = anchor_lump(msg, msg->unparsed - msg->buf, 0);
 	if (!anchor) {
 		LM_ERR("Failed to get anchor lump\n");
+		pkg_free(buf);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


liviuchircu (approved):
Nice findings, @pfournier!  About the `add_lump_rpl()` issue, it looks like the bug comes from the difference in API behavior between `insert_new_lump_before()` and `add_lump_rpl()` APIs.  By default, the first one will not cause a leak, while the second one causes it, unless you _explicitly_ provide `LUMP_RPL_NODUP`.  Probably the original author missed this detail, as they supplied the same string buffer to both functions.

Will merge down to 3.4.

PR #3769, opened 2025-12-01
nathelper: port add_contact_alias() and handle_ruri_alias()

sobomax:
**Summary**
Port `add_contact_alias()` and `handle_ruri_alias()` functions

**Details**
Those are slightly better version of the `fix_nated_contact()` originating from Kamailio.

The problem with `fix_nated_contact()` is that it could cause RURI for in-dialog requests contain a different IP address as compared to the Contact in the initial INVITE.

 Most endpoints are fine with that, however tere are some, notably MS Teams, which would reject such request with:

```
SIP/2.0 500 Internal Server Error
Reason: Q.850;cause=47;text="76a7afc3-d033-4988-9000-f70b41e4db1b;RURI in request has invalid FQDN value"
```

The full exchange when using `fix_nated_contact()` is as follows:

UA->OpenSIPS (initial INVITE):
```
opensips[NNNN]: RECEIVED message from 10.2.80.32:5060:
INVITE sip:[...]
Contact: <sip:[...]@10.2.80.22:5060>
                    ^^^^^^^^^^
```
OpenSIPS->UA:
```
SIP/2.0 100 Giving it a try
```
```
SIP/2.0 200 OK
```
UA->OpenSIPS:
```
ACK sip:[...]
```
OpenSIPS->UA (re-INVITE):

```
opensips[NNNN]: SENDING message to 10.2.80.32:5060:
INVITE sip:[...]@10.2.80.32:5060 SIP/2.0
                 ^^^^^^^^^^
```
UA->OpenSIPS:
```
SIP/2.0 500 Internal Server Error
```

**Solution**
Use `add_contact_alias()` in the INVITE/200 OK paths and `handle_ruri_alias()` in in-dialog request path.

UA->OpenSIPS (initial INVITE):
```
opensips[NNNN]: RECEIVED message from 10.2.80.32:5060:
INVITE sip:[...]
Contact: <sip:[...]@10.2.80.22:5060>
                    ^^^^^^^^^^
```
OpenSIPS->UA:
```
SIP/2.0 100 Giving it a try
```
OpenSIPS->B2BUA (initial INVITE):
```
opensips[NNNN]: SENDING message to [...]:
INVITE sip:[...]
Contact: <sip:[...]@10.2.80.22:5060;alias=10.2.80.32~5060~1>
                    ^^^^^^^^^^
```
B2BUA->OpenSIPS (re-INVITE):
```
opensips[NNNN]: RECEIVED message from [...]:
INVITE sip:[...]@10.2.80.22:5060;alias=10.2.80.32~5060~1 SIP/2.0
                 ^^^^^^^^^^
```
OpenSIPS->UA (re-INVITE):
```
opensips[NNNN]: SENDING message to 10.2.80.32:5060:
                                   ^^^^^^^^^^
INVITE sip:[...]@10.2.80.22:5060 SIP/2.0
                 ^^^^^^^^^^
```

**Compatibility**
Should not be any if used correctly. Special test case for the voiptests has been created and will be enabled once the change is merged into the opensips/master.

### DIFF
--- a/modules/nathelper/README
+++ b/modules/nathelper/README
@@ -43,6 +43,8 @@ nathelper Module
               1.5.3. add_rcv_param([flag]),
               1.5.4. fix_nated_register()
               1.5.5. nat_uac_test(flags)
+              1.5.6. add_contact_alias([ip_addr, port, proto])
+              1.5.7. handle_ruri_alias()
 
         1.6. Exported MI Functions
 
@@ -91,7 +93,9 @@ nathelper Module
    1.22. add_rcv_paramer usage
    1.23. fix_nated_register usage
    1.24. nat_uac_test usage
-   1.25. nh_enable_ping usage
+   1.25. add_contact_alias usage
+   1.26. handle_ruri_alias usage
+   1.27. nh_enable_ping usage
 
 Chapter 1. Admin Guide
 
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Hi @sobomax , thanks for the contribution here. Here are some first thoughts:
* yes, indeed, the problem addressed by this PR is real
* there are already ways to address the problem described here:
** use the topo hiding with dialog module - it is doing more or less the same as you manually do , but is a simpler and more elegant way
** you can achieve 100% the same results with scripting, by adding the extra Contact param via fix_nated_contact() (see  uri_params func param, where you can store as base64 encoded the full original URI); on sequential requests, it's a simple transformation to fetch (from URI param) and replace the RURI.
** even simpler, if using the dialog module (without TH), you can store the received Contact URI into a dlg val and restore it later, it something trivial.
* storing data into the SIP req is not always the best approach; beside leaking network information, you rely on the end points to correctly do their jobs (like mirroring that param) - history taught us (see the RR/Route mess) that reality is a bit different.

The benefit of these new functions is (as far as I can understand) to simplify a bit the scripting IF you do NOT WANT to use dialog support. Otherwise, it may complicate even more the scripting, as the user has to use more functions. Maybe we can add a new flag to fix_nated_contact(), in order to add this new param ? And we can use the "maddr" param (standard) which is automatically recognized by OpenSIPS (in routing RURI) for overriding the actual RURI domain/port

bogdan-iancu:
I might have said something wrong in the above comment - the `maddr` param is of no use here, as it works exactly the opposite - it acts as a outbound proxy, without changing the host part; while we need to actually change the host part of the RURI. So please ignore that part.
On the other hand, as OpenSIPS is a proxy and we are in the NAT traversal area, it is mandatory for OpenSIPS to stay in the path of the dialog for all the sequential requests - this translates into doing dialog+TH (but you do not want to use this), or to do record_route. So, the only applicable scenario is doing RR + LR -> why nor incorporate this "fixing" into the loose_route part ?

bogdan-iancu:
one more ping for @sobomax  :)

sobomax:
Works fine, I've added scenario into voiptests here https://github.com/sippy/voiptests/commit/8c38690d14e8d17f44d980af54ec34b5bea0f0f6

PR #3791, opened 2025-12-26
Fix format specifier warnings on 32-bit architectures (i686)

lemenkov:
**Summary**
Fix format specifier warnings on 32-bit architectures (i686) with GCC 15

**Details**
This PR fixes compilation warnings that appear on 32-bit architectures (i686) when using GCC 15. The warnings are caused by using incorrect format specifiers for `size_t`, `int64_t`, and `uint64_t` types, which have different sizes on 32-bit vs 64-bit platforms.

On x86_64, `size_t` and `int64_t` are aliases for `unsigned long` and `long int` respectively, so `%ld` and `%lu` work fine. However, on i686 (32-bit):
- `size_t` is `unsigned int` (4 bytes), not `unsigned long`
- `int64_t` is `long long int` (8 bytes), not `long int`
- `uint64_t` is `unsigned long long int` (8 bytes), not `unsigned long int`

This causes GCC to emit `-Wformat` warnings on 32-bit builds, as the format specifiers don't match the actual argument types.

Affected modules:
- **http2d**: Incorrect `%ld` for `size_t` in server.c
- **aaa_diameter**: Incorrect `%ld`/`%lu` for `int64_t`/`uint64_t` in dm_impl.c

**Solution**
Replace architecture-specific format specifiers with portable C99 standard specifiers:

1. Use `%zu` for `size_t` (defined in C99, works on all architectures)
2. Use `%" PRId64` for `int64_t` (from `<inttypes.h>`, expands to correct specifier per platform)
3. Use `%" PRIu64` for `uint64_t` (from `<inttypes.h>`, expands to correct specifier per platform)

These portable specifiers automatically adapt to the platform:
- On 64-bit: `%zu` → format as `unsigned long`, `PRId64` → `"ld"`, `PRIu64` → `"lu"`
- On 32-bit: `%zu` → format as `unsigned int`, `PRId64` → `"lld"`, `PRIu64` → `"llu"`

**Compatibility**
No compatibility impact. This is purely a build-time fix that doesn't change runtime behavior or APIs. The same values are printed, just with correct format specifiers.

**Closing issues**
N/A

### DIFF
--- a/modules/aaa_diameter/dm_impl.c
+++ b/modules/aaa_diameter/dm_impl.c
@@ -20,6 +20,7 @@
 
 #include <freeDiameter/extension.h>
 #include <sys/eventfd.h>
+#include <inttypes.h>
 
 #include "../../ut.h"
 #include "../../lib/list.h"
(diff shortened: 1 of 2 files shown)


lemenkov:
Ping, @razvancrainea :)

PR #3794, opened 2026-01-02
rtpproxy: add rtpp_bind_local support for the offer/answer/engage.

sobomax:
**Summary**

This PR adds more flexible control over socket that RTPProxy picks for a new or existing media session. The existing mechanism with I/E designators is a bit crude and limiting.

**Details**

When `rtpp_bind_local` AVP is set, its value would be provided to the rtpproxy with "l" modifier, allowing proper address to be selected for the session.

**Solution**

This can be used either globally by setting the rtpproxy module parameter:

```
modparam("rtpproxy", "rtpp_bind_local_avp", "$socket_out(ip)")
```

Or on logic level by setting up the AVP:

```
route[] {
...
   $avp(rtpp_bind_local) = '1.2.3.4';
   rpproxy_engage();
...
```

**Compatibility**

Should be none, it's a new API.

**Closing issues**

None

### DIFF
--- a/modules/rtpproxy/README
+++ b/modules/rtpproxy/README
@@ -30,6 +30,7 @@ rtpproxy Module
               1.5.13. generated_sdp_port_min (integer)
               1.5.14. generated_sdp_port_max (integer)
               1.5.15. generated_sdp_media_ip (string)
+              1.5.16. rtpp_bind_local_avp (string)
 
         1.6. Exported Functions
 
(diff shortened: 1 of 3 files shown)


razvancrainea:
@sobomax I love the idea of passing the local socket to rtpproxy, but we're trying to avoid using AVPs for passing information like that, as it might confuse people (happened too many times). That's why we'd rather explicitely using the argument as a flag to rtpproxy_offer/answer/engage(). That's actually why we like to pass all the other flags explicitely (see [`n[<SOCKET>]`](https://opensips.org/docs/modules/3.6.x/rtpproxy.html) param), as it is more clear and avoids erronously inherritting the AVP on replies. That's also why we dropped setting the `set_id` through an AVP as well, even though the AVP inherittance would be a plus in that scenario. Can you please refactor this to pass the bind socket as a parameter, something like`l<SOCKET>` (I know we already use `l` as a forced lookup, but this time is followed by `<`, so we can clearly differentiate; or simply use a different letter)

Furthermore, not sure if the protocol supports, but it would be nice to have some sort of interface alias/tag - that's useful when using multiple rtpproxy instances in a set, where you have different IPs configuration, and simply choose internal or external interface. Your example shows an IP, so I presumed that currently only IPs are allowed.

razvancrainea:
@sobomax I've moved the interface specification to the rtpproxy_offer/answer functions. TBH, I would not worry too much about rtpproxy_engage, as that logic became more or less deprecated in favor of the new rtp_relay module, which does all that part quite well. The only missing thing is to hook the new labels in the rtpproxy command when they are provisioned over rtp_relay [flags](https://opensips.org/docs/modules/4.0.x/rtp_relay.html#pv_rtp_relay), (i.e. `$rtp_relay(iface)` and `$rtp_relay(oface)`). Will keep this open until we sort this out.

PR #3800, opened 2026-01-16
b2b_logic bridge max_duration flag fix

ErhanOnur:
**Summary**
 The fix for max_duration flag of the b2b_bridge function in the b2b_logic module

**Details**
The max_duration flag of the b2b_bridge function in the b2b_logic module was not working as expected because the default max_duration setting was overriding the flag value. This issue has been resolved by introducing a new module parameter.

**Solution**
When b2b_bridge is triggered , the lifetime of the call was set as 5, because max_duration flag is 5:
Oct 2 07:10:35 [19] ERROR:b2b_logic:b2b_script_bridge: params->lifetime: 5 (max_duration)

When the 200 OK reply of initial INTIVE is arrived, the lifetime is updated again with default lifetime value:
Oct 2 07:10:35 [19] ERROR:b2b_logic:b2b_add_dlginfo: max_duration : 43200 tuple->lifetime 43201

The second update is prevented with the new module parameter.

**Compatibility**
If the new module parameter is set to false, the code behaves as before. The default value is false.

**Closing issues**
N/A

### DIFF
--- a/modules/b2b_logic/b2b_logic.c
+++ b/modules/b2b_logic/b2b_logic.c
@@ -199,6 +199,7 @@ struct b2b_ctx_val *local_ctx_vals;
 
 unsigned int ent_term_interval;
 struct b2b_term_timer *ent_term_timer;
+int use_lifetime_of_bridge = 0;
 
 static const cmd_export_t cmds[]=
 {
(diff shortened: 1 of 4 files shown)


razvancrainea:
Thank you for this contribution! Merging both upstream soon.

PR #3804, opened 2026-01-19
Sipmsg validate all contacts

davidtrihy-genesys:
**Summary**
Extends `sipmsg_validate` with `c` flag set for validating `Contact` headers to multiple different requests

**Details**
In response to my testing of the fix outlined in this PR https://github.com/OpenSIPS/opensips/pull/3803 I found limitations of the `sipmsg_validate` function which only applied `Contact` validation to `INVITE`.

This extends validation to `REGISTER|UPDATE|SUBSCRIBE|REFER` - not taking into account what type of request it is eg. initial or in dialog, so with a `REFER` a dialog can be created from a `REFER` request according to https://www.rfc-editor.org/rfc/rfc3515#section-2 but an in dialog `REFER` does not honour the `Contact` header sent but this validation logic does not check that and it will fail, it is designed to be up to the script writer to do the in dialog check themselves.

**Solution**
New static function created to specifically check the `Contact` header which will take into account multiple different aspects of the SIP message.

Failure conditions

1. No Contact header
2. Contact header contains * as the body and one of these other conditions
    i.  Is a SIP reply
   ii. is not a REGISTER request
  iii. is a REGISTER request but the `Expires` header is missing or it is set to a value greater than 0

New return code of -30 for when a * is the value but is invalid for the three types of conditions listed in 2.

**Compatibility**
Depending on how you have written your script and if the behaviour of your OpenSIPs deployment is relying implicitly on the Contact header not being checked for the new requests it could break scripts.

### DIFF
--- a/modules/sipmsgops/sipmsgops.c
+++ b/modules/sipmsgops/sipmsgops.c
@@ -1589,13 +1589,11 @@ static enum sip_validation_failures validate_contact_header(struct sip_msg *msg,
 	contact_t * contacts = NULL;
 	exp_body_t *expires_body = NULL;
 	struct sip_uri test_contacts;
-	enum sip_validation_failures ret = SV_NO_ERROR;
 
 	if (method & METHOD_WITH_CONTACT_HDR) {
 		if (!msg->contact) {
 			strcpy(reason, "SIP message doesn't have 'Contact' header");
-			ret = SV_NO_CONTACT;
-			goto failed;
+			return SV_NO_CONTACT;
 		}
 
 		/* iterate through Contact headers */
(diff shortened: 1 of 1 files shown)


modules/sipmsgops/sipmsgops.c
```diff
@@ -1576,18 +1577,101 @@ enum sip_validation_failures {
...
+						strcpy(reason, "Contact header for SIP message contains '*' only valid for REGISTER request");
+						ret = SV_BAD_STAR_CONTACT;
+					}
+					goto failed;
+				}
+			}
+		}
+	}
+
+	ret = SV_NO_ERROR;
```
bogdan-iancu: see the first comment. Maybe this set should on top, to cover the methods without contact (or at least on the `else` branch of the test. 
Also it should the default for the non-REGISTER replies.

Issue #3809, opened 2026-02-02
[FEATURE] Select "type" of gateway during route_to_carrier() in drouting.


osgjps:
**Is your feature request related to a problem? Please describe.**

If you have different types of gateways such as inbound, outbound, or inbound/outbound defined on a single carrier, there is no way of telling route_to_carrier() to not select inappropriate types of gateways (i.e. sending an outbound call to an inbound only gateway).

**Describe the solution you'd like**

To be able to pass a type selector to route_to_carrier() that allows inclusion of only the specified types of gateways.

**Implementation**
 
  - Component: drouting
  - Type: additional optional parameter on route_to_carrier() 

**Describe alternatives you've considered**

No other viable alternative exists. This course of action was recommended by Vlad.

**Additional context**

vladpaiu:
think what is the point here is that he is trying to make route_to_carrier be 'symmetric' in functionality with is_from_gw ( which already is able to filter based on gateway type ), so the idea is 'filter in based on gw type' vs 'being able to filter out based on gw type'

PR #3810, opened 2026-02-02
Add support to select 'type' of gateway during route_to_carrier

osgjps:
**Summary**

This adds the optional "type" parameter to route_to_carrier()

**Details**

A solution to https://github.com/OpenSIPS/opensips/issues/3809

**Solution**

This adds the optional "type" parameter to route_to_carrier()

**Compatibility**

Should not break backwards compatibility.  The new parameter was added to the end of the parameter list so existing dialplans should continue to work without issues.  If the option is omitted, all gateways are considered.

**Closing issues**

### DIFF
--- a/modules/drouting/drouting.c
+++ b/modules/drouting/drouting.c
@@ -267,10 +267,10 @@ static int w_do_routing(struct sip_msg* msg, int *grp, long flags, str *wl,
 		pv_spec_t* rule_att, pv_spec_t* gw_att, pv_spec_t* carr_att,
 		struct head_db *part);
 static int do_routing(struct sip_msg* msg, struct head_db *part, int grp,
-		int flags, str* wl);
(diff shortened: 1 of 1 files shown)


modules/drouting/drouting.c
```diff
@@ -267,10 +267,10 @@ static int w_do_routing(struct sip_msg* msg, int *grp, long flags, str *wl,
 		pv_spec_t* rule_att, pv_spec_t* gw_att, pv_spec_t* carr_att,
 		struct head_db *part);
 static int do_routing(struct sip_msg* msg, struct head_db *part, int grp,
-		int flags, str* wl);
```
bogdan-iancu, line 270: We need this "type" GW selection also added to the do_routing() script function

bogdan-iancu (changes requested):

PR #3820, opened 2026-02-11
sockets_mgm: fix race in receive_fd causing infinite loop on reload

### LINKED ISSUE #3789: [BUG] sockets_mgm module causes high CPU usage and "freezes" OpenSIPS
```
version: opensips 3.6.3 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, F_PARALLEL_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
git revision: d5222226a
main.c compiled on  with gcc 12
```

**Describe the bug**
Hi, the new sockets_mgm module seems to be causing high CPU usage and affecting the uptime of the OpenSIPS static UDP sockets as well as the dynamic sockets. Sometimes, OpenSIPS stops responding after I do a sockets_reload. Other times the issue occurs right at OpenSIPS start up. For the past few days I tried to manage it by restarting OpenSIPS whenever it freezed, and by freezing I mean it would stop responding even at the static 5060/5061 UDP/TLS sockets. Originally, I was doing mi(sockets_reload) in a timer route every 2 minutes, and I thought maybe the timer route was misbehaving, so I disabled the timer route and switched over to MI HTTP and did a sockets_reload once and the issue still occurred. Also, I noticed that when OpenSIPS freezes, both the opensips-cli through FIFO and MI HTTP stop working, I don't get any response at all.

I played around with the udp_workers and memory limits, but had no luck. I checked the GitHub issues and the mailing list, but didn't find anything related to this issue which makes sense as the sockets_mgm is a newer module so it's usage might not be as widespread.

By the way, thanks for the amazing piece of software and for the open source contributions!

**To Reproduce**
1. Add two UDP sockets to the sockets table in the database.
2. Start OpenSIPS, with sockets_mgm module.
3. Either wait a little while, or issue a sockets_reload command.
4. CPU usage will spike up and OpenSIPS will stop responding (htop screenshot attached).

**Expected behavior**
Allow the creation of new dynamic sockets and cleanup of old ones on sockets_reload/OpenSIPS startup without any abnormal CPU usage.

**Relevant System Logs**
On the debug level, I don't see anything specific logged related to this but I can grep you any specific pattern from the logs you need. Please let me know.

 - Operating System: Debian GNU/Linux 12 (bookworm)
 - OpenSIPS installation: OpenSIPS version 3.6.3 from official APT repository
 - other relevant information: Intel(R) Xeon(R) CPU E5-2696 v4 @ 2.20GHz, 8 CPU Cores, 8 GB RAM, VMware hypervisor

**Additional context**
Another thing that I noticed is that sometimes, OpenSIPS will fail to route back a reply (like a 200-OK) from the dynamic socket. Restarting fixes it, but it can still happen while running. OpenSIPS will log the following (Please note that the 10.10.43.13 IP address is the actual destination where the 200-OK reply is expected to be routed, not the dynamic socket. There is a Cisco CUCM SIP server running at 10.10.43.13:5060 and it always accepts traffic and responds normally):

```
2025-12-20T13:21:21.055677+04:00 freeswitch /usr/sbin/opensips[108247]: ERROR:core:proto_udp_send: sendto(sock,0x7fa18a419488,467,0,0x7fa186d72550,16): Socket operation on non-socket(88) [10.10.43.13:5060]
2025-12-20T13:21:21.055931+04:00 freeswitch /usr/sbin/opensips[108247]: ERROR:tm:msg_send: send() to 10.10.43.13:5060 for proto udp/1 failed
2025-12-20T13:21:21.613415+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:core:proto_udp_send: sendto(sock,0x7fa185b0f280,467,0,0x7fa186d72550,16): Socket operation on non-socket(88) [10.10.43.13:5060]
2025-12-20T13:21:21.613790+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:tm:msg_send: send() to 10.10.43.13:5060 for proto udp/1 failed
2025-12-20T13:21:22.619632+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:core:proto_udp_send: sendto(sock,0x7fa185b0f280,467,0,0x7fa186d72550,16): Socket operation on non-socket(88) [10.10.43.13:5060]
2025-12-20T13:21:22.620158+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:tm:msg_send: send() to 10.10.43.13:5060 for proto udp/1 failed
2025-12-20T13:21:24.601315+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:core:proto_udp_send: sendto(sock,0x7fa185b0f280,467,0,0x7fa186d72550,16): Socket operation on non-socket(88) [10.10.43.13:5060]
2025-12-20T13:21:24.601757+04:00 freeswitch /usr/sbin/opensips[108259]: ERROR:tm:msg_send: send() to 10.10.43.13:5060 for proto udp/1 failed
```

Image

NormB:
## Summary

Fixes #3789 — `sockets_mgm` reload causes all OpenSIPS worker processes to spin at 100% CPU in an infinite loop inside `push_sock2list()` → `sock_listadd()`.

### Root Cause

During `sockets_reload`, every process receives an IPC RPC to run `rpc_socket_reload_proc()`. Worker (non-dynamic) processes close their copy of each dynamic socket, then call `receive_fd()` on the shared `sock_mgm_unix[0]` socketpair to obtain a fresh fd from the mgm process.

Since `sock_mgm_unix[0]` is a single `SOCK_STREAM` socket shared across **all** worker processes, concurrent `receive_fd()` calls race: worker A can consume the fd+metadata response intended for worker B. When this happens, worker B receives the wrong `struct sock_mgm *` pointer — one that references a socket already present in worker B's listener list. The `sock_listadd()` macro then corrupts the linked list into a circular loop (`si->next == si`), causing `push_sock2list()` to spin indefinitely.

GDB confirmed: `si->next` pointed to itself, and `protos[1].listeners` formed an infinite cycle through the same node.

### Fix

Add a `sock_mgm_reload_lock` that serializes the entire send-IPC-to-mgm + receive-fd sequence across worker processes, ensuring only one worker at a time performs the fd-passing handshake on the shared socketpair.

Dynamic (mgm) processes are **excluded** from this lock because they create sockets directly via `sock_mgm_add_listener()` and never call `receive_fd()`. Including them would deadlock: the worker holding the lock blocks on `receive_fd()` waiting for the mgm process to handle `rpc_sockets_send()`, but the mgm process would be spinning on the same lock.

### Changes

- **`sockets_mgm.c`**: Add `sock_mgm_reload_lock` (shared memory lock), allocated and initialized in `mod_init()`. In `rpc_socket_reload_proc()`, non-dynamic processes acquire the lock before the reload sequence and release it after all `receive_fd()` calls complete.

## Test plan

- [x] Start OpenSIPS with 2 dynamic UDP sockets from DB — 22 processes, 16 dynamic socket entries, all sleeping
- [x] 1st `sockets_reload` (add 3rd socket) — MI returns OK, 24 socket entries, no CPU spinning
- [x] 2nd `sockets_reload` (add 4th socket) — MI returns OK, 32 socket entries, no CPU spinning
- [x] All processes remain in S (sleeping) state throughout — verified with `ps` and `top`
- [x] Clean shutdown after testing

### DIFF
--- a/modules/sockets_mgm/sockets_mgm.c
+++ b/modules/sockets_mgm/sockets_mgm.c
@@ -60,6 +60,7 @@ static db_func_t sock_mgm_db_func;
 static unsigned long *sock_mgm_version;
 static unsigned int sock_mgm_max_sockets = SOCKETS_MGM_DEFAULT_MAX_SOCKS;
 static gen_lock_t *sock_mgm_lock;
+static gen_lock_t *sock_mgm_reload_lock;
 static int *sock_mgm_proc_no;
 static int sock_mgm_unix[2];
 extern int is_tcp_main;
@@ -217,6 +218,11 @@ static int mod_init(void)
 	sock_mgm_lock = lock_alloc();
 	if (!sock_mgm_version || !lock_init(sock_mgm_lock)) {
 		LM_ERR("initializing sock_mgm_version lock\n");
+		return -1;
+	}
+	sock_mgm_reload_lock = lock_alloc();
+	if (!sock_mgm_reload_lock || !lock_init(sock_mgm_reload_lock)) {
+		LM_ERR("initializing sock_mgm_reload lock\n");
 		return -1;
 	}
 
@@ -1145,6 +1151,21 @@ static void rpc_socket_reload_proc(int sender_id, void *_ver)
 	int sockets_update_count = 0, fd;
 
 	LM_NOTICE("Reloading process for version %lu\n", version);
+
+	/* Serialize the entire send-IPC-to-mgm and receive-fd sequence across
+	 * all non-dynamic (worker) processes. Without this, multiple workers
+	 * calling receive_fd() concurrently on the shared sock_mgm_unix
+	 * socketpair can steal each other's fd responses, causing the same
+	 * socket to be added to a worker's listener list twice and corrupting
+	 * it into a circular linked list (infinite loop in sock_listadd).
+	 *
+	 * Dynamic (mgm) processes don't use receive_fd - they create sockets
+	 * directly - so they must NOT acquire this lock, otherwise they would
+	 * deadlock with the worker that holds the lock while blocked on
+	 * receive_fd waiting for the mgm to process rpc_sockets_send. */
+	if (!sock_mgm_dynamic_proc)
+		lock_get(sock_mgm_reload_lock);
+
 	lock_get(sock_mgm_lock);
 	if (*sock_mgm_version > version) {
 		LM_WARN("new version %lu available (current=%lu)\n", *sock_mgm_version, version);
@@ -1171,6 +1192,9 @@ release:
 			sock_mgm_update_fd(sock, fd);
 		}
 	}
+
+	if (!sock_mgm_dynamic_proc)
+		lock_release(sock_mgm_reload_lock);
 }
 
 static int sockets_pool_init(void)


razvancrainea:
Hi, Norm! Did you manage to get this working, or the race is no longer valid?

NormB:
Hey Răzvan — thanks for the nudge, and sorry for the slow turnaround.

Short answer: the race is still valid on current master, and the lock in this PR does fix it. I went back and reproduced it from scratch on today's master tip to make sure I wasn't chasing a ghost.

PR #3825, opened 2026-02-13
#3798 Don't require certificate for TLS clients

### LINKED ISSUE #3798: [FEATURE] TLS client_domain without certificate

We currently use the same certificate for Client Auth and Server Auth (commonly done) however given impending changes to TLS certificates most publicly trusted certificates will lose their clientAuth EKU soon (see https://keytalk.com/news/client-authentication-eku-phased-out) 

OpenSIPS requires a certificate when configuring a client_domain, even if mTLS is not required. As a result it's easiest to use the server certificate. However, this means that OpenSIPS will send this certificate (without a clientAuth EKU) if the server requests a client certificate.  

It should be possible to establish the connection without the Client Certificate (see RFC 8446 4.4.2.4 and RFC 5426 7.4.6 ("If no suitable certificate is available, the client MUST send a certificate message containing no certificates.  That is, the certificate_list structure has a length of zero")

I would suggest that it should be valid to set `certificate` and `private_key` to NULL in tls_mgm where type = 2 (client) and that this should sent a certificate message as described above. 

**Implementation**

  - Component: proto_tls or tls_mgm? 
  - Type: Change to existing behaviour 
  - Name: Allow null certificate and private_key in client domains

Generating publicly-trusted TLS certificates with the clientAuth EKU will be impossible soon, as any public CA *must* not include this in issued certificate due to changes to to the Browser Root Program https://www.rsaconference.com/library/blog/sunsetting-the-clientauth-eku-what-why-and-how-to-prepare-for-the-change 

Using self-signed certificates (or private CA) is an option, but requires the other party to accept this CA. In testing, this causes problems with some providers. 

**Additional context**

The correct behaviour is not to send a client certificate if one is not required. Some SIP servers do _request_ the client certificate even if they do not require them for authentication, if OpenSIPS then sends a certificate containing only a serverAuth EKU then the session will likely be rejected. 

(I appreciate that whilst RFC 5280 Requires that If EKU is present and excludes clientAuth, the certificate MUST NOT be used for TLS client authentication, there doesn't seem to be anything in RFC 5246 or 8446 that I can see to preclude requiring the client to enforce EKU when selecting a certificate to send, so technically it's valid to send a certificate that has only a serverAuth EKU but it will likely be rejected in some cases)

rsmck:
**Summary**
Implements a fix for https://github.com/OpenSIPS/opensips/issues/3798 

**Details**
OpenSIPS requires a certificate when configuring a client_domain, even if mTLS is not required. As a result it's easiest to use the server certificate. However, this means that OpenSIPS will send this certificate (these will be without a clientAuth EKU) if the server requests a client certificate. It should be possible to establish the connection without the Client Certificate (see RFC 8446 4.4.2.4 and RFC 5426 7.4.6 ("If no suitable certificate is available, the client MUST send a certificate message containing no certificates. That is, the certificate_list structure has a length of zero")

**Solution**

Server domains: Continue to require certificates (using defaults if not specified), should be no change to behaviour here. 

Client domains: Can now be configured without certificate and private_key parameters.

When the destination TLS server requests a client certificate but one isn't configured, an empty certificate message is sent per RFC 8446 / 5246 

**Compatibility**
Tested with OpenSIPS 3.6 + LibreSSL 3.3.6
Not tested with WolfSSL

**Closing issues**
Closes #3798

### DIFF
--- a/modules/tls_mgm/tls_mgm.c
+++ b/modules/tls_mgm/tls_mgm.c
@@ -642,11 +642,21 @@ static int init_tls_dom(struct tls_domain *d)
 	}
 
 	if (!d->cert.s) {
-		init_flags |= TLS_DOM_CERT_FILE_FL;
-		LM_NOTICE("no certificate for tls domain '%.*s' defined, using default '%s'\n",
-				d->name.len, ZSW(d->name.s), tls_cert_file);
-		d->cert.s = tls_cert_file;
-		d->cert.len = len(tls_cert_file);
+		/* Client domains can operate without a certificate (RFC 8446 4.4.2.4,
+		 * RFC 5246 7.4.6) - an empty certificate message will be sent if the
+		 * server requests one. Server domains require a certificate. */
+		if (d->flags & DOM_FLAG_SRV) {
+			init_flags |= TLS_DOM_CERT_FILE_FL;
+			LM_NOTICE("no certificate for tls server domain '%.*s' defined, "
+					"using default '%s'\n",
+					d->name.len, ZSW(d->name.s), tls_cert_file);
+			d->cert.s = tls_cert_file;
+			d->cert.len = len(tls_cert_file);
+		} else {
+			LM_INFO("no certificate for tls client domain '%.*s', "
+					"will not send client certificate\n",
+					d->name.len, ZSW(d->name.s));
+		}
 	}
 
 	if (!d->ca.s) {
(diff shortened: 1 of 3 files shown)


liviuchircu:
@rsmck and @NormB -- I really appreciate the work and fully documented issue and PR here, bit of a pity that I didn't merge this sooner.  I went ahead and played with it a bit, didn't spot any regressions either.  The patch is fully backwards-compatible too, for _opensips.cfg_ scripts which explicitly set the client `certificate` and `private_key` modparams (all of them?).  But now, you can omit them in order to get the empty-client-cert behavior, which worked just fine against a `openssl s_server ...` CLI TLS test server.

Merging ASAP, I will go for a backport to 3.6 and 3.4 as well, so we keep TLS behavior consistent across all stables.

PR #3834, opened 2026-03-04
event_rabbitmq: fix dupl_string() NUL-inclusive len corrupting AMQP shortstr

### LINKED ISSUE #3828: [BUG] OpenSIPS 3.6.3: AMQP shortstr length includes trailing NUL (exchange ends with \x00)

```
3.6.3~20260210~0e56b0ca4-1
```

**Describe the bug**

After upgrading a test registrar to OpenSIPS 3.6.3, EVI events published via event_rabbitmq started to fail routing in RabbitMQ. Packet captures show the exchange is encoded as an AMQP shortstr with length +1, causing a trailing NUL byte (\x00) to be included in the value (e.g., events.topic.name\x00). The same configuration on PROD (older OpenSIPS build) publishes without the extra NUL.

I am using code:
```
subscribe_event("E_REG_STAT", "rabbitmq:user:pass@1.1.1.1:5672/exchange=events.user.reg?");
```
On 3.3 OpenSIPS pass this in rmq packet - it is correct:
```
events.user.reg\x00\x00\xce
```
On 3.6.3 OpenSIPS pass this in rmq packet - it is broken:
```
events.user.reg\x00\x00\x01\xce
```

Possible problem is here in "modules/event_rabbitmq/event_rabbitmq.c":
```
dst->s[dst->len] = 0;
dst->len++;          /* <----- HERE */
```

**To Reproduce**

Just use `3.6.3~20260210~0e56b0ca4-1` and example of above 'subscribe_event()'

**Expected behavior**

On `3.6.3~20260210~0e56b0ca4-1` should work in same way, that it happened on 3.3.

**Relevant System Logs**

 - Operating System: 
 - OpenSIPS installation: 
 - other relevant information:

**Additional context**

NormB:
**Summary**

`dupl_string()` in `event_rabbitmq.c` increments `dst->len` after NUL-terminating the unescaped string (`dst->len++`), causing `.len` to include the trailing NUL byte. This makes `amqp_basic_publish()` encode exchange and routing-key as AMQP shortstr fields with an extra `0x00` byte, breaking broker routing.

The bug was introduced in commit `0260de0c1` (2014) when `dupl_string()` was refactored to use `un_escape()`. The original `memcpy`-based implementation had the same off-by-one (`dst->len = end - begin + 1`).

**Details**

AMQP 0-9-1 encodes exchange and routing-key as shortstr fields (1-byte length prefix + N content bytes). With the inflated `.len`, the broker receives a shortstr containing the field value followed by `0x00`, which fails to match the declared exchange or creates one with an embedded NUL.

Over time, several `- 1` compensations were added downstream to work around the inflated length:
- `tls_dom_name.len--` in `rmq_parse()` (commit `472ae546c`, TLS support)
- `address.len - 1`, `exchange.len - 1`, `routing_key.len - 1` in `rmq_print()`
- `strlen(user) - 1` in `rmq_print()` — this one was wrong even pre-patch since `strlen()` never counts the NUL

**Solution**

Remove the `dst->len++` in `dupl_string()` and all downstream compensations. Three additional issues fixed while testing:

1. **Memory leak**: `dupl_string()` calls `shm_malloc` before `un_escape()`, but on `un_escape()` failure the buffer was never freed. Added `shm_free` + cleanup.
2. **Default user NUL termination**: `shm_malloc(rmq_static_holder.len)` + `memcpy(..., len)` for `"guest"` didn't copy the NUL terminator. The pointer is passed to `amqp_login()` → `strlen()`. Changed to `len + 1`.
3. **`strlen(user) - 1`**: The `/* skip 0 */` comment reveals the author believed `strlen()` includes the NUL. It doesn't — this always truncated the last character of the username in `rmq_print()` output.

Tested with:
- AMQP wire capture (tshark) confirming correct shortstr encoding
- TLS domain name lookup (instrumented debug + regression simulation)
- Standalone test harness (1019 lines) with 100K-iteration stress test confirming shm leak fix

**Compatibility**

No backward compatibility issues. All behavioral changes are strictly correctional:
- Wire encoding: NUL byte no longer included (fixes the reported bug)
- `rmq_match()`: symmetric — both sides were inflated equally, no change
- `rmq_print()`: compensations removed, output unchanged
- `amqp_login()` user/password: uses `char*` via `strlen()`, unaffected

**Closing issues**

Closes #3828

### DIFF
--- a/modules/event_rabbitmq/event_rabbitmq.c
+++ b/modules/event_rabbitmq/event_rabbitmq.c
@@ -336,12 +336,15 @@ static inline int dupl_string(str* dst, const char* begin, const char* end)
 		return -1;
 	}
 
-	if (un_escape(&tmp, dst) < 0)
+	if (un_escape(&tmp, dst) < 0) {
+		shm_free(dst->s);
+		dst->s = NULL;
+		dst->len = 0;
 		return -1;
+	}
 
 	/* NULL-terminate the string */
 	dst->s[dst->len] = 0;
-	dst->len++;
 
 	return 0;
 }
@@ -516,7 +519,6 @@ static evi_reply_sock* rmq_parse(str socket)
 						if (dupl_string(&param->conn.tls_dom_name,
 							it->s.s+RMQ_TLS_DOM_LEN, it->s.s + it->s.len) < 0)
 							goto err;
-						param->conn.tls_dom_name.len--;
 						param->conn.flags |= RMQ_PARAM_TLS;
 					} else if (it->s.len == RMQ_PERSISTENT_LEN &&
 						!memcmp(it->s.s, RMQ_PERSISTENT_S, RMQ_PERSISTENT_LEN)) {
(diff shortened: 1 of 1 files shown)


razvancrainea:
This looks good, thanks for taking care of it!

Issue #3712, opened 2025-08-31
[FEATURE]Janus websocket fragmentation


Olatechworld:
Hello Opesisps team,
I am currently using opensipsjs and the new Janus module from opensips to send commands to Janus. All requests seem to be working fine, but the subscribe ptype subscribe request returns the following error:

2025-08-31 23:01:03.092	
{ "time": "Aug 31 21:01:03", "level": "ERROR", "pid": 22, "message": **"Failed to get janus answer !**!! :( ", "function": "w_janus_send_request", "module": "janus" }
	2025-08-31 23:01:02.559	
{ "time": "Aug 31 21:01:02", "level": "ERROR", "pid": 10, "message": "Failed JANUS process - releasing conn ", "function": "janus_handle_data", "module": "janus" }
	2025-08-31 23:01:02.557	
{ "time": "Aug 31 21:01:02", "level": "ERROR", "pid": 10, "message": "Failed to read from janus on 80", "function": "handle_io", "module": "janus" }
	2025-08-31 23:01:02.557	
{ "time": "Aug 31 21:01:02", "level": "ERROR", "pid": 10, "message": "**We do not support fragmemntation yet. Dropping...**", "function": "janus_ws_parse", "module": "janus" }

Thank you in advance for your help

<img width="1359" height="150" alt="Image" src="https://github.com/user-attachments/assets/ac0d87a1-4a6e-4f86-a216-953bf6e3b4fa" />

razvancrainea:
Can you please post a pcap for this?

Olatechworld:
[pcapopensipsjanus.zip](https://github.com/user-attachments/files/22089809/pcapopensipsjanus.zip)

Please find attached .zip (pcap and some of the logs)

razvancrainea:
Indeed, Janus is sending a fragmented message in frame 221, which is currently not supported by OpenSIPS. I will convert this ticket to a feature request until we add support for this.

PR #3839, opened 2026-03-07
janus: fix pkg memory leaks in cJSON_Print/cJSON_Parse paths

NormB:
**Summary**

Fix three pkg memory leaks and a potential NULL dereference crash in the janus module's cJSON usage paths.

**Details**

The janus module uses `cJSON_InitHooks()` to route all cJSON allocations through `pkg_malloc`. Three call sites had missing cleanup that leaked pkg memory on every `janus_send_request()` call (~350 bytes/request total). Under sustained load, the SIP worker pkg pool is exhausted, causing `cJSON_Print()` to return NULL and a subsequent `strlen(NULL)` crash.

Found during stress testing of the fragmentation fix for #3712 — the memory leaks are independent of the fragmentation issue.

**Solution**

1. **`janus_ipc_send_request()` in `janus_proc.c`**: `cJSON_Print()` allocates from the SIP worker's pkg pool. The result was copied to shm via `shm_nt_str_dup()` but the original was never freed. Added `pkg_free(full_cmd.s)` after the shm copy and a NULL check before `strlen()`.

2. **`w_janus_send_request()` in `janus_mod.c`**: `cJSON_Parse()` builds a cJSON tree from pkg. The tree was serialized and sent via IPC but `cJSON_Delete()` was never called. Added cleanup after `janus_ipc_send_request()` returns and on the `get_janus_connection_by_id()` error path.

3. **`janus_common.c`**: Added NULL checks after `cJSON_Print()` in `janus_raise_event()` and `handle_janus_json_request()`. Fixed missing `pkg_free(full_json)` on the `shm_strdup()` failure path.

Verified with 100k messages — zero pkg memory growth across all SIP workers (MI `pkmem:` stats flat at 228.3 kB).

**Compatibility**

No behavioral changes for correct inputs. Error paths that previously crashed now return appropriate error codes.

**Closing issues**

Related to #3712 (the memory leaks surface under sustained janus module usage)

### DIFF
--- a/modules/janus/janus_common.c
+++ b/modules/janus/janus_common.c
@@ -112,6 +112,10 @@ int janus_raise_event(janus_connection *conn, cJSON *request)
 	}
 
 	full_json = cJSON_Print(request);
+	if (!full_json) {
+		LM_ERR("cJSON_Print failed\n");
+		goto err_free_params;
+	}
 	cJSON_Minify(full_json);
 	full_json_s.s = full_json;
 	full_json_s.len = strlen(full_json);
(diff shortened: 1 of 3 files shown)


liviuchircu (approved):
Great finds, @NormB!   These are not just leak fixes on the error edge-cases (e.g. out-of-memory, JSON printing errors, etc.), but also on the happy-cases!

Merging & backporting to all stable branches.

PR #3852, opened 2026-03-29
cachedb_redis: add ASK redirect handling for cluster resharding

dondetir:
## Summary

Add support for Redis ASK redirects during cluster resharding, completing the cluster redirect handling started in PR #3639.

### Background

When a Redis cluster is resharding (migrating slots between nodes), the source node returns an `ASK` response instead of `MOVED`. The key difference:

- **MOVED** = permanent redirect (slot ownership changed) — already handled by PR #3639
- **ASK** = temporary redirect (slot is mid-migration) — requires sending `ASKING` to the target node before retrying

Without ASK handling, queries during resharding operations fail with unhandled error responses.

### Implementation

- Detects ASK responses alongside existing MOVED handling in the query retry loop
- Sends `ASKING` command to the target node before retrying the original query
- Does NOT update the slot table (correct per Redis spec — ASK is one-time, unlike MOVED)
- Refactors `parse_moved_reply()` into `parse_redirect_reply()` that accepts the prefix as a parameter, with inline wrappers `parse_moved_reply()` and `parse_ask_reply()` for backward compatibility

### Note on hash slot fix

The original version of this PR also included a hash slot calculation fix. That has been removed per @NormB's feedback, as his PR #3815 addresses this more comprehensively (including hash tag support). This PR now contains only the ASK redirect handling.

### Testing

- Builds clean, zero warnings
- All 2545 unit tests pass
- @NormB verified the `parse_redirect_reply()` refactor passes integration tests with live slot migrations across a 3-node cluster

Partially addresses #2811

### DIFF
--- a/modules/cachedb_redis/cachedb_redis_dbase.c
+++ b/modules/cachedb_redis/cachedb_redis_dbase.c
@@ -564,48 +564,83 @@ static int _redis_run_command(cachedb_con *connection, redisReply **rpl, str *ke
 					reply,reply?(unsigned)reply->len:7,reply?reply->str:"FAILURE",
 					node->context->errstr);
 
-				if (match_prefix(reply->str, reply->len, MOVED_PREFIX, MOVED_PREFIX_LEN)) {
-    					// It's a MOVED response
+				if (match_prefix(reply->str, reply->len, MOVED_PREFIX, MOVED_PREFIX_LEN) ||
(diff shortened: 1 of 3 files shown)


liviuchircu (approved):
Great work here, @dondetir -- LGTM.  As a plus, the patch is backwards-compatible too, as it purely adds support for "ASK" redirects, complementing the existing MOVED redirection support added in #3639 last year.  Merging.

PR #3854, opened 2026-03-30
cachedb_redis: fix safety issues in cluster redirect parsing

NormB:
## Summary

Fix several correctness and safety issues in `parse_moved_reply()` and the MOVED redirect handler.

## Details

**Integer overflow in slot parsing** — the digit accumulation loop `slot = slot * 10 + (*p - '0')` has no upper bound check. A malformed MOVED reply with a long digit string causes signed integer overflow (undefined behavior in C). Fixed by returning `ERR_INVALID_SLOT` when `slot > 16383` during parsing.

**Integer overflow in port parsing** — same issue in the port digit loop. The existing post-loop check `port > 65535` is insufficient because signed overflow is UB before the check executes. Fixed by returning `ERR_INVALID_PORT` when `port > 65535` during digit accumulation.

**Undefined behavior in no-colon fallback** — when the endpoint has no colon (no explicit port), the code compares `out->endpoint.s < end` where `out->endpoint.s` was set to `NULL` two lines above. Comparing a null pointer to a non-null pointer is undefined behavior per the C standard. Fixed by using `p < end` instead, which holds the correct scan position.

**Unnecessary heap allocation** — the `redis_moved` struct in the MOVED handler is heap-allocated via `pkg_malloc` and freed within the same scope. Since the struct is ~24 bytes and never outlives the enclosing block, this is replaced with stack allocation, eliminating the `pkg_malloc`/`pkg_free` calls and the OOM error path.

## Compatibility

No behavioral change for correctly-formed Redis replies. The overflow checks add early rejection of malformed input that would previously have caused undefined behavior.

## Closing issues

Partially addresses #2811

### DIFF
--- a/modules/cachedb_redis/cachedb_redis_dbase.c
+++ b/modules/cachedb_redis/cachedb_redis_dbase.c
@@ -565,29 +565,24 @@ static int _redis_run_command(cachedb_con *connection, redisReply **rpl, str *ke
 					node->context->errstr);
 
 				if (match_prefix(reply->str, reply->len, MOVED_PREFIX, MOVED_PREFIX_LEN)) {
-    					// It's a MOVED response
-					redis_moved *moved_info = pkg_malloc(sizeof(redis_moved));
-						if (!moved_info) {
-						LM_ERR("cachedb_redis: Unable to allocate redis_moved struct, no more pkg memory\n");
-							freeReplyObject(reply);
-							reply = NULL;
-							goto try_next_con;
-					} else {
-							if (parse_moved_reply(reply, moved_info) < 0) {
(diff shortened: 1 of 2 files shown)


liviuchircu:
LGTM - some common-sense improvements, can only do good.  Solved conflicts & merged.

liviuchircu:
Also, @NormB in case it helps you, note that in this PR, the commit was not done by your email, rather by:

```
commit e0ead4f57b4468c9d453622d8245d0166d27c74d
Author: Debian <g...r@o...com>
Date:   Mon Mar 30 04:30:49 2026 +0000
```

PR #3856, opened 2026-03-30
cachedb_redis: add Unix socket transport and lazy connection

NormB:
## Summary

- Add Unix domain socket support as an alternative to TCP
- Add `lazy_connect` parameter to defer connection until first use

### Unix socket transport

The module now supports connecting to Redis via Unix domain sockets using a query parameter in the `cachedb_url`:

```
modparam("cachedb_redis", "cachedb_url",
    "redis:local://localhost/?socket=/var/run/redis/redis.sock")
```

The `REDIS_UNIX_SOCKET` flag distinguishes socket connections from TCP. MI output includes `transport` (tcp/unix) and `socket_path` fields. Unix socket connections are supported in both single-instance and cluster topology refresh paths.

### Lazy connection

The new `lazy_connect` parameter (integer, default 0) defers the Redis connection from `child_init` to the first cache operation. This is useful when:
- Redis may not be available at OpenSIPS startup
- Multiple `cachedb_url` groups are configured but not all are needed immediately
- Startup time matters and Redis connection latency is non-trivial

Both TCP and Unix socket connections support lazy mode.

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `lazy_connect` | integer | 0 | Defer Redis connection until first cache operation |

### Testing

| Suite | Tests | Description |
|-------|-------|-------------|
| test_unix_socket | 19 | Store/fetch, MI reporting, PING latency, recovery |
| test_lazy_connect | 17 | Deferred connect for all transport types, recovery |

## Compatibility

No behavioral change for existing TCP configurations. The `lazy_connect` parameter defaults to 0 (disabled).

## Dependencies

- Depends on #3855 (cluster topology management)


dondetir (commented):
Nice addition @NormB — Unix socket support and lazy connect are both clean and well-tested.

I verified both features against a local setup:
- Unix socket: store/fetch, MI reporting (`"transport": "unix"`), ping — all working
- Lazy connect: confirmed empty node list before first operation, then full cluster auto-discovery on demand

One small thing: in `redis_get_ctx()` (line ~83) and `redis_get_ctx_unix()` (line ~129), the `if (!ctx)` NULL check after `redisConnect()`/`redisConnectUnix()` seems to have been dropped during the rewrite. The `if (ctx && ctx->err != REDIS_OK)` guard handles the error-context case, but if hiredis returns NULL (OOM), execution falls through to `redisSetTimeout(ctx, ...)`. It looks like commit `8fb569cb3` had the fix but `16652c169` lost it in the rewrite. Easy to re-add:

```c
if (!ctx) {
    LM_ERR("failed to connect to redis - out of memory\n");
    return NULL;
}
```

Everything else looks solid. The Unix socket + cluster mode guard and the lazy connect flow are well thought out.

modules/cachedb_redis/cachedb_redis_dbase.c
```diff
@@ -1859,7 +2104,6 @@ int redis_raw_query_handle_reply(redisReply *reply,cdb_raw_entry ***ret,
 							}
 						}
 
-
```
liviuchircu, line 1862: Please avoid doing these random whitespace edits in the future -- the PR is already big enough, no need to include additional code styling.

modules/cachedb_redis/doc/cachedb_redis_admin.xml
```diff
@@ -213,7 +210,7 @@
 			<itemizedlist>
 			<listitem>
 			<para>
-				<emphasis>If a <xref linkend="param_use_tls"/> is defined, the <emphasis role="bold">tls_mgm</emphasis> and <emphasis role="bold">tls_openssl</emphasis> modules will need to be loaded as well</emphasis>.
+				<emphasis>If a <xref linkend="param_use_tls"/> is defined, the <emphasis role='bold'>tls_mgm</emphasis> module will need to be loaded as well</emphasis>.
```
liviuchircu, line 213: I think something went wrong with this `cachedb_redis_admin.xml` change, as it seems to be undo-ing the work done in 7b8fe71ffd.

modules/cachedb_redis/cachedb_redis_utils.c
```diff
@@ -316,7 +318,7 @@ int parse_cluster_shards(redis_con *con, redisReply *reply)
 			}
 		}
 
-		if (!master_ip || master_port < 1 || master_port > 65535)
+		if (!master_ip || master_port <= 0 || master_port > 65535)
```
liviuchircu, line 321: What is this?

modules/cachedb_redis/cachedb_redis_utils.c
```diff
@@ -565,7 +565,6 @@ static int parse_redirect_reply(redisReply *reply, redis_moved *out,
 	while (p < end && *p >= '0' && *p <= '9') {
 		slot = slot * 10 + (*p - '0');
 		p++;
-		if (slot > 16383) return ERR_INVALID_SLOT;
```
liviuchircu, line 568: Similar issue as earlier.  Probably a commit done separately, so when put together in a branch, it cancels work done in commits such as e0ead4f57.  I'll try to cancel this diff out...

liviuchircu (approved):
Great work, @NormB -- let's see the New UNIX URL format in action.  The feature is quite safe and also backwards-compatible, so I will merge it.

Still, there are some strange issues with the diff, as it seems to randomly cancel out some of the works done in earlier commits of yours, such as e0ead4f5 or 7b8fe71ffd.  I'll try to remove these diffs as part of the merge, but do take another look yourself at the final result.

PR #3866, opened 2026-04-07
lib/dbg: enhance opensips debug capabilites with backtrace

ovidiusas:
**Summary**
 - sometimed the file/function/line is not enough to troubleshoot
 - a backtrace gives more details about the code path

**Details**

By default, this new feature is not available/compiled in.
Use `-DBACKTRACE_DBG` to enable it (see `Makefile.conf`).
This will add backtrace info for each log into the log buffer.

### DIFF
--- a/lib/dbg/struct_hist.c
+++ b/lib/dbg/struct_hist.c
@@ -21,6 +21,12 @@
 #include <stdio.h>
 #include <stdarg.h>
 
+#ifdef BACKTRACE_DBG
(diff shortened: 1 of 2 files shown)


liviuchircu:
I finally got to test & merge this one, @ovidiusas!  While it does add quite a bit of logging, the patch is well-guarded by the `-DBACKTRACE_DBG`, so it's a nice resource to have, for debugging builds.

PR #3883, opened 2026-04-29
mid_registrar/usrloc: fix De-REGISTER failures on cluster peers (#3845)

### LINKED ISSUE #3845: [BUG] ERROR:mid_registrar:unregister_record: 'from' key not found, skipping De-REGISTER
```
OpenSIPS v3.6.2
```

**Describe the bug**
```
Hey!

The problem occurs when the following modules have been configured:
clusterer + mid_registrar + usrloc (full-sharing-cluster)

The schema is classic:  User --> OpenSIPS cluser (anycast IP + Path) --> Registrar
```

**To Reproduce**
```
Simply configure the clusterer + mid_registrar + usrloc (full-sharing-cluster) and perform the registration.
```
Below is Claude's report on the analysis of the issue: [CLUSTER_REPLICATION_BUG_REPORT.md](https://github.com/user-attachments/files/26019433/CLUSTER_REPLICATION_BUG_REPORT.1.md)

I appreciate that @liviuchircu worked on the issue https://github.com/OpenSIPS/opensips/commit/343f452c1, but there are probably still some things left to fix.

Thanks in advance!

denyspozniak:
Description

  Closes #3845

  ### Problem

  In `full-sharing-cluster` mode, every replicated AoR insert logs the
  following on peer nodes, and the AoR can never be De-REGISTER'd
  upstream when it expires:

  ERROR:usrloc:store_deserialize: bad JSON input or oom
  ERROR:mid_registrar:unregister_record: 'from' key not found, skipping De-REGISTER
  ERROR:mid_registrar:mid_reg_aor_event: failed to unregister contact

  ### Cause

  Since `343f452c1`, `replicate_urecord_insert()` ships
  `r->kv_storage`. But in `mid_registrar` (`save_restore_req_contacts`)
  `store_urecord_data()` runs *after* `ul.insert_urecord()`, so the
  INSERT packet is broadcast with an empty map. Peers `cJSON_Parse("")`
  → NULL, and the missing keys (`from`, `to`, `callid`, `main_reg_uri`,
  `ct_uri`, `last_cseq`) break De-REGISTER on expiry.

  ### Fix

  Mirror the existing ucontact pattern (`pre_replicate_cb`): add an
  optional pre-replication callback to `insert_urecord()` so the caller
  can fill `kv_storage` between `mem_insert_urecord()` and
  `replicate_urecord_insert()`. `mid_registrar` plugs
  `mid_reg_store_aor_data` into that hook in the throttle-AoR path.
  All other call sites pass `NULL, NULL` — no behaviour change.

  ### Verified

  2-node `127.0.0.1` cluster, `mid_registrar` mode 2 + `usrloc`
  `full-sharing-cluster`. After the fix, node2's `bin_pop_str` for
  `REPL_URECORD_INSERT` already contains the full kv map, no
  `store_deserialize` or `unregister_record` errors, and De-REGISTER
  reaches the main registrar on expiry.

### DIFF
--- a/modules/mid_registrar/save.c
+++ b/modules/mid_registrar/save.c
@@ -236,6 +236,14 @@ struct mr_ct_data {
 	int last_cseq;
 };
 
+struct mr_aor_data {
+	struct mid_reg_info *mri;
+	const str *ct_uri;
+	int expires_out;
+	int last_reg_ts;
+	int last_cseq;
+};
+
 static int mid_reg_store_ct_data(ucontact_t *c, void *info)
 {
 	struct mr_ct_data *data = (struct mr_ct_data *)info;
@@ -245,6 +253,19 @@ static int mid_reg_store_ct_data(ucontact_t *c, void *info)
 		data->expires_out, data->last_reg_ts, data->last_cseq);
 	if (rc != 0)
 		LM_ERR("failed to attach ucontact data - oom?\n");
+
+	return rc;
+}
+
+static int mid_reg_store_aor_data(urecord_t *r, void *info)
+{
+	struct mr_aor_data *data = (struct mr_aor_data *)info;
+	int rc;
+
+	rc = store_urecord_data(r, data->mri, data->ct_uri, data->expires_out,
+		data->last_reg_ts, data->last_cseq);
+	if (rc != 0)
+		LM_ERR("failed to attach urecord data - oom?\n");
 
 	return rc;
 }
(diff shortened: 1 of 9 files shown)


liviuchircu:
@denyspozniak I allocated some time to review this one.  Indeed, in `343f452c1` we tried to quick-fix those errors on the backup box, but unfortunately, despite we replicated the dictionary, its data was not fully populated when building the "AoR inserted replication" packet.

Your fix seems reasonable -- while it adds a couple params to the `.insert_urecord()` API which will be invoked as `NULL, NULL` from a lot of places, it's still a clean way of solving all locking issues around creating the AoR and populating the key/values just before replicating the packet -- everything done by usrloc.

LGTM, let's merge it.  Will also backport down to 3.6 as a fix.

PR #3885, opened 2026-05-04
add ip.resolve4/resolve6 transformations

### LINKED ISSUE #3781: [FEATURE] Add possibility to explicitly request address family on DNS lookups

I'm having a dual-stack (IPv4/IPv6) system. For testing and maintenance reasons I need to select which address family I need to use in every particular request.

One of the possible solutions - on forcing outbound socket to a specific address family, DNS request results are adopted to address family of the outbound socket.

Some core function like `dns_query()` that returns all IP addresses resolved might be helpful.

Other option might be in string transformation of {ip.resolve} where the family of address might be specified.

For the last 2 options, I'm not sure if setting $du to ipv4 (for example) address will fill all the headers (like Via) in a correct format, plus, provide seamless IPv4/v6 bridge.

**Implementation**

Not sure at all, but I guess, `resolve.c` in `core`

None for the moment.

**Additional context**

I need an ability to change protocol and address family during one call, like in one call with re-INVITE change proto and address family (for testing reasons)

igorolhovskiy:
**Summary**
Adding explicit transformations to get IPv4/IPv6 address resolution in a script

**Details**
More details can be found in https://github.com/OpenSIPS/opensips/issues/3781.

**Solution**
Per advice of @NormB , new function `resolvehost_af` was introduced, keeping `resolvehost` for backward compatibility.

**Compatibility**
Not found, as it's a new functionality.

**Closing issues**
Closes #3781 

**Additional notes**
Documentation need to be updated accordingly


bogdan-iancu:
Hi @igorolhovskiy , thanks for the contribution, it looks useful and clean. Only one thing to do - now with the new documentation system in place, please include in your PR the doc side for this transformation - see the docs/manual/Script-Tran.md file. Thanks!

PR #3886, opened 2026-05-05
Fix sip instance in REGISTER reply

dsandras:
The SIP Instance of the 200 OK response is not correctly formatted.

It should be surrounded by angle brackets following section 4.1 from RFC
5626:
    
      [RFC3840] defines equality rules for callee capabilities parameters,
      and according to that specification, the "sip.instance" media feature
      tag will be compared by case- sensitive string comparison.  This means
      that the URN will be encapsulated by angle brackets ("<" and ">") when
      it is placed within the quoted string value of the "+sip.instance"
      Contact header field parameter.

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/registrar/reply.c
+++ b/modules/registrar/reply.c
@@ -163,7 +163,7 @@ static inline unsigned int calc_buf_len(ucontact_t* c,int build_gruu,
 				/* sip.instance */
 				len += SIP_INSTANCE_SIZE
 					+ 1 /* quote */
-					+ (c->instance.len - 2)
+					+ c->instance.len
 					+ 1 /* quote */
 					;
 			}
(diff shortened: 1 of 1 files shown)


dsandras:
@liviuchircu @bogdan-iancu Hi! 

Maybe one of you could check this one and consider it to be included? Thanks a lot. 

Damien

dsandras:
Hello,

Any interest here?

Thanks

dsandras:
@bogdan-iancu Hi, could you consider this pull request? It seems trivial. Thanks.

bogdan-iancu:
Hi @dsandras , this is on @liviuchircu  's todo list, he will check on it and update. Thanks !

dsandras:
Hi @bogdan-iancu thanks!

liviuchircu:
Hi @dsandras!  I finally took some time to look into this one.  The fix is quite specific, as it addresses only REGISTER packets with `Supported: gruu`, when the `disable_gruu` on the server-side is explicitly flipped to `0`... So the risk of breakage is near-zero.

Now, on the actual fix:  I added a couple unit tests and tested the GRUU stuff, with the `+sip.instance="<urn:uuid:...` Contact output, which passed just fine after your fix.  I'll get them pushed as well.  So, just merging & backporting here, cheers!

dsandras:
Thanks @liviuchircu !

PR #3890, opened 2026-05-06
dialog: add options_ping_max_retries parameter for OPTIONS ping retries

digipigeon:
Summary

Add a new options_ping_max_retries module parameter to the dialog module that allows configuring additional OPTIONS ping attempts before declaring a dialog leg dead and terminating the call.

Details

In UDP deployments, a single OPTIONS ping request or its 200 OK response can be lost in transit. With the current behavior, one lost packet immediately terminates an otherwise healthy call, which can be disruptive for end users.

This is particularly problematic in networks with occasional packet loss where calls are functioning normally but a single missed keepalive ping triggers unnecessary call teardown.

Solution

Introduce a new parameter options_ping_max_retries (default: 0) that specifies how many additional OPTIONS pings to send after a timeout before declaring the leg dead:

When an OPTIONS ping times out (408/FAKED_REPLY), the module now checks if retries remain
Retries are sent immediately from the reply callback, not waiting for the next options_ping_interval tick
A per-leg ping_retries counter tracks failed attempts and resets on any successful response
481 responses are treated as definitive failures and not retried
Worst-case detection time: (options_ping_max_retries + 1) * fr_timeout
Changes:

Added options_ping_max_retries parameter in dialog.c
Added ping_retries field to dlg_leg structure
Implemented retry logic in dlg_handle_seq_reply() in dlg_timer.c
Added comprehensive documentation in dialog_admin.xml
Example usage:

Compatibility

Fully backward compatible. The default value of 0 preserves the existing behavior where a single failed OPTIONS ping immediately terminates the dialog.

### DIFF
--- a/modules/dialog/README.md
+++ b/modules/dialog/README.md
@@ -446,6 +446,32 @@ modparam("dialog", "options_ping_interval", 20)
 ```
 
 
+#### options_ping_max_retries (integer)
+
+
+The number of additional in-dialog OPTIONS pings to send when the
+previous OPTIONS transaction times out, before declaring the dialog leg
+unreachable. Retries are sent immediately after each transaction timeout,
+without waiting for the next `options_ping_interval` tick. The retry counter
+is maintained independently for each dialog leg and is reset by any final
+response other than 408 or 481. A 481 response is considered definitive and
+is not retried.
+
+The maximum time needed to detect an unreachable peer is approximately
+`(options_ping_max_retries + 1) * fr_timeout`, where `fr_timeout` is configured
+by the "tm" module.
+
+
+*Default value is "0" (no additional retries).*
+
+
+```opensips title="Set options_ping_max_retries parameter"
+...
+modparam("dialog", "options_ping_max_retries", 2)
+...
+```
+
+
 #### reinvite_ping_interval (integer)
 
 
--- a/modules/dialog/dialog.c
+++ b/modules/dialog/dialog.c
@@ -72,6 +72,7 @@ static char* profiles_nv_s = NULL;
 int dlg_bulk_del_no = 1; /* delete one by one */
 int seq_match_mode = SEQ_MATCH_FALLBACK;
 int options_ping_interval = 30;      /* seconds */
+int options_ping_max_retries = 0;    /* extra OPTIONS pings to send on timeout before declaring leg dead */
 int reinvite_ping_interval = 300;    /* seconds */
 int dlg_del_delay = 0;               /* in seconds, default off */
 str dlg_extra_hdrs = {NULL,0};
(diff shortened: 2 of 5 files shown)


bogdan-iancu:
Hi @digipigeon , thanks for the contribution here . I understand that your change introduces a bit of tolerance to the OPTIONS pinging (like 3 pings in a row must fail in order to terminate the call, instead of a single one), but I need to correct you on a statement.  You say:

>In UDP deployments, a single OPTIONS ping request or its 200 OK response can be lost in transit. With the current behavior, one lost packet immediately terminates an otherwise healthy call, which can be disruptive for end users.

And this is not true at all - one single lost packet does not trigger the call termination - the sending of the OPTIONS is stateful, meaning it is subject to retransmissions (for the request), to it is tolerant to a single packet lost, even on UDP.
Again, I understand the idea behind your PR, but I wanted to clarify this statement, things are not so dramatic as you described :) 

Now, to move on with this PR, we need two more things from your side:
* do a rebase as the patch does not apply anymore
* update the docs - drop the changes over the XML file (which does not exist anymore) and update the README.md file. (maybe do this before the above point, at the conflicts may be because of all these doc related changes)

digipigeon:
Hi Bogdan, you are right. It looks like I overlooked these retransmissions due to a seperate bug in the tracer module.

I have prepared a PR for that one as well here: https://github.com/OpenSIPS/opensips/pull/4227

I will resolve the outstanding requests on this one shortly.

bogdan-iancu:
Hi @digipigeon - thanks for updating here. Nevertheless, something is puzzling me. Maybe I misunderstand something, but in this implementation, upon OPTIONS timeout, you start doing more OPTIONS on the spot (without waiting for the pinging interval) and also without disabling the timer OPTIONS pinging (based on the pinging interval) - isn't this a possible overlapping in pining?
Ex: ping interval is 30 sec. The fr_timer is 4 seconds. The options_ping_max_retries is 10. In this case, when the first ping (triggered by the timer) fails with timeout, it will cascade in a chain of 10 pings, each of 4 seconds timeout -> 40 seconds until giving up. And in this time, the timer will generate yet another OPTIONS (after 30 sec), as the DLG_PING_SUCCESS is set. Which is wrong :( 

And one more thing - do we really want this burst of OPTIONS upon timeout ? For example, the dispatcher module is simply counting the failed pings before failing - see https://docs.opensips.org/manual/3-5/modules/dispatcher/#param_ds_probing_threshold

digipigeon:
Hi Bogdan, you're right. I did a version before where it would keep the ping interval the same, but not seeing the retransmissions because of #4227 I modified this to essentially behave like a retransmissions. I think it would be better to revert back to the previous fixed interval behaviour, please leave this one with me.

PR #3892, opened 2026-05-15
core: implement FAST_LOCK for aarch64

ovidiusas:
**Summary**
Implement FAST_LOCK for aarch64


razvancrainea:
This looks legit, but I have no means to replicate it. I will push it upstream in master until we have further details on it.

PR #3908, opened 2026-06-05
Transaction only topology_hiding enhancements

davidtrihy-genesys:
## Summary
Additional features to the topology_hiding module in transaction only mode.

1. Compact encoding of the thinfo param
2. One way hiding based on socket tags, this allows for asymmetric transaction topology hiding where the request may not strip any information but the response does or vice versa
3. Selection of the encoding algorithm, legacy or new compact encoding along with password rotation support
4. New autoroute functionality toggleable on or off

## Details
A strong effort was made to not alter dialog topology_hiding functionality and only target transaction functionality but there was a refactor done to separate the logic as the transaction aware logic grew in size due to the changes so a lot of the common code was refactored into separate source files to make development easier and additional functionality that could be common was added to those same files

- This is the refactor of the logic in a commit `75ff8a6bdd87466f44a5c8dbeac417cf82493ae4` the idea is to show that there is no functionality change that would affect the dialog aware module
- This is the new codec for the compact encoding `44b86215d8bc06137e4e7a44ad549f645576d21b`
- pseudovars support commit `ac6d588b96741d6aaf6597528b56d2abc6462889`
- This is the modparam changes `bbba1f626f1478a040e947ef246da9bd66656ba3`

The major logic rework is part of this commit here `11783c1c4faa561541e7b0e26ee4bffa9adb88a0` this is where the bulk of the work was done to support the new model of topology_hiding there are other smaller commits too but these ones give the bulk of changes.

Once the modparam name and scheme has been agreed upon the documentation can be added in another commit before merged the PR

**Socket tag matching**

Purpose of these tags is to define the sockets that will be used for topology hiding, internal is when the request comes in and the socket is switched, it will allow the topology_hiding to be engaged but also not strip topology information for requests and only for responses, the inverse applies for requests initiated from internal.

External is purely for failover scenarios, in this case if you don't have a DNS advertised for your internal socket and that instance fails, you can do some script initiated failover to another instance that may have the same IP address but has the same tag and it should be able to decode the socket information using the tags to match the information as opposed to IP addresses
If this tag is not set then failover will not work in this scenario, this only affects the auto_route

```
modparam("topology_hiding", "th_internal_trusted_tag", "my_internal_socket_tag")
modparam("topology_hiding", "th_external_socket_tag", "my_external_socket_tag")

socket = tcp:203.0.100.10:5060 tag my_external_socket_tag
socket = udp:10.100.20.40:5080 tag my_internal_socket_tag
```

**auto route modparam**

This modparam is purely to generate an auto route on the trusted internal socket when one way hiding is engaged due to no Contact header update, the script writer can turn it off and choose to do sequential requests in the script
```
modparam("topology_hiding", "th_auto_route_on_trusted_socket", 1)
```

**param encoding and password**

Allows for setting the param name, password and encoding type and choosing which one to use when doing topology_hiding will allow two definitions and supports password rotation
```
modparam("topology_hiding", "th_contact_encode_param_password", "thinfo:password1:C")
modparam("topology_hiding", "th_use_param", "thinfo")
```

## Solution
The solution has some limitations that could be addressed in future with modparams but I don't think it's entirely necessary for now

1. New pseudovars do not support legacy encoding scheme, they were written to directly read the decoded buffer of the compact encoding rather than copy
2. Parts of the buffer management were moved from using pkg_malloc to using static buffers of fixed size, the sizing is 4096 bytes and if the decoding goes above that size it will fail to decode, this could also be made as a modparam and the allocation of the buffer could be done at module init time, the idea is to remove unneeded dynamic allocations during transaction processing
3. The amount of URIs that can be decoded is fixed at 12, it was a random number I picked, once again it could be made as a modparam or fully dynamic

These listed limitations only really apply to the new encoding type and the target usage of this algorithm is in the one way hiding mode of operation which the script writer should be aware of the internal SIP message Routes and Contact headers so they should be aware of the limitations, in any case I plan to address them in a follow up PR

## Compatibility
Retains full backwards compatibility with the current mode of operation.

sipssert tests PR - https://github.com/OpenSIPS/sipssert-opensips-tests/pull/32


modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+        goto error;
+    }
+
+    return 1;
+error:
+	shm_free_wrap(p);
+	return -1;
+}
+
+static int th_no_dlg_auto_route_seq_handling(struct sip_msg *msg, rr_t auto_route[static 1], str thinfo[static 1], int self_route, const thinfo_options_t *options) {
```
david-trihy, line 228: This is the function that handles when you call `topology_hiding_match()` in the script on a request that has the auto route generated, it will decode the thinfo in that Route header to get the socket information

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+	decoded_uri_buf.pos = 0;
+
+	if (thinfo_decode_socket(&decoded_uri_buf, &proto, &host, &port) <= 0) {
+		LM_ERR("Failed to decode socket 0\n");
+		return -1;
+	}
+
+	LM_DBG("Decoded socket host [%.*s] - Port - %d - Proto %d\n", host.len, host.s, port, proto);
+
+	if (self_route) {
```
david-trihy, line 279: This branch is for the case where we identify an auto Route header and if there is the th_external_socket_tag defined then we can use that socket in the scenario where the advertised address in the Route header is not a local domain or socket interface. This is to satisfy the scenario where the server that originally did the topology_hiding has failed and you're routing sequential requests in a fallback scenario and identifying the auto Route by the param and removing it before using the external tag to get the socket instead of getting it from the Route header in the sipssert tests the scenario `10.internal-socket-tag-match` and `11.internal-socket-tag-mismatch` tests this functionality and why it could potentially be needed.

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+			pkg_free(route_free_str);
+		}
+		return TOPOH_MATCH_FAILURE;
+	}
+
+	if (route_free_str != NULL) {
+		pkg_free(route_free_str);
+	}
+
+	after_auto = auto_route->next;
```
david-trihy, line 328: At this point we've identified and classified the auto route and we need to get the next one if exists and set the dst uri to that on line 334

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		return -1;
+	}
+
+	if (parse_headers(msg, HDR_EOH_F, 0) == -1) {
+		LM_ERR("failed to parse route headers\n");
+	}
+
+	request_uri = &msg->parsed_uri;
+
+	if (msg->route == NULL && check_self(&request_uri->host, request_uri->port_no ? request_uri->port_no : SIP_PORT, 0)) {
```
david-trihy, line 380: This is the existing branch which is the full decode of the thinfo param to rebuild the topology, nothing changed outside of compact or not

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+						request_uri->u_val[i].len, request_uri->u_val[i].s);
+					return th_no_dlg_seq_handling(msg, &request_uri->u_val[i], decode_info_buffer, thinfo_decode);
+				} else {
+					LM_DBG("We found legacy param in R-URI with value of %.*s\n",
+						request_uri->u_val[i].len, request_uri->u_val[i].s);
+					return th_no_dlg_seq_handling(msg, &request_uri->u_val[i], decode_info_buffer_legacy, thinfo_decode);
+				}
+			}
+		}
+	} else if (msg->route != NULL && auto_route_on_trusted_socket) {
```
david-trihy, line 397: This branch will do some checking of the route header and received bind address before trying to use the route header which could be an auto route, so this scenario is for a sequential request from your internal UA who has a complete picture of the network topology

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		LM_ERR("cannot parse TO header\n");
+		return;
+	}
+
+	/* do_rr determined by if the request has a tag, don't add them on sequential */
+	is_sequential = get_to(req)->tag_value.len > 0 && get_to(req)->tag_value.s != NULL;
+
+	LM_DBG("Original request trusted send sock '%d' reply is trusted send sock '%d'\n", req_one_way_hiding, one_way_hiding);
+
+	if (!req_one_way_hiding && (lmp = restore_vias_from_req(req, rpl)) == NULL) {
```
david-trihy, line 512: On the reply we care about whether the request or the response has one way hiding applied, so in this branch it's checking if the original request has one way hiding and if not rebuild the Via headers

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		LM_ERR("Failed to restore VIA headers from request \n");
+		return;
+	}
+
+	if (!is_sequential && req_one_way_hiding) {
```
david-trihy, line 517: This branch is for the original offer answer and what is essentially happening is that we are getting an intersection of the set of Record-Route headers so that we can delete a certain amount of them based on the original request and the response.

For example

inbound request has rr1 and rr2 and one way hiding is applied, rr3, rr4, rr5 is now added along the way so the response from the internal user agent has 5 Record-Route nameaddrs and this function essentially finds that intersection, validates it and create a small structure with delete count and encode count.

In the reply that is sent back we ensure that we only send rr1 and rr2 and encode the rest into the Contact header, the auto route is accounted for in the encoding by skipping it if it's there as it is not needed for encoding. 

This is the best case scenario where you just delete a bunch of URIs and encode them, this is a new function I wrote `topo_delete_record_route_uris` it is more explicitly in how it deletes Record-Route headers, in the implementation used for the dialog module and the old code it deleted all headers but this I needed to be able to delete URIs within existing headers hence the new function, it was added to the common logic file but I did not replace the dialog mode deletion with this as I didn't think it was necessary.

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		route_count = th_no_dlg_match_record_route_or_route_uris(req, rpl, HDR_RECORDROUTE_T, auto_route_on_trusted_socket && req_one_way_hiding);
+
+		if (topo_delete_record_route_uris(rpl, route_count.delete_count) < 0) {
+			LM_ERR("Failed to remove '%d' Record-Route URIs\n", route_count.delete_count);
+			goto cleanup;
+		}
+	} else if (!is_sequential && one_way_hiding) {
```
david-trihy, line 524: This is the inverse where the reply is going through a trusted socket and for consistency sake I felt the best thing to do here is to rebuild the Record-Route headers in single URI per header format, the reason why we need to do this is we may need to add in the auto Route header and it needs to be placed in between the original request Record-Routes and the added reply Record-Routes

Original request is rr1, rr2 and reply adds rr3, rr4 what this does is it will place autoroute in between them so your set now is rr1, rr2, autoroute, rr3, rr4

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+            if (rr_lmp == NULL) {
+                LM_ERR("Failed to add Record-Route header\n");
+				if (suffix)
+                	pkg_free(suffix);
+                goto cleanup;
+            }
+		}
+
+		rebuild_req_rrs = req->record_route != NULL;
+    } else {
```
david-trihy, line 568: No hiding one way hiding on either side, so delete and rebuild if the request has headers

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+    int do_rr = 0;
+	str route_s = STR_NULL;
+
+	LM_DBG("Request callback with flags %u\n", flags);
+
+	/* parse all headers to be sure that all RR and Contact hdrs are found */
+	if (parse_headers(req, HDR_EOH_F, 0) >= 0) {
+		one_way_hiding = th_no_dlg_one_way_hiding(req->force_send_socket != NULL ? req->force_send_socket : req->rcv.bind_address);
+        do_rr = get_to(req)->tag_value.len == 0 || get_to(req)->tag_value.s == NULL;
+		if (!one_way_hiding) {
```
david-trihy, line 645: Simple enough branch, if request is not going through a trusted socket, strip the network topology and encode else add the auto route, much easier scenario than the reply.

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+	}
+
+	return -1;
+}
+
+#define HAS_NO_CONTACT_BODY(_m) (((contact_body_t *) ((_m)->contact->parsed))->star == 1 || \
+							    ((contact_body_t *) ((_m)->contact->parsed))->contacts == NULL || \
+                                ((contact_body_t *) ((_m)->contact->parsed))->contacts->next != NULL)
+
+static char* build_encoded_contact_suffix_legacy(struct sip_msg* msg, str rr_set, int *suffix_len, int flags) {
```
david-trihy, line 698: This is a straight port of the older function using the string based encoding, I made some minor changes but nothing notable

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+	if (suffix_plain)
+		pkg_free(suffix_plain);
+	if (rr_set_free_str)
+		pkg_free(rr_set_free_str);
+	if (head)
+		free_rr(&head);
+	return NULL;
+}
+
+static int th_binary_encode_record_route(rr_t *record_route, rr_t **out_rr, int encode_self) {
```
david-trihy, line 893: This is a function to decompose the binary encoding logic which has simple logic, if not double rr just encode as is, if double rr check if the address of the domain is the same then you can dual encode and only encode a small set of differences(protocol, port) this can save space but if it is double rr but if it's for example a scenario where it is public <---> private interface the domain won't match so they're encoded as two separate sockets

Now that I go through this I am going to add a test specifically for that r2=on with domains in each route

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		if (!(lump = insert_subst_lump_after(lump, SUBST_SND_ALL, 0))) {
+			LM_ERR("failed inserting SUBST_SND buf\n");
+		}
+	}
+	if (prefix) pkg_free(prefix);
+	if (suffix) pkg_free(suffix);
+	return -1;
+}
+
+static inline int topo_no_dlg_route(struct sip_msg *msg, str rr_buf[static 1]) {
```
david-trihy, line 1217: This function is so we can classify the Route headers and apply logic to the message when decoding the buffer, if its loose or strict, parses the structures and sets the destination uri, this is when you match topology information from the request domain.

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+	}
+
+cleanup:
+	if (head != NULL)
+		free_rr(&head);
+
+	return route_flags;
+}
+
+static inline int topo_no_dlg_rewrite_contact_as_next_route(struct sip_msg *msg, const str contact_buf[static 1]) {
```
david-trihy, line 1312: This is for strict routing only

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		LM_ERR("failed inserting remote contact route\n");
+		pkg_free(remote_contact);
+		return -1;
+	}
+
+	msg->msg_flags |= FL_HAS_ROUTE_LUMP;
+	return 1;
+}
+
+static decoded_info_buffer_t decode_info_buffer(str *info, const thinfo_options_t *options) {
```
david-trihy, line 1344: logic to call the functions to use the compact decoding, it creates a small structure of pointers to a static buffer

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		goto error;
+	}
+
+	for (i = 0; i < dec_len; i++)
+		decoded_uri_buf.buf[i] ^= options->param_password.s[i % options->param_password.len];
+
+    decoded_uri_buf.len = dec_len;
+    decoded_uri_buf.pos = 0;
+
+    uri_count = thinfo_get_uri_count(&decoded_uri_buf);
```
david-trihy, line 1378: Once validation of buffer sizes for decoding has been done we will get the URI count, as mentioned there's a limit of 12, no particular reason other than 12 routes seems like a lot and that is used as the basis of indexing through the buffer to decode each uri, first being Contact and the rest being Routes and applying them to the SIP message

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+			LM_WARN("non-local socket <%.*s:%d>...ignoring\n", host.len, host.s, port);
+		}
+
+		if (decoded_buffer.sock) {
+			decoded_buffer.state |= HAS_SOCK;
+		}
+    }
+
+	if (decoded_uris[0].s != NULL && decoded_uris[0].len > 0) {
+		decoded_buffer.contact = (str) { .s = decoded_uris[0].s, .len = decoded_uris[0].len };
```
david-trihy, line 1414: I felt directly indexing the buffer from decoding was the most performant way to handle this, they're decoded then the OpenSIPs parsing logic is applied after so they're validated once decoded and it means no copying or dynamic allocation and also a transaction context is applied to essentially invalidate the buffer too, the pseudovars also uses this for the script too and it needs to be set for them to actually be read.

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+		decoded_buffer.routes.len, decoded_buffer.routes.s,decoded_buffer.contact.len, decoded_buffer.contact.s,
+		decoded_buffer.flags, host.len, host.s, port, proto);
+
+	FINALIZE_DECODED_BUF_STATE(decoded_buffer);
+	return decoded_buffer;
+error:
+	return (decoded_info_buffer_t) { .state = INVALID_BUF };
+}
+
+static decoded_info_buffer_t decode_info_buffer_legacy(str *info, const thinfo_options_t *options) {
```
david-trihy, line 1439: The existing decode logic was put into a similar function signature so it can be plugged in as needed, nothing other than the return and signature has changed here, same logic as existing codebase

modules/topology_hiding/th_no_dlg_logic.c
```diff
@@ -0,0 +1,2109 @@
...
+FINALIZE_DECODED_BUF_STATE(decoded_buffer);
+	return decoded_buffer;
+error:
+	return (decoded_info_buffer_t) { .state = INVALID_BUF };
+}
+
+static char user_buf[UINT16_MAX];
+static str ct_user_buf = { .s = user_buf, .len = 0 };
+
+static int th_no_dlg_seq_handling(struct sip_msg *msg, str *info, decode_info_fn decode_fn, const thinfo_options_t *options) {
```
david-trihy, line 1522: This is much the same as the existing sequential handling but can take a function pointer to one of the decoding functions, legacy or compact and it rebuilds the message and then sets next destination etc...

liviuchircu (changes requested):
Hi @davidtrihy-genesys ,

I managed to complete the review, and I'll split the action points into: "Critical" and "Bugs / Quirks", the latter of which can be left to solve/assess during the remainder of the beta phase.  All feedback below refers to the "non-dialog" TH, of course, since the dialog TH is left unchanged.

So if I understand correctly, before we had `;thinfo=` working statelessly, while now it works in a hybrid way.  Because it both stores some transactional-local data via callback param, but also stores data in the `;thinfo=` just as before, only in a compact way.  There is also a 3rd storage (global buffers) for the new `$TH_decoded_routes`, `$TH_decoded_contact` variables, but that seems OK.

Here, just to make sure I understood how it works: you need to store data in the transaction in order to restore Via/RR for the replies.  But you also encode, using the new codec utilities, some minimal amount of data in the `;thinfo=`, which is used for future sequential requests and cross-transaction routing, such as the SIP route set information.  Is this accurate?  If so, perhaps these goals or invariants of the transactional TH support should be clearly stated (code or docs, both can work), in order to help maintainers know what to expect.

## Critical Issues

There is only one "stopper" kind of issue before we can merge:

### Licensing / Ownership

The new `th_common_logic.[ch]` and `th_no_dlg_logic.[ch]` files contain a significant amount of code moved from existing project sources, yet the current header appears to suggest sole Genesys ownership of code that predates this contribution.  Example functions which are near-exact copies of older code:

- `topo_parse_passed_params()`
- `topo_parse_passed_ct_params()`
- `topo_parse_passed_hdr_ct_params()`
- `topo_delete_record_routes()`
- `topo_delete_vias()`
- `delete_existing_contact()`
- `restore_vias_from_req()`
- `topo_ct_param_len()`
- `topo_ct_param_copy()`

... and functions which still clearly retain old implementation chunks or logic:

- `build_encoded_contact_suffix_legacy()`
- `th_no_dlg_encode_contact()`
- `decode_info_buffer_legacy()`
- `th_no_dlg_seq_handling()`

To fix this, I would suggest simply appending the original OpenSIPS Foundation copyright to the existing Genesys copyright, in all of these four files:

```
Copyright (C) 2026 Genesys Cloud Services, Inc.
Copyright (C) 2015 OpenSIPS Foundation
```

## Bugs / Quirks

All of these are results of static analysis, that's why I'm lowering the prio on them.  All of them can be left for post-merge phase, while we test the new code more and more.

### 1. Codec Decoding Vulnerabilities

a) the `thinfo_decode_uris()` seems to read from `p` past the input buffer, as it decodes stuff from the input `;thinfo=` data, for example coming from a SIP reply.  If the reply is malicious or hand-crafted, we could overflow. 

b) the `BUILD_URI_STRING()` seems to write past `decoded_uri_str` capacity.  Needs a double-check.

### 2. Unsafe strchr()

In `th_add_encode_param_password()`, notice how we `pkg_malloc(strlen(param_val))`, then memcpy the same amount, without a terminating `\0`.  Then do a `strchr()` over that buffer, potentially reading out-of-bounds memory.  I believe the reason it hasn't crashed is because heap memory is somewhat zeroed out on startup, which is also when the modparams get parsed.  Still, good to look again into and fully solve.

Overall feedback: great work, really looking forward to reviewing/running the tests now!  Time to have fun 😄

david-trihy:
@liviuchircu 

Thanks so much for the detailed feedback, as for the licensing yeah I think I just want new file == new license so thanks for that, I'll probably just revert the common logic one to OpenSIPs and do that dual mode you said for the no dlg logic thanks.

Regarding your statement around this being a hybrid stateful/stateless operation mode, that actually has not changed, the original implementation was also transactional stateful and relied on the tm state machine during the offer answer to generate the correct Contact header. The main difference in the new one way hiding mode instead of blindly swapping Record-Routes from uac to uas in both request and reply what this does is use that same logic to create an intersection of the Record-Routes between the uac and uas and place our new auto Record-Route in between the intersection point

Eg. rr1,rr2 from internal and rr3 and rr4 from external when the request traverses through the external untrusted socket rr1 and rr2 are stripped from the request and encoded into the Contact and when the uas replies with rr3 and rr4 the transaction is processed and rr1, rr2, auto rr are inserted into the response alongside the existing rr3 and rr4 for a complete Record-Route set as if there was no topology hiding done.

The main changes are around whether the request is traversing through a trusted socket or not and at that point it still uses the topology_hiding to keep the transaction state for response processing so you get an asymmetric aspect to it where the user agent in the trusted environment gets the full network topology without modification except for the auto Record-Route header which is placed into the request or response depending on the originator to allow for sequential routing, because we are allowing the external network topology to be percolated through to the trusted environment we want to be able to route back through to a server that can do the topology match and the auto Route is essentially a replacement for the Contact header which contains all of the information need to do the topology reversing whereas with the auto Route it just contains the flags and the socket info so it can do the topology hiding on the sequential request and pick the correct socket, it behaves exactly the same as if it was using two way hiding, this functionality can be turned off and you can route sequential requests however you want.

There was two reasons I went with this approach first one is the auto Route is implicit in the Contact header sent to the untrusted side so it doesn't have to be encoded into the thinfo thus saving more space in the encoded param, once you get a topology match from an external side and decode everything then you know you're at the right destination and secondly it could potentially be done by generating a new Contact header but then I felt like it was straying from what the intention of this change was which is to leave the external topology unchanged, though adding support for that type of Contact processing could be pretty easy too.

Hope that answer your questions, I will fix the licensing issues and I'll look at the bugs and quirks with my colleague Claude and see if we can address them too and add the docs too, so I'll three extra commits, licensing, bugs quirks fixes and docs.

david-trihy:
@liviuchircu 

I've addressed all the bugs and quirks and show stoppers, new commits are partitioned appropriately for review. Previous comment also expands on questions you asked as well.

liviuchircu:
Merged in https://github.com/OpenSIPS/opensips/commit/01d6f6d9cc6e4907f77405bb9d2da55049d70e07, as I rebased the useful commits against latest `master`, so we clean up the commit history and non-useful commits such as below (for the future, you can avoid them by rebasing all work against latest `master`, coupled with a `push --force` to your fork, any time the branch needs sync'ing):

```
Merge branch 'master' of github.com:OpenSIPS/opensips into topo_hidin...
```

Fantastic work here @davidtrihy-genesys!  Looking forward to seeing the new compact `;thinfo=` RR and Contact params out there in the wilds :fire:

PR #3911, opened 2026-06-07
rtpproxy: drain timeout notification sockets

sobomax:
**Summary**

RTPProxy may keep the notification connection open while writing a burst of timeout events.  The rtpproxy notification callback only performed a single read per reactor wakeup, so already-buffered notifications could be left pending with no further event to wake the process.

**Details**

This issue has been uncovered while working on an extended version of the voiptests that would also test "no-media notifications". As a result of this bug, only first few notifications are processed. The notification queue after that stalls and never recovers effectively making notifications disabled. The bug has been introduced somewhere along 3.2 timeline, when notification has been refactored to use reactor.

**Solution**

Set notification sockets nonblocking and drain them until EAGAIN, preserving partial commands between callbacks.  Also centralize notification connection cleanup so error, EOF and handler failure paths consistently unregister the fd, free pending data and close the socket.

Once committed, the new version of voiptests will be pushed to make sure this won't get broken again.

### DIFF
--- a/modules/rtpproxy/notification_process.c
+++ b/modules/rtpproxy/notification_process.c
@@ -55,10 +55,43 @@ struct rtpp_notify {
 	int fd;
 	char *remaining;
 	int remaining_len;
+	int linked;
 	union sockaddr_union addr;
 	struct list_head list;
 };
 OSIPS_LIST_HEAD(rtpp_notify_fds);
+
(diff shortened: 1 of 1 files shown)


sobomax:
P.S. Tested with both rto.io and external notification socket.

PR #3915, opened 2026-06-10
pua_dialoginfo: fix early state lifetime and clean up dangling early branches

### LINKED ISSUE #3802: [BUG] pua_dialoginfo: TM callback uses FR timer remainder as presence lifetime, causing orphan records
## Summary

The TM callback in `pua_dialoginfo` uses the remaining time on the transaction FR (Final Response) timer as the presence record lifetime. This results in presence records with very short lifetimes (often just 1-30 seconds), which expire before the dialog callback can update them with "confirmed" or "terminated" states. This leaves orphan "early" presence records that cause BLF (Busy Lamp Field) indicators to remain stuck on ringing/busy after calls end.

## Affected Module

`pua_dialoginfo` (modules/pua_dialoginfo/pua_dialoginfo.c)

## Version

Confirmed in OpenSIPS 3.4.16, likely affects all versions with this code pattern.

## Symptoms

- BLF indicators get stuck showing "ringing" or "early" state after calls end
- The problem occurs specifically on calls that receive 180 Ringing responses
- Calls that are answered immediately (no 180 Ringing) work correctly
- The stuck state persists until the presence record expires or is manually cleared

### The Problem Code

In `modules/pua_dialoginfo/pua_dialoginfo.c`, the TM callback `__tm_sendpublish()` calculates the presence record lifetime as:

```c
// Line 269 (for 100 Trying, when publish_on_trying=1)
expire = t->uac[branch].request.fr_timer.time_out - get_ticks();

// Line 295 (for 180 Ringing)
expire = t->uac[branch].request.fr_timer.time_out - get_ticks();
```

This calculates the **remaining time on the FR timer** as the presence lifetime.

### Why This Is Wrong

The TM callback reads `fr_timer.time_out - get_ticks()` to calculate the remaining time on the FR timer. The problem is:

1. **Timer state at callback time**: The callback fires when the 180 Ringing is received, but reads the timer value BEFORE it gets reset to `fr_inv_timeout`. If the 180 arrives late in the `fr_timeout` window, very little time remains.

2. **Mid-registrar complexity**: In mid-registrar configurations, there may be multiple transactions with different timer states. The branch/UAC referenced by the callback may have a different timer than expected.

3. **Observed behavior**: In testing, presence records were created with lifetimes as short as **1 second**, causing them to expire before the dialog callback could update them.

**Example timeline (observed):**
1. T+0.0s: INVITE sent, FR timer starts
2. T+0.1s: 180 Ringing received, TM callback fires
3. T+0.1s: `expire = fr_timer.time_out - get_ticks()` returns ~1 second (unexpectedly low)
4. T+0.1s: "early" presence record created with 1 second lifetime
5. T+1.1s: Record expires
6. T+7.0s: Call answered, dialog callback fires for CONFIRMED
7. T+7.0s: `search_htable()` finds the "early" record but **discards it** (expired 6 seconds ago)
8. T+7.0s: New "confirmed" record created with new etag
9. T+15s: Call ends, "terminated" published for new record only
10. Result: Original "early" state orphaned, BLF stuck

**Note**: The exact reason for the low FR timer remainder is still under investigation. However, the fundamental issue is that **transaction timeout is the wrong metric for presence lifetime**. The dialog callback correctly uses `dlg->lifetime` (typically 3600 seconds), and the TM callback should do the same.

### The Expiry Check in pua

The `search_htable()` function in `modules/pua/hash.c` (lines 203-208) discards expired records:

```c
if (p && p->expires < (int)time(NULL) &&
!(p->expires==0 &&  p->waiting_reply && p->etag.len==0) )
    return NULL;
```

This is correct behavior - expired records should not be returned. The bug is that the records are being created with inappropriately short lifetimes in the first place.

### Contrast with Dialog Callback

The dialog callback `__dialog_sendpublish()` correctly uses `dlg->lifetime` (typically 3600 seconds) for the presence lifetime:

```c
// Line 468 (for CONFIRMED state)
dialog_publish("confirmed", entity, peer,
    &(dlg->callid), branch, 1, dlg->lifetime, ftag, ttag);
```

This inconsistency between TM callback (short FR timer) and dialog callback (long dialog lifetime) is the root cause.

### Before Fix - Record Expires Before CONFIRMED

```
20:18:12 search_htable: found record expires=1768555086 time=1768555092
20:18:12 search_htable: DISCARDING found record due to expiry check!
```

The record expired 6 seconds before the search (1768555092 - 1768555086 = 6).

### After Fix - Record Found Successfully

```
20:36:09 search_htable: found record expires=1768559466 time=1768556169
```

The record expires 3297 seconds in the future (1768559466 - 1768556169 = 3297).

## The Fix

Replace the FR timer calculation with `DEFAULT_CREATED_LIFETIME` (3600 seconds):

```c
// Line 269 - for 100 Trying responses
- expire = t->uac[branch].request.fr_timer.time_out - get_ticks();
+ /* Use a sensible presence lifetime, not the FR timer remainder
+  * which can be very short and cause orphan presence records */
+ expire = DEFAULT_CREATED_LIFETIME;

// Line 295 - for 180 Ringing responses
- expire = t->uac[branch].request.fr_timer.time_out - get_ticks();
+ /* Use a sensible presence lifetime, not the FR timer remainder
+  * which can be very short and cause orphan presence records */
+ expire = DEFAULT_CREATED_LIFETIME;
```

## Why This Code Existed

The original intent was likely to tie the "early" presence state lifetime to the transaction timeout - if the transaction times out without a final response, the presence record would naturally expire. However, this logic is flawed because:

1. The FR timer is for detecting transaction failures, not for presence lifetime
2. When a 180 arrives late in the FR window, the remaining time is too short
3. The dialog callback uses a completely different (and correct) lifetime calculation
4. The presence record needs to survive until the dialog callback can update it

## Test Results

| Configuration | BLF Behavior |
|--------------|--------------|
| Original code (FR timer remainder) | **STUCK** - early state persists |
| With fix (DEFAULT_CREATED_LIFETIME) | **WORKS** - clears properly |
| Reverted to original | **STUCK** - confirms regression |
| Fix re-applied | **WORKS** - confirms fix |

## Environment

- OpenSIPS 3.4.16 (x86_64/linux)
- pua_dialoginfo with `publish_on_trying=0`
- Mid-registrar configuration with dialog module
- BLF monitoring via presence_dialoginfo

### TM Timer Configuration

```
modparam("tm", "fr_timeout", 30)      # 30 seconds before any response
modparam("tm", "fr_inv_timeout", 180) # 180 seconds after provisional response
```

Even with these reasonable values, the presence record lifetime calculation produced values as low as 1 second due to the timing of when the callback reads the timer state.

## Patch

```diff
--- a/modules/pua_dialoginfo/pua_dialoginfo.c
+++ b/modules/pua_dialoginfo/pua_dialoginfo.c
@@ -266,7 +266,9 @@ static void __tm_sendpublish(struct cell *t, int type, struct tmcb_params *_para
 	/* depending on the reply code, see what to publish */
 	if (_params->code<180 && _params->code>=100) {

-		expire = t->uac[branch].request.fr_timer.time_out - get_ticks();
+		/* Use a sensible presence lifetime, not the FR timer remainder
+		 * which can be very short and cause orphan presence records */
+		expire = DEFAULT_CREATED_LIFETIME;
 		if (publish_on_trying) {
 			if (should_publish_A( param->flags, mute_val.s))
 				dialog_publish("trying", entity, peer,
@@ -292,7 +294,9 @@ static void __tm_sendpublish(struct cell *t, int type, struct tmcb_params *_para
 		lock_release(&t->reply_mutex);

 		if (n) {
-			expire = t->uac[branch].request.fr_timer.time_out - get_ticks();
+			/* Use a sensible presence lifetime, not the FR timer remainder
+			 * which can be very short and cause orphan presence records */
+			expire = DEFAULT_CREATED_LIFETIME;
 			if (should_publish_A( param->flags, mute_val.s))
 				dialog_publish(caller_confirmed?"confirmed":"early",
 					entity, peer,
```

## Related

- Original (incorrect) bug report: https://github.com/OpenSIPS/opensips/issues/3801
  - This report suspected etag handling in pua module was the issue
  - Actual root cause is the expires calculation in pua_dialoginfo

NormB:
Fixes #3802, as requested in https://github.com/OpenSIPS/opensips/issues/3802#issuecomment-4667146561.

This is the patch from [the Feb 13 comment](https://github.com/OpenSIPS/opensips/issues/3802#issuecomment-3894683407), tested by @andrewyager in pre-prod/prod against 3.4, rebased onto current master with the following adjustments:

* The failure-handler loop now iterates only the last set of active branches (`t->first_branch` .. `t->nr_of_outgoings`) instead of all `MAX_BRANCHES`, per @bogdan-iancu's review comment.
* Adapted to the new branch bitmask API from f0223cc93 (`BRANCH_BM_TST_IDX`/`BRANCH_BM_SET_IDX` instead of raw `long long` shifts) and to `TM_BRANCH()`.
* The handler now also sets `bitmask_failed` for each branch it terminates, so a late-arriving negative reply on that branch cannot publish a duplicate "terminated".
* The handler honors the `DLG_PUB_A`/`DLG_PUB_B` publishing flags, consistent with the rest of the callback.
* No locking in the failure path: `TMCB_ON_FAILURE` already runs under the transaction's reply lock (via `t_should_relay_response()` -> `run_failure_handlers()`), so the bitmask test/set is safe and re-taking `t->reply_mutex` would deadlock.

### Summary of the fix

1. On 1xx replies, use `DEFAULT_CREATED_LIFETIME` for the published state lifetime instead of `fr_timer.time_out - get_ticks()` — the fr_timer is updated only after `TMCB_RESPONSE_IN` runs, so a provisional reply arriving close to the timer expiration produced a near-zero lifetime and orphaned presence records.
2. Hook on `TMCB_ON_FAILURE` and publish "terminated" for any branch that entered "early" state but never received a final negative reply, so the longer lifetime cannot leave a dangling "early" state when the UAS goes silent after ringing.

### Re-validation on current master

Re-ran the scenarios from the issue against this rebased patch (master @ 4ddb8f0ed, loopback, `fr_timeout=10`, `fr_inv_timeout=15`):

* **Late 180 (8s delay):** early state published with `expires=3600` (previously 7s remainder); 200 OK then publishes confirmed with the dialog lifetime — original bug fixed.
* **Silent UAS (180 then nothing):** on `fr_inv_timeout`, "terminated" with `expires=0` is published for both directions before the failure route runs, and pua deletes the record from its hash table — no dangling early state.
* **Successful call:** no spurious "terminated" from the failure handler.

### DIFF
--- a/modules/pua_dialoginfo/pua_dialoginfo.c
+++ b/modules/pua_dialoginfo/pua_dialoginfo.c
@@ -200,7 +200,44 @@ __tm_sendpublish(struct cell *t, int type, struct tmcb_params *_params)
 	peer = &(param->peer);
 	entity = &(param->entity);
 
-	/* this is triggered only for TMCB_RESPONSE_IN */
+	if (type==TMCB_ON_FAILURE) {
+		/* the transaction completed with a failure - explicitly terminate
+		 * any branches still in "early" state which never received a final
+		 * negative reply (e.g. UAS went silent after sending 180), otherwise
+		 * their published state will linger until it expires */
+		if (get_callid(_params->req, &callid) < 0)
+			return;
+
+		if (include_tags) {
+			if(parse_from_header( _params->req )<0
+			|| parse_to_header( _params->req )<0 ) {
+				LM_ERR("failed to parse the request\n");
+				return;
+			}
+			ftag = &(get_from(_params->req)->tag_value);
+			ttag = &(get_to(_params->req)->tag_value);
+		} else {
+			ftag = ttag = NULL;
+		}
+
+		/* note: this callback runs under the transaction's reply lock,
+		 * so it is safe to test/set the bitmasks here */
+		for (branch=t->first_branch; branch<t->nr_of_outgoings; branch++) {
+			if (!BRANCH_BM_TST_IDX( param->bitmask_early, branch)
+			|| BRANCH_BM_TST_IDX( param->bitmask_failed, branch))
+				continue;
+			BRANCH_BM_SET_IDX( param->bitmask_failed, branch);
+			if (param->flags & DLG_PUB_A)
+				dialog_publish("terminated", entity, peer,
+					&callid, branch, 1, 0, ftag, ttag);
+			if (param->flags & DLG_PUB_B)
+				dialog_publish("terminated", peer, entity,
+					&callid, branch, 0, 0, ttag, ftag);
+		}
+		return;
+	}
+
+	/* from here on, handling of TMCB_RESPONSE_IN */
 	branch = tm_api.get_branch_index();
 
 	LM_DBG("TM event %d [%d/%d] received, entity [%.*s], peer [%.*s],"
@@ -266,7 +303,7 @@ __tm_sendpublish(struct cell *t, int type, struct tmcb_params *_params)
 	/* depending on the reply code, see what to publish */
 	if (_params->code<180 && _params->code>=100) {
 
-		expire = TM_BRANCH(t,branch).request.fr_timer.time_out - get_ticks();
+		expire = DEFAULT_CREATED_LIFETIME;
 		if (publish_on_trying) {
 			if (should_publish_A( param->flags, mute_val.s))
 				dialog_publish("trying", entity, peer,
@@ -292,7 +329,7 @@ __tm_sendpublish(struct cell *t, int type, struct tmcb_params *_params)
 		lock_release(&t->reply_mutex);
 
 		if (n) {
-			expire = TM_BRANCH(t,branch).request.fr_timer.time_out-get_ticks();
+			expire = DEFAULT_CREATED_LIFETIME;
 			if (should_publish_A( param->flags, mute_val.s))
 				dialog_publish(caller_confirmed?"confirmed":"early",
 					entity, peer,
@@ -1015,8 +1052,9 @@ int dialoginfo_set(struct sip_msg* msg, str* flag_s)
 		return -1;
 	}
 
-	/* register TM callback to get access to recevied replies */
-	if (tm_api.register_tmcb( msg, NULL, TMCB_RESPONSE_IN,
+	/* register TM callback to get access to recevied replies and to
+	 * the transaction failure (to clean up dangling early states) */
+	if (tm_api.register_tmcb( msg, NULL, TMCB_RESPONSE_IN|TMCB_ON_FAILURE,
 		__tm_sendpublish, (void*)param_tm, free_cb_param) != 1) {
 		LM_ERR("cannot register TM callback for incoming replies\n");
 		goto end;


bogdan-iancu:
Thank you @NormB

andrewyager:
Thank you!

Issue #3914, opened 2026-06-09
[BUG] leak memory with a repeat async request


vadimd333:
**OpenSIPS version you are running**

```
version: opensips 3.6.6 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, F_PARALLEL_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 26c0c4e33
main.c compiled on 04:03:06 May 21 2026 with cc 8
```

**Describe the bug**

A shmem memory leak has been detected. The issue manifests itself when retrying an async request (async(rest_post(...) ). After the first attempt and receiving a 500 response from the HTTP service, a failover occurs and a second async request is made to the backup HTTP server. For the bug to manifest, it is essential that the original SIP message be modified before the async requests are made. If no modifications are made, the issue does not occur.

**To Reproduce**

To reproduce the issue, you need to send an INVITE call to OpenSIPS. During processing, an async request (async(rest_post(...)) is executed to the HTTP server. This returns a 500 status code, fails, and a second async request is made. The original SIP message must be modified prior to these requests (headers added/removed, field data changed, etc.). Example config code that reproduces the issue:

```
route {
...

    if ($rm=="INVITE") {
        append_hf("X-Test: test\r\n");
        $du="sip:IP:PORT"; #modify to UAS
        $var(i) = 1;
        $var(sock) = "127.0.0.1:9997"; #first address of http-serice
        route(TEST_RESUME_1);
    }
    route(TEST_RELAY);
}


route[TEST_RESUME_1] {
    async(rest_post($var(sock), $ci, , $avp(rest_rsp), , $avp(rest_result)), TEST_RESUME_2);
}


route[TEST_RESUME_2] {
    if ($var(i) < 2) {
        $var(i) += 1;
        $var(sock) = "127.0.0.1:9998"; #second address of http-serice
        route(TEST_RESUME_1);
    }
    route(TEST_RELAY);
}


route[TEST_RELAY] {
    if (!t_relay()) {
        sl_reply_error();
    }
[...]

PR #3916, opened 2026-06-10
Bug async_repeat_request. Update async.c

vadimd333:
https://github.com/OpenSIPS/opensips/issues/3914

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/tm/async.c
+++ b/modules/tm/async.c
@@ -432,7 +432,8 @@ int t_handle_async(struct sip_msg *msg, struct action* a,
 	} else {
 		/* update the cloned UAS (from transaction)
 		 * with data from current msg */
-		if ((t->uas.request) && (route_type==REQUEST_ROUTE))
+		if ((t->uas.request) && (route_type==REQUEST_ROUTE)  &&
+			((msg->msg_flags & FL_TM_FAKE_REQ) == 0))
 			update_cloned_msg_from_msg( t->uas.request, msg);
 	}
 


modules/tm/async.c
```diff
@@ -432,7 +432,8 @@ int t_handle_async(struct sip_msg *msg, struct action* a,
 	} else {
 		/* update the cloned UAS (from transaction)
 		 * with data from current msg */
-		if ((t->uas.request) && (route_type==REQUEST_ROUTE))
+		if ((t->uas.request) && (route_type==REQUEST_ROUTE)  &&
+			((msg->msg_flags & FL_TM_FAKE_REQ) == 0))
```
bogdan-iancu, line 436: @vadimd333 , I'm not questioning the leak (and its fixing), but with this change, the transaction will not be updated from a faked request (like after a async call).
So, if you have this scenario:
* incoming request changed (changeA) in request route
* async jump (a first update will be done here, pushing the changeA into transaction)
* resume in script, doing another set of changes (changeB)
* async jump (no update here the request used in the async resume is a FAKE one)
-> we will loose the changeB, right ?

bogdan-iancu (changes requested):

bogdan-iancu:
This fix is not the correct one, the replacement is here d5d1760ff135dbfcd6f48c15a91d1fd6bdc2c858

PR #3918, opened 2026-06-10
[Bug fix 3711] Close the file descriptor when TLS handshake fails with a TLS server …

GouthamPMadhava:
…successfully listening on the ports.

**Summary**

**Details**

**Solution**

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/proto_tls/proto_tls.c
+++ b/modules/proto_tls/proto_tls.c
@@ -555,6 +555,7 @@ static int proto_tls_send(struct socket_info* send_sock,
 			if (n<0) {
 				LM_ERR("failed async TLS connect\n");
 				rlen = -1;
+				close(fd);
 				goto con_release;
 			}
 			if (n==0) {
@@ -565,6 +566,7 @@ static int proto_tls_send(struct socket_info* send_sock,
 				}
 
 				LM_DBG("Successfully started async SSL connection \n");
+				close(fd);
 				goto con_release;
 			}
 


modules/proto_tls/proto_tls.c
bogdan-iancu, line 629: There is here a `close(fd)` and both reported cases (where you did your extra closes) are landing here - so it shouldn't be needed the close you added .
If so, could you check the value of the `c->proc_id` ? (just add a LM_DBG or let me know how to do it).

GouthamPMadhava: You are right in 3.4 branch, but I do not see the code to close the socket in 3.4.9 as I see the following in there:
`con_release:
	sh_log(c->hist, TCP_SEND2MAIN, "send 1, (%d)", c->refcnt);
	tcp_conn_release(c, (rlen < 0)?0:1);
	return rlen;`
Looks like this is fixed in 3.13 and above. See https://github.com/OpenSIPS/opensips/commit/849bd63f60a0fd390d08d1607f1813b2795b765c

My PR Is based on 3.4 like you suggested as I couldn't figure out a way to fix in 3.4.9.

The scenario is quite simple. Ensure a TLS client tries to create a TLS connection with TLS server where TLS server is listening on the port but the handshake fails (may be due to unknow CA or any other error), and that was creating a leak in 3.4.

bogdan-iancu: There is a single 3.4 branch, holding all all the 3.4.x releases (as tags) - the latest is 3.4.18 minor release. See https://www.opensips.org/About/Version-3-4-x. The fixes are pushed only to the latest minor of the branch. So 3.4.9 was obsoleted by following 3.4.x, the latest beibg 3.4.18 where the fix is present - get the latest 3.4 git head or the 3.418 binaries and you will have the fix.

Issue #3919, opened 2026-06-12
[BUG] late in conversation PRACK not handled by topology_hiding


ovidiusas:
**OpenSIPS version you are running**
```
version: opensips 4.0.0-rc1 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, F_PARALLEL_MALLOC, DBG_MALLOC, CC_O0, USE_POSIX_SEM
MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: 43c43ae42
main.c compiled on 13:53:53 Jun 12 2026 with cc 11
```

**Describe the bug**
If 100rel is enabled on a call and 200ok is received right after a 18x provisional reply, the PRACK may arrive late.
Example:
```
|--INVITE-->|           |
|           |--INVITE-->|
|           |<---180----|
|<---180----|           |
|           |<---200----|
|<---200----|           |
|--PRACK--->|           |
|           |--+PRACK
|           |<-+
|--ACK----->|           |
|           |--ACK----->|
```
The PRACK is kind of handled: the call-id of the sent PRACK is correct, but the RURI is not (it is left untouched) and this makes the PRACK loop back. Also, the via headers are left untouched.

Because PRACK is not routed back to the caller, the call will eventually drop.

bogdan-iancu:
Fixed by #3920

PR #3920, opened 2026-06-12
dialog: handle late PRACK received in conversation

ovidiusas:
**Summary**
Handle late PRACK received in conversation while using topology_hiding module.

**Details**
Related to #3919

**Solution**
Handle PRACK in conversation.

### DIFF
--- a/modules/dialog/dlg_handlers.c
+++ b/modules/dialog/dlg_handlers.c
@@ -2311,7 +2311,7 @@ after_unlock5:
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


bogdan-iancu:
thank you @ovidiusas , reviewing the RFC 3262 (with help of codex) , the situation seams legal - the UAC may send the 200 OK before receiving the PRACK for a prev 18x. So, in this case, the PRACK will be like a regular sequential request, after dialog being established.

PR #3921, opened 2026-06-15
regex: new pcre.subst transformation (PCRE2 based)

ovidiusas:
**Summary**
Add a new transformation, similar to the core `re.subst`, but based on PCRE2.

**Details**
Implemented in the regex module (which is already PCRE2 compatible).
The new transformation is called `pcre.subst` and follows the existing syntax for the `re.subst` core transformation.


bogdan-iancu (changes requested):
Could you update the PR and push the documentation to the modules/regex/README.md file ? the XML is to be dropped in the next days.

bogdan-iancu:
Thank you @ovidiusas , merging now !

PR #3948, opened 2026-06-16
ci: make rtp.io pipeline more robust

sobomax:
**Summary**

Make rtp.io pipeline more robust

**Details**

Current pipeline runs both pull and test in one step, which may fail if the registry is overloaded and pushes test beyond 3-minute timeout.

**Solution**

Split pull into a separate step so it does not contribute to the timeout.

### DIFF
--- a/.github/workflows/.rtp.io.yml
+++ b/.github/workflows/.rtp.io.yml
@@ -23,11 +23,18 @@ on:
         default: debian_12-slim
 
 jobs:
+  get_container_platforms:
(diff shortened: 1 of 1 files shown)


sobomax:
Here is run where this issue happened:

https://github.com/OpenSIPS/opensips/actions/runs/27617919126/job/81708323458

bogdan-iancu:
Does this require any backporting @sobomax  ?

PR #3983, opened 2026-06-26
tls_mgm: match SNI / SIP domain filters case-insensitively

### LINKED ISSUE #3982: [BUG] tls_mgm SNI / SIP DOMAIN FILTER matching is case-sensitive (violates RFC 6066)

```
version: opensips 4.1.0-dev
```
(Present on current `master`, git revision `749dfbb9d4`.)

**Describe the bug**

When `tls_mgm` selects a server TLS domain based on the SNI (Server Name
Indication) hostname from the TLS ClientHello, it matches that hostname against
the configured `SIP DOMAIN FILTERS` (the `match_sip_domain` column / domain
filter) using a **case-sensitive** comparison. RFC 6066 §3 states the
`server_name` is a DNS hostname, and DNS hostname comparison is required to be
case-insensitive (RFC 4343). As a result, a client that sends SNI `Example.COM`
will fail to match a filter configured as `example.com` (and vice versa), even
though they refer to the same name.

This makes it effectively impossible to reliably match TLS certificates with
SIP DOMAIN FILTERS whenever the SNI casing presented by the client differs from
the configured filter casing.

**Root cause**

The runtime match is done in `tls_find_domain_by_filters()`:

`modules/tls_mgm/tls_domain.c:406`
```c
if (!fnmatch(dom_array->arr[i].hostname->s.s, fnm_s, 0)) {
```

`fnmatch()` is case-sensitive by default; the `FNM_CASEFOLD` flag (which would
make it case-insensitive) is not passed. The SNI hostname is taken verbatim
from the ClientHello with no case normalization:

- `modules/tls_openssl/openssl_config.c:297` — `SSL_get_servername(...)`
- `modules/tls_mgm/tls_mgm.c:563-572` — forwarded unchanged to
  `tls_find_domain_by_filters(..., DOM_FLAG_SRV)`

Neither side of the comparison is normalized.

For reference, other OpenSIPS modules already perform this kind of hostname
match case-insensitively, establishing in-tree precedent:
- `modules/identity/identity.c:1877` — `fnmatch(certHostname, fromHostname, FNM_CASEFOLD)`
- `modules/sipmsgops/sipmsgops.c:435` — `fnmatch(pat->s, hf->name.s, FNM_CASEFOLD)`

(There is a secondary case-sensitive `str_strcmp` at
`modules/tls_mgm/tls_domain.c:802`, but that is only used to de-duplicate
filter entries per address at config-load time, not for the runtime match.)

**To Reproduce**

1. Configure a server TLS domain in `tls_mgm` with a SIP DOMAIN FILTER of
   `example.com` (e.g. `match_sip_domain`), bound to a distinct certificate.
2. Start OpenSIPS listening on a TLS socket.
3. Connect with a matching-but-differently-cased SNI:
   ```
   openssl s_client -connect <ip>:5061 -servername Example.COM
   ```
4. Observe that the domain is **not** matched — the connection falls back /
   fails to select the intended domain. The log shows:
   ```
   No domain found matching host: Example.COM in servername extension
   ```
5. Repeating with `-servername example.com` matches correctly, confirming the
   only difference is letter case.

**Expected behavior**

SNI hostname matching against SIP DOMAIN FILTERS should be case-insensitive, per
RFC 6066 / RFC 4343. `Example.COM`, `example.com`, and `EXAMPLE.com` should all
match a filter of `example.com`.

**Relevant System Logs**

```
INFO:tls_mgm:tls_sni_cb: No domain found matching host: Example.COM in servername extension
```

 - Operating System: Ubuntu 24.04
 - OpenSIPS installation: git (master)
 - other relevant information: reproducible with both `tls_openssl` and
   `tls_wolfssl` backends (shared matching code in `tls_mgm`)

**Additional context**

A fix is suggested in #3983 :
it passes `FNM_CASEFOLD` to the `fnmatch()` call
at `modules/tls_mgm/tls_domain.c:406`, matching the existing usage in the
`identity` and `sipmsgops` modules (`_GNU_SOURCE` is defined before
`<fnmatch.h>` so the flag is exposed on glibc). The `match_sip_domain`
documentation is updated to note the case-insensitive behavior.

The config-load de-dup comparison at line 802 was left case-sensitive on
purpose: making it case-insensitive would silently merge filter entries that
look distinct in the config (e.g. `foo.com` and `FOO.com`), which is a separate
behavioral decision.

kvwake:
**Summary**

  Make the matching of the TLS SNI (Server Name Indication) hostname against the configured `SIP DOMAIN FILTERS` case-insensitive, as required by RFC 6066 for DNS hostnames.

  **Details**

  This is a bug fix.

  When `tls_mgm` selects a server TLS domain based on the SNI hostname from the TLS ClientHello, it matches that hostname against the configured `match_sip_domain` filters in `tls_find_domain_by_filters()`
  (`modules/tls_mgm/tls_domain.c`). The match was done with:

  ```c
  fnmatch(dom_array->arr[i].hostname->s.s, fnm_s, 0)
  ```

  `fnmatch()` is case-sensitive unless `FNM_CASEFOLD` is passed, and the SNI hostname is taken verbatim from the ClientHello (via `SSL_get_servername()` in `tls_openssl`, forwarded unchanged through `tls_sni_cb()`), with no
  case normalization on either side.

  RFC 6066 §3 defines `server_name` as a DNS hostname, and DNS hostname comparison is case-insensitive (RFC 4343). The consequence of the current behavior is that a client presenting SNI `Example.COM` fails to match a
  filter configured as `example.com`, making it effectively impossible to reliably match TLS certificates with SIP domain filters whenever the casing differs. This affects both the `tls_openssl` and `tls_wolfssl` backends,
  since the matching code is shared in `tls_mgm`.

  **Solution**

  Pass `FNM_CASEFOLD` to the `fnmatch()` call so the SNI hostname is matched case-insensitively. This matches the existing in-tree precedent in the `identity` (`identity.c:1877`) and `sipmsgops` (`sipmsgops.c:435`) modules,
  which already use `FNM_CASEFOLD` for hostname/header matching.

  `FNM_CASEFOLD` is a GNU extension only exposed by glibc when `_GNU_SOURCE` is defined, so `_GNU_SOURCE` is defined at the top of `tls_domain.c` before any include (this is more correct than `identity.c`, which defines it
  after `<fnmatch.h>`). On BSD/macOS the flag is available by default.

  The `match_sip_domain` documentation is updated to note the case-insensitive behavior.

  I deliberately left the config-load de-duplication comparison (`str_strcmp` at `tls_domain.c:802`) case-sensitive: making it case-insensitive would silently merge filter entries that look distinct in the configuration
  (e.g. `foo.com` and `FOO.com`), which is a separate behavioral decision and not required to fix this bug.

  Verified by building the full tree (all modules except the un-vendored `tls_wolfssl` submodule) under `-Werror` with gcc 13 on Ubuntu 24.04, and running the unit test suite — both pass. A standalone check confirms
  `fnmatch("example.com", "Example.COM", FNM_CASEFOLD)` matches while the old flag `0` does not.

  **Compatibility**

  No configuration migration is required. The change only makes matching more permissive (case-insensitive), so all filters that matched before continue to match; additionally, SNI hostnames that differ only in case now
  match as intended. No API or database schema changes.

  **Closing issues**

  closes #3982

### DIFF
--- a/modules/tls_mgm/doc/tls_mgm_admin.xml
+++ b/modules/tls_mgm/doc/tls_mgm_admin.xml
@@ -986,6 +986,8 @@ modparam("tls_mgm", "match_ip_address", "[dom1]10.0.0.10:5061, 10.0.0.11:5061")
 			The SIP domains used to match a TLS connection with a
 			virtual TLS domain. For TLS server domains, these values will be
 			matched against the hostname provided in the TLS Servername extension(SNI).
+			As the SNI hostname is a DNS name, the matching is case-insensitive
+			(as per RFC 6066).
 			For TLS client domains, the values will be compared with the value of
 			the <xref linkend="param_client_sip_domain_avp"/> AVP.
 			</para>
--- a/modules/tls_mgm/tls_domain.c
+++ b/modules/tls_mgm/tls_domain.c
@@ -33,6 +33,11 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
  */
 
+/* needed to expose FNM_CASEFOLD from <fnmatch.h> on glibc */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include "../../mem/mem.h"
 #include "../../lib/csv.h"
 #include "tls_domain.h"
@@ -403,7 +408,9 @@ tls_find_domain_by_filters(struct ip_addr *ip, unsigned short port,
 	for (i = 0; i < dom_array->size; i++) {
 		memcpy(fnm_s, domain_filter->s, domain_filter->len);
 		fnm_s[domain_filter->len] = 0;
-		if (!fnmatch(dom_array->arr[i].hostname->s.s, fnm_s, 0)) {
+		/* SNI hostnames are DNS names, so match case-insensitively
+		 * (RFC 6066 / RFC 4343) */
+		if (!fnmatch(dom_array->arr[i].hostname->s.s, fnm_s, FNM_CASEFOLD)) {
 			ref_tls_dom(dom_array->arr[i].dom_link);
 			if (dom_lock)
 				lock_stop_read(dom_lock);


bogdan-iancu:
Thank you @kvwake for the report and fix, I will do the backports to 4.0 and 3.6 too

PR #4070, opened 2026-07-10
Feature/clusterer controller, clusterer: add clusterer_controller module for zero-config HA clustering

Lt-Flash:
**Summary**

Add a new `clusterer_controller` module that provides automatic peer discovery and topology management for the `clusterer` module via AES-256-GCM encrypted UDP multicast, eliminating the need for static node configuration or a database.

Also extends the `clusterer` module with a `use_controller=1` modparam that allows it to receive all topology information dynamically at runtime.

---

**Problem**

Currently, clusterer requires either a database table or static `my_node_info`/`neighbor_node_info` modparams listing all cluster nodes. This makes zero-touch deployment impossible — every node must know the addresses of all other nodes in advance, and adding or removing a node requires a config change and restart on all remaining nodes.

---

**Solution**

The `clusterer_controller` module:
- Discovers peers automatically via UDP multicast (configurable group IP and port)
- Allocates unique `node_id` values via master election (highest IP wins, deterministic across all NTP-synchronised nodes)
- Discovers local BIN socket addresses automatically from the `proto_bin` listener list — no manual configuration needed
- Advertises BIN socket information to all peers via encrypted multicast
- Calls the clusterer internal API (`set_my_identity`, `add_node`, `remove_node`, `update_identity`) to drive topology at runtime
- Handles node join, graceful departure (GOODBYE packet) and timeout-based removal
- Supports multiple simultaneous clusters, each with its own independent worker process, peer table and node_id space

---

**Module Parameters**

`cluster` (repeatable) — defines a cluster to join:
```
    "id=1,multicast=239.0.90.1:3333[,password=X][,bin_socket=bin:IP:PORT]")
```

`my_ip` — explicitly set the controller's own identity IP for multicast source and master election. Independent of the BIN socket IP used by clusterer. Three resolution modes are supported:
- **Mode 1** (`my_ip` set): use given IP directly, resolve interface automatically
- **Mode 2** (`interface` set, `my_ip` not set): use first IPv4 on named interface
- **Mode 3** (neither set, default): auto-detect via `getsockname()` probe to multicast group

`interface` — explicitly set the network interface for multicast traffic. The controller derives its identity IP from the first IPv4 address on this interface. Ignored if `my_ip` is also set.

`query_time` — ALIVE heartbeat interval in seconds (default: 5). Controls election window (3×) and purge window (6×).

`password` — global AES-256 encryption password, overridable per cluster.

**Note:** `my_ip` and `interface` control the controller's own identity only. The BIN socket addresses advertised to clusterer peers are discovered separately from the `proto_bin` listener list and are independent of these settings.

---

**Multiple Clusters**

A single instance can participate in multiple clusters simultaneously by repeating the `cluster` modparam. Cluster group identity is defined by the combination of multicast IP **and** UDP port — two clusters may share the same multicast IP but use different ports to remain fully isolated. The `password` per cluster provides an additional AES-256-GCM encryption boundary.

When multiple clusters are defined and the node has more than one BIN socket, `bin_socket=` must be specified per cluster to indicate which BIN socket to advertise for that cluster.

```
# Same multicast IP, different ports
    "id=1,multicast=239.0.90.1:3333,password=Secret1")
    "id=2,multicast=239.0.90.1:3334,password=Secret2")

# Different networks, explicit BIN sockets
    "id=2,multicast=239.0.90.2:3333,bin_socket=bin:10.0.2.10:5566")
```

---

**Changes to clusterer module**

- `use_controller=1` modparam: skip DB/static load, receive topology from controller at runtime
- `current_id` moved to shared memory so all forked processes see identity updates written by the controller worker process
- `GET_CURRENT_ID` macro replacing process-local reads throughout topology, capability and sync code paths
- NULL guards for `current_node` during the startup window before controller identity is established
- Auto-creation of cluster stub in `cl_register_cap()` when `use_controller=1`, so consumer modules (tm, dialog, dispatcher, usrloc) work regardless of `loadmodule` order
- `dep_export_t` additions in `clusterer_controller` enforce that `proto_bin` and `clusterer` initialise first; `dep_export_t` additions in tm, dialog and dispatcher provide a soft dependency on `clusterer_controller` so the dependency system can enforce correct order without requiring a specific `loadmodule` sequence in the config file
- New `clusterer_ctrl.c` / `clusterer_ctrl.h`: internal API — `set_my_identity()`, `add_node()`, `remove_node()`, `update_identity()`, `sync_current_id()`
- Seed flag set on all controller-managed nodes so capability sync donors are always available
- NULL guards added in `sync.c`, `topology.c`, `clusterer.c` and `sharing_tags.c` for `current_node` and `cluster_list` pointers that may be NULL during the controller startup window

---

**Wire Protocol**

UDP multicast, AES-256-GCM encrypted with per-packet 12-byte random nonce and timestamp-based replay protection (±30s window). Packet types: ALIVE, JOIN_REQ (carries BIN socket list), MEMBER_LIST, NODE_ASSIGN, GOODBYE.

---

**Minimal config per node**

```
# Only the BIN socket IP needs to differ per node
socket=bin:10.22.23.191:3857

loadmodule "proto_bin.so"
loadmodule "clusterer.so"
modparam("clusterer_controller", "cluster", "id=1,multicast=239.0.90.1:3333")
loadmodule "dialog.so"
modparam("dialog", "dialog_replication_cluster", 1)
loadmodule "dispatcher.so"
modparam("dispatcher", "cluster_id", 1)
```

The same config file is used on every node — no IPs, node IDs or BIN URLs hardcoded anywhere.

---

**Limitations (alpha)**

- IPv4 only
- Wildcard BIN sockets (`bin:*:PORT`) are rejected at startup — explicit IP required
- Node IDs are not persistent across a full cluster restart (all nodes simultaneously down); rolling restarts are fully supported
- Requires NTP synchronisation between nodes (±30s clock skew tolerance for replay protection)
- Multicast routing required for nodes on different L3 segments

---

**Testing**

- 2 and 3 node clusters on OpenSIPS 3.4 (commit 98f5fb4749)
- dialog, tm and dispatcher replication verified Ok on all nodes
- Capability sync (donor/receiver) verified between nodes
- Node join, graceful departure (GOODBYE) and timeout-based removal verified
- `loadmodule` order independence verified — any order in opensips.cfg produces correct initialisation
- Multiple clusters on a single instance verified
- sharing_tag failover verified

---

**Backward compatibility**

No impact. `use_controller=1` is fully opt-in. Existing clusterer DB mode and static `my_node_info`/`neighbor_node_info` config mode are completely unchanged.

---

**Documentation**

Module admin guide included in `modules/clusterer_controller/doc/` (DocBook XML + PDF) covering all modparams, MI commands (`cc_list_members`, `cc_node_info`), identity resolution modes, multiple cluster configuration, protocol description and known limitations.


Lt-Flash:
Incorrectly arranged PR, sorry

PR #4114, opened 2026-07-17
topology_hiding: keep the dialog-less state in a shared store (fixes long Contact / SUBSCRIBE)

Lt-Flash:
## Problem

When topology hiding runs **without a dialog**, the whole encoded state (route set, original Contact, flags, bind address) rides in the Contact URI parameter — making it long. Some UAs (e.g. Cisco CP-88xx) truncate it, after which `topology_hiding_match()` can't decode it and every sequential request of that call is lost. This is unavoidable for **non-INVITE dialogs like SUBSCRIBE**: `dlg_create_dialog()` is INVITE-only, so a SUBSCRIBE can never get a dialog to hold its state server-side, and the payload is dominated by the route set.

## What this adds

An opt-in mode for the dialog-less path: when **`th_state_url`** points at a `cachedb` backend, the state is stored there under a short key and only the key travels in the Contact:

```
stateless:  ;thinfo=VG9JbzAdIFsyPz0MNS9yRmdPY31kQWFVd0djUGtSVF9CbzYQIFtiYWNBYEFzWmFbZmZiWQ--   (72 chars)
stateful:   ;thinfo=_9ea80d5443fe5c5a                                                           (17 chars)
```

The Contact stays short regardless of topology size, making SUBSCRIBE (and long route sets) workable on constrained endpoints. When `th_state_url` is unset, the encoding path is **byte-for-byte unchanged**. A call that *does* get a dialog never reaches this mode — the dialog holds its state and puts only the dialog id in the Contact, so `force_dialog=1` is the better answer for calls; this mode serves what a dialog cannot.

The full detail — the marker-safety proof, the notifier-granted-expires handling, the measured TTLs, and the per-backend matrix — is in the module docs (admin.xml / README). This is the summary.

## Design

- **In-Contact path unchanged.** A state that still rides in a Contact is XOR-obfuscated + word-encoded as before; a stored state skips both (pointless off the wire) and is kept as plain length-prefixed text. Decode tells them apart by a marker on the key; the restore logic is shared, not duplicated.
- **The state is immutable** (route set / contact / socket are fixed at call creation), so a plain KV store with a TTL is the right fit — no update or conflict path.
- **One key per dialog leg, not per request.** The key is derived (not random) from the leg it belongs to — the call-id + that party's own tag, keyed with the contact-encoding password — so every re-encode overwrites the **same** key. A live dialog holds exactly two states (one per leg) however often it refreshes; the password keeps the key unguessable from the clear-text call-id/tags. A message that cannot be keyed — no call-id, or no tag, as with the reply of an out-of-dialog transaction (a 200 OK to an OPTIONS keepalive carries a Contact but no To tag) — is simply not stored: its state stays inline in the Contact, and it has no sequential traffic to be matched against anyway.
- **Coalesced writes.** Since the state is immutable, a re-store only bumps the TTL. Each node keeps a small shm record of the expiry it last wrote per key and skips the backend write while the state still has enough life left — collapsing a busy dialog's identical writes into occasional refreshes. A write that would *shorten* a state (teardown) is never skipped.
- **Availability.** A sequential request may land on any node, so the backend must be shared (e.g. `cachedb_redis`); `cachedb_local` is fine for a single instance. The storage sits behind a small `th_store.h` interface (backend enum), so a clusterer-replication backend can be added later without touching the encode/decode sites.

## State lifetime

A state is stored every time a Contact is encoded, but the deterministic key overwrites in place (no pile-up), and each write takes its TTL **from the message**, not a flat value:

| Message | State lifetime |
|---|---|
| `SUBSCRIBE` / `NOTIFY` | subscription expires + 30 s (NOTIFY's granted `Subscription-State` wins, else Contact `;expires`, else `Expires`) |
| expires `0` / `Subscription-State: terminated` | matched state dropped now; the outgoing Contact gets `th_state_ttl_short` (routes the final NOTIFY) |
| `OPTIONS` / `MESSAGE` / `INFO` / `PUBLISH` | `th_state_ttl_short` (default 60) — transaction-only, no target refresh |
| `INVITE` / `UPDATE` with session timers | `Session-Expires` + 30 s (the `Expires` ring timeout is deliberately not used) |
| `INVITE` / `UPDATE` without session timers | **not stored** — stays inline in the Contact (a call bounded by nothing must not be able to expire) |
| any message that cannot be keyed (no call-id/tag, e.g. an OPTIONS 200 OK with no To tag) | **not stored** — stays inline in the Contact |
| anything else | `th_state_ttl` (default 3600) |

A message that carries its dialog's lifetime also **refreshes** the state it was sent to, in place — so a long-lived subscription that only ever re-SUBSCRIBEs doesn't lose the state it keeps using. An in-dialog OPTIONS does *not* refresh (its short TTL would otherwise cut a call to a minute). All of this is measured end-to-end against Redis; the full TTL / refresh / notifier-granted tables and the `_`-marker safety proof are in the docs.

## Security

The stored state is **in the clear** — not encrypted, and (unlike the in-Contact copy) not obfuscated — so anyone who can read the backend sees the topology this hides (real Contact, route set, receiving socket). This is intentional: off the wire the obfuscation/encoding serve no purpose, and plain text is inspectable. The store is expected to be a **trusted, internal backend** reachable only by the OpenSIPS nodes that share it — deploy it on a private / DMZ network, bind it to the internal interface, enable the backend's own auth, and do not point `th_state_url` at anything reachable beyond that trust boundary.

## Backend compatibility

Any `cachedb` backend that honours the `expires` argument of `set()` (and provides `remove()`) works — verified against `cachedb_redis` and `cachedb_local`, and applies to memcached / cassandra / couchbase / dynamodb / sql.

**Recommended backend.** For a single node, `cachedb_perf` (#4118) is the backend to prefer, and it exists because of this feature: the measurements below showed `cachedb_local`'s fixed-size hash table degrading as the live-state count grows (finding 3), which is precisely what that module was written to remove. It keeps lookups flat as the table fills, and at 100 000 held calls a `cachedb_perf`-backed `th_state_url` costs about as much as doing no topology hiding at all. Use a shared backend (`cachedb_redis`) when sequential requests can land on a different node than the initial one; use `cachedb_perf` when they cannot, or in front of a shared store. **`cachedb_mongodb` silently ignores the TTL** (states would never expire), so it is **rejected at startup** by a URL-scheme check (the cachedb interface has no "supports expiration" capability — a general gap left out of this PR).

## Configuration

```opensips
loadmodule "cachedb_redis.so"
loadmodule "topology_hiding.so"
modparam("topology_hiding", "th_state_url", "redis://127.0.0.1:6379/0")  # shared, TTL-expiring backend
modparam("topology_hiding", "th_state_ttl", 3600)        # calls with no bound + fallback
modparam("topology_hiding", "th_state_ttl_short", 60)    # OPTIONS/MESSAGE/INFO/PUBLISH

route {
    if (has_totag()) {
        if (!topology_hiding_match())
            xlog("L_WARN", "no topology hiding state for $ru\n");
    } else {
        record_route();
        topology_hiding();   # works for SUBSCRIBE, no dialog needed
    }
}
```

`cachedb_local://` (single instance) and a password-protected `redis://:pass@host/db` cluster are also supported. On a single node, prefer `cachedb_perf` (#4118), which was built for this workload:

```opensips
loadmodule "cachedb_perf.so"
modparam("cachedb_perf", "cache_collections", "th=16")
modparam("topology_hiding", "th_state_url", "perf:///th")
```

## Testing

Against `cachedb_redis` (and `cachedb_local`, confirming backend-agnostic):

- A dialog-less SUBSCRIBE gets its state stored; the Contact param shrinks **72 → 17 chars** (a 16-hex key plus its `_` marker); a sequential request carrying only the key is matched, decoded, and the R-URI restored to the original endpoint.
- **Stateless mode unchanged** — same 72-char Contact, nothing written.
- Per-message TTLs measured correct (subscription expires, session timers, short-TTL for OPTIONS, teardown removal, notifier-granted correction); the `mongodb://` guard refuses startup.

## Pending follow-up — an allocation-free read (waits on #4118)

The sequential path fetches the hidden state through the cachedb `get()`, which returns a `pkg` buffer the caller must free — a `pkg_malloc` + `pkg_free` on **every in-dialog request**. #4118 adds an optional `get_buf()` cachedb endpoint that reads into a buffer the caller already owns (no allocation, one copy instead of two); against `cachedb_perf` it takes the lookup from **397.7 to 290.1 ns/op** at the median (~27%).

get vs get_buf

The topology_hiding side of that change — `th_store_get_buf()`, which reads into a stack buffer where the backend supports it and falls back to an allocated read otherwise, with ownership tracked explicitly so no error path can free a borrowed address — **is written and tested end-to-end** (both the in-place and the fallback path, with the R-URI correctly restored and no leak or double free). It is **deliberately not in this PR**: it calls `get_buf()`, which does not exist until #4118 lands, so it would not compile here. **It will be pushed to this PR once #4118 is merged.** Nothing in the current PR depends on it — this is purely an optimisation for a later commit.

## Roadmap (not in this PR)

- **Clusterer replication backend** — keep states in shm and replicate to peers (mirroring `dlg_replication.c`), for deployments that don't want an external store in the signalling path; the immutable state makes it simpler (create / read / expire only).
- **A `CACHEDB_CAP_EXPIRES` capability** so TTL-ignoring backends are rejected through the interface rather than by a URL-scheme match (touches the core plus every backend).

## Notes

- New files `th_store.c` / `th_store.h`; docs + regenerated README included. `th_state_url` declares a `cachedb_url` dependency so the backend module loads first.

---

# Measured cost

The design section above says the state backend is a choice (`cachedb_local` for a single instance, a
shared store otherwise). That choice has a concrete price, so here it is measured, together with the
alternatives it competes with.

## Test environment

| | |
|---|---|
| Host | KVM guest, **8 vCPU** |
| CPU | **Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz** (Broadwell), `hypervisor` flag present |
| Memory | **7.8 GB** |
| OS / kernel | Debian 13 (trixie) / 6.12.96-amd64 |
| OpenSIPS | **4.0.0** &mdash; `STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, F_PARALLEL_MALLOC, FAST_LOCK-ADAPTIVE_WAIT` |
| Runtime | `-m 4096` shm, `-M 32` pkg, `udp_workers=8` (= vCPU count), `tcp_workers=4` |
| Generator | SIPp 3.6.0, 4 parallel UAC processes; SIPp UAS pool selected via `dispatcher` (weighted round-robin) |

## What was exercised

A dual-homed proxy: upstream leg and downstream pool on separate interfaces. Every call is a complete
`INVITE -> 100 -> 200 -> ACK -> BYE -> 200` carrying `Session-Expires`, so a **storable** state is produced
on **every** call &mdash; deliberately the heaviest path for this feature. `topology_hiding()` runs on the
initial INVITE and `topology_hiding_match()` on every sequential request; a call is only counted as
successful if its BYE was matched and answered, so the decode path is verified, not just the encode.

Two measurement notes for anyone reproducing this: CPU must be sampled over a **steady-state mid-run
window** (including the call-teardown tail dilutes it badly &mdash; 46% vs the real 67% in one run), and a
single SIPp UAC process saturates one core at roughly 4.8k CPS, so the load has to be generated from
several processes or you measure the generator instead of the proxy.

## UDP &mdash; what the state backend costs

LB CPU vs achieved throughput for five topology-hiding backends

Throughput ceiling and CPU cost at 3000 CPS by backend

| Backend | CPU @3k CPS | CPU @6k CPS | Ceiling |
|---|---|---|---|
| no TH (plain record-route) | 8 % | ~47 % | **~10 000 CPS** |
| TH + dialog (`force_dialog=1`) | 10 % | ~73 % | **~9 000 CPS** |
| TH + `cachedb_local` | 7 % | 77 % | **~7 100 CPS** |
| TH + `cachedb_redis`, loopback | 48 % | 80 % | **~6 900 CPS** |
| TH + `cachedb_redis`, 3-master LAN cluster | 54 % | &mdash;* | **~5 300 CPS** |

\* the LAN-cluster case never becomes CPU-bound &mdash; it caps at ~65%, see findings.

### Findings (UDP)

1. **The encode/decode is not what costs &mdash; the backend round-trip is.** Holding the same state in a
   dialog runs at 10% CPU at 3k CPS; putting it through a loopback Redis costs 48% for identical calls,
   and the ceiling falls from ~9k to ~6.9k CPS. The per-call `SET`/`GET` dominates, not the topology
   hiding itself.
2. **`cachedb_local` is the cheapest option at low/medium load but walls early.** At 3k CPS it is the
   lowest of all five (7%), because it avoids both the socket round-trip and dialog bookkeeping. Then CPU
   goes 9% -> 77% -> 100% between 4k and 7.1k CPS: the shared cache hash serialises the workers, so it
   caps ~2k CPS *below* the dialog path despite being cheaper per call. A good single-instance choice up
   to a few thousand CPS; not the one to pick for peak throughput.
3. **A remote store is latency-bound, not CPU-bound.** The LAN cluster caps at ~5.3k CPS while the proxy
   still has ~35% CPU idle. The cachedb call is synchronous, so a worker blocks for the whole round-trip
   and stops draining its socket &mdash; every extra millisecond of backend RTT comes straight off the
   ceiling. Colocate the store where possible; this is also exactly what the coalesced-write path is
   worth the most for.
4. **These are worst-case numbers for this feature.** `Session-Expires` on every call means a state was
   stored for every call. Real traffic has far fewer storable messages, so production cost sits between
   the no-TH and TH lines &mdash; nearer the former the fewer states are actually stored. Note also that
   each call used a distinct key, so write coalescing is *not* what these figures measure.
5. **The load is ~50/50 user/system CPU.** At these rates a UDP proxy is as much kernel/packet-path bound
   as application bound; all workers share one UDP socket, which is why even the no-TH line plateaus
   around 83-97% rather than pinning all 8 cores.
6. **Worker count should track core count.** Raising `udp_workers` 8 -> 32 on the same 8 vCPU made things
   worse: ~1.6x the CPU for equal throughput and 2.5% call loss at only 1000 CPS (shared-socket
   contention plus scheduler churn), for a ~13% ceiling gain.

## TCP &mdash; connection ceiling

Not specific to this PR &mdash; transport characterisation of the same box, included for completeness.
Connections were opened at a steady ~76/sec and held:

TCP connection saturation: linear climb stopping dead at 2048

### Findings (TCP)

1. **Simultaneous TCP connections cap at `tcp_max_connections` (default 2048).** At that point OpenSIPS
   logs `maximum number of connections exceeded` and refuses new connections. The failure is graceful:
   **existing connections and UDP signalling keep working** &mdash; only new TCP is rejected.
2. **A much lower limit bites first for bursts: `tcp_socket_backlog` defaults to 10.** The accept queue
   overflows on rapid arrivals (`listen queue of a socket overflowed` in `netstat -s`) long before 2048 is
   approached; reaching the real ceiling at all required raising it. For a public-facing edge proxy a
   backlog of 10 is easy to overrun with a reconnect storm or a burst of TLS handshakes &mdash; this may be
   worth revisiting as a default.
3. **Kernel SYN handling was not implicated.** No `Possible SYN flooding` messages, and
   `tcp_max_syn_backlog` (512) is far above the 10-deep accept queue; the constraint is purely the
   application's `listen()` backlog. File-descriptor limits were also not a factor (65536 available).

---

# Under concurrent load: 50 000 calls held

The figures above ramp the call rate with calls torn down immediately, which measures the cost of
*establishing* state. This second test measures the cost of *holding* it: the hold time is set to
`50000 / rate`, so once the ramp completes there are ~50 000 calls in flight, and the load is then
sustained. Concurrency is verified rather than assumed - it is the summed in-flight call count reported
by the load generators, and it lands within 0.3% of 50 000 on every valid rung.

**These numbers are not directly comparable to the section above.** That test failed any call whose
response took over 8 s; this one allows 120 s, because calls are held deliberately. The proxy therefore
posts higher call rates here at considerably worse latency. They are two different measurements, not a
revision of the same one.

A rung counts only where the proxy actually kept pace: under 5% failures **and** in-flight calls still
near 50 000. That second condition matters. Judged on failure rate alone the loopback-Redis case appears
to reach 9 829 CPS at 1.7% failures - but its in-flight count had grown to 112 000, meaning calls were
arriving faster than they cleared. That is a growing backlog, not 50 000 calls being held, so it is not
counted.

LB CPU vs throughput while holding 50,000 concurrent calls

Sustained call rate and CPU cost per backend at 50,000 concurrent calls

| Backend | Sustained CPS | at CPU | in-flight | CPU @4 000 CPS | Memory @4 000 CPS |
|---|---|---|---|---|---|
| no TH (plain record-route) | **11 788** | 93 % | 64 k | 8 % | 870 MB |
| TH + dialog (`force_dialog=1`) | **9 875** | 90 % | 54 k | 17 % | 1 203 MB |
| TH + `cachedb_redis`, loopback | **6 000** | 80 % | 56 k | 57 % | 924 MB |
| TH + `cachedb_local`, `th=16` | **5 944** | 89 % | 51 k | 29 % | 1 014 MB |
| TH + `cachedb_local`, default | **5 907** | 97 % | 52 k | 45 % | 913 MB |
| TH + `cachedb_redis`, 3-master LAN cluster | **3 996** | 57 % | 52 k | 57 % | 918 MB |

### Findings (50 000 held calls)

1. **Holding state in a dialog is the strongest option once concurrency is high** — *of the backends
   measured here*. 9 875 CPS against ~6 000 for every cachedb variant, and 17% CPU against 45-57% for
   identical traffic at 4 000 CPS. The ranking is the same as in the first test, but the gap widens
   sharply with in-flight state.
   **This reverses at higher concurrency with a backend that scales.** Re-running the same comparison at
   **100 000** held calls against `cachedb_perf` (#4118) put cache-backed topology hiding *ahead* of the
   dialog: 93% CPU at 4 000 CPS and tracking the no-topology-hiding curve, while dialog-TH saturated by
   3 000 CPS, broke at 4 000 (8.2% failures) and carried ~2.5x the resident memory. The dialog's advantage
   here is real but bounded — it comes from the cachedb backends of the day degrading with live-state
   count, not from the dialog scaling better.
2. **A remote store degrades furthest.** The LAN cluster sustains 3 996 CPS and never exceeds 57-63% CPU
   - roughly a third of the machine stays idle while workers block on the synchronous round-trip. Backend
   RTT, not proxy CPU, sets that ceiling, which is the case the coalesced-write path helps most.
3. **`cachedb_local`'s default collection is undersized for tens of thousands of live states.** The
   default is `1 << HASH_SIZE_DEFAULT` = 512 buckets, so ~50 000 entries means ~100-entry chains walked
   under each bucket lock. Setting `modparam("cachedb_local", "cache_collections", "th=16")` (65 536
   buckets) cut CPU from **45% to 29%** at 4 000 CPS. It barely moved the ceiling (5 907 -> 5 944), so it
   buys headroom below saturation rather than a higher wall - but it is a one-line change and worth
   documenting for anyone using `cachedb_local` to hold TH state at scale. This finding is what led to
   `cachedb_perf` (#4118): a table that resizes itself at runtime, so the cliff cannot form and no
   collection has to be hand-sized in the first place.
4. **Dialog pays for its CPU advantage in memory**: 1 203 MB at 4 000 CPS versus ~915 MB for the cachedb
   variants, growing to 2 304 MB by 10 000 CPS. Holding 50 000 calls costs roughly 350 MB of shared
   memory even with no topology hiding at all, from the in-flight transactions alone.
5. **Failure counts alone are a misleading ceiling metric under sustained concurrency.** Three of the six
   configurations post low failure rates at rungs where in-flight calls had doubled or tripled past the
   target. Any figure quoted from this kind of test should be paired with the in-flight count.

### DIFF
--- a/modules/topology_hiding/README
+++ b/modules/topology_hiding/README
@@ -24,6 +24,9 @@ topology_hiding Module
               1.3.9. th_contact_caller_username_var (string)
               1.3.10. th_contact_callee_username_var (string)
               1.3.11. th_callid_loop_protection (int)
+              1.3.12. th_state_url (string)
+              1.3.13. th_state_ttl (int)
+              1.3.14. th_state_ttl_short (int)
 
         1.4. Exported Functions
 
@@ -63,11 +66,14 @@ topology_hiding Module
    1.9. Set th_contact_caller_username_var parameter
    1.10. Set th_contact_callee_username_var parameter
    1.11. Set th_callid_loop_protection parameter
-   1.12. topology_hiding usage
-   1.13. Calling topology_hiding_match() function for topology
+   1.12. Set th_state_url parameter
+   1.13. Set th_state_ttl parameter
+   1.14. Set th_state_ttl_short parameter
+   1.15. topology_hiding usage
+   1.16. Calling topology_hiding_match() function for topology
           hiding sequential requests
 
-   1.14. topology_hiding_match_dialog() usage
+   1.17. topology_hiding_match_dialog() usage
 
 Chapter 1. Admin Guide
 
@@ -257,6 +263,210 @@ ding_username_var__")
 modparam("topology_hiding", "th_callid_loop_protection", 1)
 ...
 
+1.3.12. th_state_url (string)
+
+   The URL of a cachedb backend where the topology hiding state of
+   dialog-less calls is to be kept.
+
+   When topology hiding runs without a dialog, the whole encoded
+   state normally travels inside the Contact URI parameter (see
+   th_contact_encode_param), which makes that URI long. Some user
+   agents cannot cope with it and truncate the parameter, after
+   which topology_hiding_match() can no longer decode it and the
+   sequential requests of those calls are lost. This is most
+   visible for non-INVITE dialogs, such as SUBSCRIBE, as those can
+   never be given a dialog to hold the state - the dialog module
+   only handles INVITE based dialogs.
+
+   With this parameter set, the state is stored in the given
+   backend under a short key, and only that key travels in the
+   Contact URI, keeping it short regardless of how large the
+   hidden topology is. Off the wire the obfuscation and URI-safe
+   encoding a Contact-borne state carries serve no purpose, so the
+   stored copy holds neither: it is kept as plain, length-prefixed
+   text.
+
+   A state is only ever kept on the server side when the message
+   it belongs to tells how long its dialog is going to live, since
+   a state which expires from under a dialog which is still up
+   cannot be matched anymore, and takes down everything that was
+   still to be routed through it - the BYE of a call, most
+   notably. A subscription is bounded by its expires, and the
+   messages which open no dialog only live as long as their
+   transaction, so those are stored. A call is not bounded by
+   anything, unless the session timers are in use, so the state of
+   a call which has none keeps travelling in its Contact, where it
+   cannot expire - exactly as it does without this parameter. Both
+   may well be in use at the same time, as this is decided for
+   each dialog as its Contact is encoded.
+
+   Note that a call which is given a dialog never gets here at
+   all: the dialog module holds its state and bounds it for
+   exactly as long as the call lives, and its Contact carries just
+   the dialog id, so it is already both short and correct.
+   Engaging a dialog for the calls (see force_dialog) is therefore
+   a better answer for them than this parameter could ever be,
+   which is left to serve what a dialog cannot: everything that is
+   not INVITE based.
+
+   The state must be readable by whichever node receives the
+   sequential requests, so when the traffic is distributed or
+   fails over between several OpenSIPS instances, the backend must
+   be shared between them (e.g. a common cachedb_redis instance or
+   cluster). For a single instance, a local backend such as
+   cachedb_local is enough - it is kept in shared memory, so all
+   the OpenSIPS workers do see it, but its contents are lost on
+   restart and are not visible to the other nodes.
+
+   Security: the state kept in the backend is stored in the clear.
+   It is not encrypted, and - unlike the copy that travels inside
+   a Contact - it is not obfuscated either, so anyone able to read
+   the backend can see the topology it hides: the real Contact,
+   the route set and the receiving socket of each call. This is
+   intentional. The store is meant to be a trusted, internal
+   backend reachable only by the OpenSIPS nodes that share it:
+   deploy it on a private or DMZ network, bind it to the internal
+   interface, enable the backend's own authentication, and keep it
+   off any untrusted network. Do not point th_state_url at a
+   backend that is reachable beyond that trust boundary.
+
+   The state is immutable, so re-encoding a Contact on a
+   sequential request only pushes its expiration further out. To
+   spare the backend a write on every such request, each node
+   keeps a small in-memory record of what it last stored and skips
+   the write while the state still has enough of its lifetime left
+   - a write that would actually shorten a state (a subscription
+   being torn down) is never skipped. This is transparent and
+   needs no configuration.
+
+   A state is only ever removed from the backend when it expires
+   (see th_state_ttl), therefore the backend must implement the
+   expiration of the values it stores. All the cachedb backends
+   do, except for cachedb_mongodb, which silently ignores it -
+   using it here would pile up the states forever, so it is
+   rejected at startup.
+
+   By default this parameter is not set, and the state travels in
+   the Contact URI.
+
+   Example 1.12. Set th_state_url parameter
+...
+# a single instance - state kept in its own shared memory
+loadmodule "cachedb_local.so"
+modparam("topology_hiding", "th_state_url", "local://topohiding")
+
+# several instances sharing a Redis server
+loadmodule "cachedb_redis.so"
+modparam("topology_hiding", "th_state_url", "redis://127.0.0.1:6379/0")
+
+# a password protected Redis cluster - one node is enough,
+# the rest of the cluster is discovered from it
+modparam("topology_hiding", "th_state_url",
+         "redis://:secretpass@10.0.0.20:6379/0")
+
+# a Memcached server
+loadmodule "cachedb_memcached.so"
+modparam("topology_hiding", "th_state_url",
+         "memcached://127.0.0.1:11211/")
+...
+
+1.3.13. th_state_ttl (int)
+
+   For how long, in seconds, a topology hiding state is kept in
+   the backend configured through th_state_url. It should be at
+   least as long as the calls whose topology is hidden are
+   expected to last, as a call whose state has expired can no
+   longer have its sequential requests matched.
+
+   Whenever a message which keeps its dialog alive is matched, the
+   state it was sent to has its expiration pushed further, for as
+   long as that message says the dialog is going to live. This
+   matters because the party which sent it only learns of a new
+   state if the reply it gets back carries a Contact of its own -
+   otherwise it goes on using the state it already knows, which
+   therefore has to outlive its original expiration. Only the
+   messages which do carry the lifetime of their dialog may
+   refresh it - a SUBSCRIBE or NOTIFY through its expires, an
+   INVITE or UPDATE through its Session-Expires - so that, say, an
+   in-dialog OPTIONS cannot cut the state of the call it runs into
+   down to th_state_ttl_short.
+
+   A state is stored each time a Contact is encoded, and is
+   otherwise only removed when it expires, so the module tries not
+   to keep any state for longer than it may be needed:
+     * for SUBSCRIBE and NOTIFY, the state has to outlive the
+       subscription, so its duration is taken from the very
+       message being encoded (plus a small margin, to cover the
+       refresh). A NOTIFY reports how long the subscription
+       actually has left in its Subscription-State header (RFC
+       6665 4.1.3), which is what the notifier granted rather than
+       what the subscriber asked for, so it is preferred. Failing
+       that, it is looked for as an “expires” parameter of the
+       Contact first and as the Expires header afterwards, the
+       parameter taking precedence over the header. This matters
+       because a notifier is free to grant a shorter subscription
+       than the one asked for, and its NOTIFYs are what bring the
+       state of a subscription back in line with what was really
+       granted. Note that a subscription may last much longer than
+       this parameter, in which case its state is kept for longer
+       as well. A message which ends a subscription (an expires of
+       “0”, or a NOTIFY reporting it as “terminated”) drops the
+       state it was sent to right away, instead of leaving it
+       behind for as long as the subscription was going to last -
+       the Contact it gets encoded with still takes
+       th_state_ttl_short, so that the final NOTIFY may be routed
+       back;
+     * for INVITE and UPDATE, when the session timers are in use,
+       the session gets refreshed every Session-Expires seconds
+       (RFC 4028) and each refresh encodes the Contact anew, so
+       the state only has to survive one interval - that value
+       (plus the margin) is used then. Should a refresh not come,
+       the session is torn down anyway, so this can never expire
+       the state of a live call. Note that the Expires header is
+       deliberately not used here: on an INVITE it limits the
+       validity of the invitation, not the duration of the call it
+       may lead to;
+     * OPTIONS, MESSAGE, INFO and PUBLISH neither open a dialog
+       nor refresh the target of one, so their Contact is only of
+       interest while their transaction is running - they get
+       th_state_ttl_short;
+     * everything else, and any of the above whose expiry could
+       not be told (no session timers, no Expires), may last for
+       any amount of time, so it gets this parameter.
+
+   This parameter is only used when th_state_url is set.
+
+   Default value is “3600” (one hour).
+
+   Example 1.13. Set th_state_ttl parameter
+...
+# hide the topology of calls which may last up to 12 hours
+modparam("topology_hiding", "th_state_ttl", 43200)
+...
+
+1.3.14. th_state_ttl_short (int)
+
+   For how long, in seconds, to keep the topology hiding state of
+   the messages which do not open a dialog nor refresh the target
+   of one - see th_state_ttl for the exact list. Their Contact is
+   only used while their transaction is alive, so keeping their
+   state for as long as a call's would needlessly fill up the
+   storage.
+
+   It only needs to cover the lifetime of a transaction. Raise it
+   to the value of th_state_ttl if you have user agents which
+   wrongly latch onto the Contact of such a request and keep using
+   it afterwards.
+
+   This parameter is only used when th_state_url is set.
+
+   Default value is “60”.
+
+   Example 1.14. Set th_state_ttl_short parameter
+...
+modparam("topology_hiding", "th_state_ttl_short", 120)
+...
+
 1.4. Exported Functions
 
 1.4.1.  topology_hiding()
@@ -322,7 +532,7 @@ modparam("topology_hiding", "th_callid_loop_protection", 1)
    is advertised on both legs. If the separator is being used, you
    can control the username put in contact per leg.
 
-   Example 1.12. topology_hiding usage
+   Example 1.15. topology_hiding usage
 ...
 if(!has_totag() && is_method("INVITE")) {
         topology_hiding();
@@ -355,7 +565,7 @@ if(!has_totag() && is_method("INVITE")) {
 }
 ...
 
-   Example 1.13. Calling topology_hiding_match() function for
+   Example 1.16. Calling topology_hiding_match() function for
    topology hiding sequential requests
 ...
 if (has_totag())
@@ -384,7 +594,7 @@ ogy hiding dialog\n");
 
    This function can be used from REQUEST_ROUTE.
 
-   Example 1.14. topology_hiding_match_dialog() usage
+   Example 1.17. topology_hiding_match_dialog() usage
 ...
     if (has_totag()) {
         if (!topology_hiding_match() ) {
--- a/modules/topology_hiding/README
+++ b/modules/topology_hiding/README
@@ -164,6 +164,12 @@ ram;custom_param")
    not already created ). This will only work for INVITE based
    dialogs, and the dialog module must be loaded.
 
+   This only controls whether the module creates a dialog of its
+   own. A dialog created by the script itself, with
+   create_dialog(), has the same effect: once a dialog exists, it
+   carries the topology hiding state. See th_state_url for when
+   that is what you want and when it is not.
+
    Default value is “0”
 
    Example 1.5. Set force_dialog parameter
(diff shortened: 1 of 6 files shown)


Lt-Flash:
Hi, whenever you have a moment, could I kindly ask for a review of this PR? I'd really appreciate it.

I'll hold off on any further commits until the review is complete, so the branch stays stable for you to look at.

Thanks very much for your time!

Lt-Flash:
# Follow-up already implemented: fetching a hidden-dialog state from the node that has it

Building on the `th_state_url` shared store this PR introduces, the next stage is implemented and tested on a development branch: when a sequential request lands on a node whose store does **not** hold the state — a failover, a re-hashed load balancer, an anycast flap — the node can fetch it from the cluster instead of failing the match.

Three pieces, all working:

**An asynchronous match.** `topology_hiding_match()` is now also exported as an async function:

```
async(topology_hiding_match(), resume);
...
route[resume] { if ($rc > 0) t_relay(); else send_reply(404, "Not Here"); }
```

On a local miss the transaction suspends, the cross-node fetch's eventfd goes to the reactor, and the worker returns to the pool; the resume route retries the ordinary match once the answer arrives. Measured end-to-end: on an idle LAN blocking and suspending are identical (a fetch is sub-millisecond), but with a peer 100 ms away, 60 concurrent requests across 4 workers take 1.52 s blocking versus 0.31 s suspended — the difference is pure worker-occupancy arithmetic, and it is why the async form exists.

**The key says where it was put.** The `_`-prefixed state key this PR carries in the Contact is read off the wire by the matching side, never recomputed — which makes its first two characters available as a covert channel. They are rewritten as `(node_id − 1) XOR mask`, the mask drawn from `MD5(encode_pw || rest-of-key)`, so keys stay uniform-looking to an outsider and the token never changes length (rolling upgrades keep working; a decoded id that is not a live cluster member simply means "broadcast instead"). A node that receives a foreign key can therefore ask the owning node **unicast first** rather than broadcasting to everyone.

**Demonstrated, then measured under load.** The functional demo: a sequential request for a state held only by a sibling returns 404 through the synchronous path and 200 through the async one, the state fetched over the cluster's encrypted plane and stored locally so the next request is an ordinary local hit. Under sustained load on a 3-node containerized cluster (30,000 states, ~90,000 reads/s), a node missing two thirds of all states converges to holding everything in ~13 s while serving full traffic, and the fetch costs the triggering request 200 µs–1.7 ms — [full scenarios and charts on the cachedb_perf PR](https://github.com/OpenSIPS/opensips/pull/4118#issuecomment-5104368008).

Two smaller items from the same work that touch this PR's code directly: `th_store_get()` logged an ordinary expired-state miss at ERROR (a cachedb miss returns −2, the code tested `< 0`) — the follow-up distinguishes a clean miss from a backend failure, which quiets a steady stream of misleading ERROR lines in production logs; and a pulled state whose stored send-socket names an interface the fetching node doesn't have falls back to any local socket of the matching protocol — with the operational rule that UA-leg blobs store the anycast address and are therefore portable across nodes.

This lands after the cachedb_perf side merges (the fetch machinery lives there; this module only consumes its API), which is the same ordering note already in the PR body's pending-follow-up section.

Lt-Flash:
# Backend compatibility for `th_state_url` — surveyed, and a docs gap closed

Prompted by a good question: which cachedb backends actually work here, and can a node get a state it doesn't hold from a sibling? I went through every backend in the tree rather than answering from memory. Summary first, then what changed.

`th_store` needs only `init` / `get` / `set` / `remove` / `destroy`, so the real question is whether the backend honours the expiry on `set` — and, separately, what it does when more than one node is involved.

| Backend | Expires values | Multi-node behaviour | Verdict |
|---|---|---|---|
| **cachedb_local** (plain) | yes | node-local, alone | ✅ single instance only |
| **cachedb_local** `…/r` + `cluster_id` | yes | **replicated to the cluster over BIN on every write, full sync on join** | ✅ **works across nodes today** |
| **cachedb_redis** | yes (`EXPIRE`) | one shared store | ✅ |
| **cachedb_memcached** | yes (expiry passed on insert) | one shared store | ✅ |
| **cachedb_cassandra** | yes (`TTL` in the CQL) | one shared store | ✅ |
| **cachedb_couchbase** | yes (`lcb_cmdstore_expiry`) | one shared store | ✅ |
| **cachedb_dynamodb** | yes (passed to the insert) | one shared store | ✅ |
| **cachedb_sql** | yes, but **swept periodically, not enforced on read** | one shared database | ⚠️ works; a state can outlive its TTL by up to one `cache_clean_period` |
| **cachedb_mongodb** | **no** — `mongo_con_set()` never uses its `expires` argument | one shared store | ❌ rejected at startup, as now |

So the honest answer to "can a node fetch a state it doesn't have" is **yes, in two different ways that already exist**, and I had previously under-sold one of them:

- a **shared** backend (redis and friends) — every node reads the same store, so nothing is ever missing; every lookup is a network round trip;
- **cachedb_local with `/r`** — `lcache_htable_insert()` calls `replicate_cache_insert()` for a replicated collection, pushing every write to the cluster over BIN, and `receive_sync_request()` gives a joining node a full copy. Reads stay local and fast; the cost moves to the write side, one send per peer per stored state.

The trade is eager push versus lazy pull, and which wins depends entirely on the miss rate. Replication pays on every write regardless of whether anyone ever asks; it is the better choice when misses are common or immediate consistency matters. That is worth knowing when reading the cross-node fetch work described in the earlier comment, which pays only on a miss — on a production cluster with call-id affinity we measured a cross-node miss rate of 0.0000%, where eager replication would be pure overhead, but the ordering reverses as soon as the balancer stops being sticky.

## What changed

The parameter's documentation explained what the store is *for* and noted the mongodb exception, but never answered the question an operator actually has — what happens when a request lands on the wrong node. It now has a **Choosing a backend** section covering the three shapes with their costs, and an example showing the replicated-local form beside the shared one, because that form is easy to miss: it is two characters on the end of a collection name plus a modparam on a different module.

The expiration note is also sharpened for `cachedb_sql` — it does expire, but by a periodic sweep, so a state can outlive its lifetime by up to one cleanup interval. Worth setting that module's `cache_clean_period` deliberately if you use it here.

No functional change in this PR — documentation only.

Lt-Flash:
Hi @bogdan-iancu, hoping you might be able to help with something here.

This PR (the `th_state_url` shared-store feature for dialog-less topology hiding state) was ready and working when I opened it back in July, and I've been holding off on further commits while waiting for review, per my earlier comment on the branch.

In the meantime, it looks like `topology_hiding` went through a significant restructuring upstream (the file split into dialog/no-dialog/common logic, plus what looks like a broader rework of the one-way-hiding and thinfo compaction logic). I went to rebase this branch onto current master to clear the merge conflicts, and it turns out this isn't a simple rebase - several of the functions this PR touches were renamed and restructured, not just moved, so a mechanical conflict resolution isn't really possible here. Making this work against current master means re-implementing the feature against the new structure, which is a meaningfully bigger undertaking than what I originally wrote and tested.

Before I put that time in, I wanted to ask: what would you recommend? A few possibilities I can think of, but you'd know better than me:

- Go ahead and adapt/re-port the feature onto the current code - in which case, is there anything about the *design* that should change given whatever motivated the recent rework, or is it purely a structural relocation from your perspective?
- Hold off on adapting it until a review happens first, in case the approach itself needs to change before it's worth re-implementing.
- Something else entirely - e.g. if the recent rework already covers part of what this was trying to do, or there's a preferred way to sequence this against other in-flight topology_hiding work.

Happy to do whatever's most useful here - just didn't want to sink a lot of time into a re-port before hearing whether the approach is even still the right one.

PR #4115, opened 2026-07-19
tls_mgm: fix require_cert and verify_cert enforcement for virtual domains

ipoddubny:
**Summary**
Fixes the enforcement of require_cert and verify_cert settings for virtual TLS domains sharing a single socket.

**Details**
This is a generic bug fix.

Multiple virtual TLS domains can be defined on a single socket via match_sip_domain (using SNI), each with distinct require_cert and verify_cert settings.
However, these domain-specific verification rules are never applied to the connection during the TLS handshake, leaving only the socket's default settings enforced.

**Solution**
Adding SSL_set_verify to the SNI callbacks for both OpenSSL and WolfSSL to ensure the correct domain-specific rules are applied.

**Compatibility**
There should be no issues. The change may affect invalid configurations with mixed verify/require values for domains on the same socket.

**Closing issues**
I don't see any previous issues reporting this bug, submitting PR directly.

### DIFF
--- a/modules/tls_openssl/openssl_config.c
+++ b/modules/tls_openssl/openssl_config.c
@@ -326,12 +326,17 @@ int openssl_reg_sni_cb(tls_sni_cb_f cb)
 
 int openssl_switch_ssl_ctx(struct tls_domain *dom, void *ssl_ctx)
 {
+	int verify_mode = 0;
+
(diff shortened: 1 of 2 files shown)


razvancrainea:
This definitely makes sense, thanks for reporting and taking care of it!

PR #4118, opened 2026-07-23
cachedb_perf: high-performance local memory cache built on modern kernel features

Lt-Flash:
## Summary

`cachedb_perf` is a local in-memory `cachedb` backend, selected by URL scheme (`perf://`). It implements the standard `cachedb_funcs` vtable, so every module that takes a `cachedb_url` (`topology_hiding`, `dialog` profiles, `sql_cacher`, ...) and the core `cache_*` script functions work unchanged — changing the URL is the whole migration.

It exists because `cachedb_local`'s single fixed-size table and per-bucket locking become the bottleneck at high churn: at 50k resident entries and the topology-hiding access pattern, `cachedb_perf` does **448 ns/op vs 1,013 ns** (2.3×; 7.9× vs `cachedb_local`'s 512-bucket default), pure writes are 5.6–9.9× faster and stay flat with concurrency, and end-to-end topology hiding sustains **6,000 CPS where `cachedb_local` breaks at ~3,900**.

How: lock-free reads (per-bucket version + copy-out validation; writers hold a bucket lock for ~hundreds of ns), named collections that grow at runtime, instant expiry with a cheap background sweep, and records in a size-class slab arena — in core shm or a dedicated reservation (`memory_backing`), with full reclaim in the dedicated modes. Keys hash with MurmurHash3 (the core's hash collides on sequential keys).

Optional, each degrading explicitly when absent: persistence to any `db_*` backend, cluster warm-up sync, and cross-node pull-on-miss — over the clusterer's bin links, the `clusterer_controller` plane, or the module's own udp/tcp sockets.

Outside the module, one optional cachedb capability: `CACHEDB_CAP_GET_BUF`, an allocation-free read into the caller's buffer. Backends and consumers that do not know it are unaffected.

## Usage

```
loadmodule "cachedb_perf.so"
modparam("cachedb_perf", "cache_collections", "th=16")
modparam("cachedb_perf", "cachedb_url", "perf:///th")

loadmodule "topology_hiding.so"
modparam("topology_hiding", "th_state_url", "perf:///th")
```

From the script the backend id is `perf` (or `perf:<group>` when a grouped URL is declared): `cache_store("perf", "call-$ci", "$var(state)", 3600);`

Complete worked configurations - standalone, clustered over the stock clusterer, and clusterer_controller-managed, with every pull mode and transport explained: [Configuration examples](https://github.com/Lt-Flash/opensips/blob/feature/cachedb-perf-devel/modules/cachedb_perf/doc/CONFIG-EXAMPLES.md).

## Parameters

| parameter | default | |
|---|---|---|
| `cache_collections` | `default=14` | collections, `name[=log2(buckets)];...` — the size is only a starting point, tables grow |
| `cachedb_url` | `perf://` | URL(s) to resolve; `perf:///name` picks the collection; repeatable |
| `memory_backing` | `auto` | where records live: `core` shm, `own-hg` (dedicated huge-page arena), `own` (dedicated slab in shm, full reclaim) |
| `arena_hugepage_mb` | `0` | huge-page reservation size for the records; `0` = plain shm |
| `arena_hugepage_cap_mb` / `arena_profile` | — | growth cap / sizing profile of the dedicated arena |
| `reclaim_keep`, `reclaim_quiet_s`, `reclaim_cooloff_s`, `reclaim_giveback` | 1/5/10/1 | own-backing reclaim tuning |
| `expiry_sweep_period` | `1` | seconds between reclaim passes (expiry itself is instant) |
| `growth_load_factor` / `growth_budget` | `2` / `4096` | entries-per-bucket target / max splits per tick |
| `db_url`, `db_table`, `db_mode`, `persist_collections` | — | persistence: backend, table, auto load/save policy, which collections |
| `sync_cluster_id`, `sync_shtag` | — | cluster + sharing tag for warm-up sync |
| `replicate_collections`, `pull_on_miss` | — | which collections may pull from peers; repair misses transparently |
| `pull_transport` | `bin` | how pulls travel: `bin`, `clctr`, or module-owned `udp`/`tcp` sockets |
| `pull_bind`, `pull_port` | — | the udp/tcp transport's own address |
| `pull_timeout_ms`, `pull_negative_ms` | 50/300 | pull backstop / negative-answer cache |
| `event_expired_collections`, `arena_selftest`, `htable_selftest` | — | opt-in events and startup self-tests |

MI (`perf_stats`, introspection, `perf_save`/`perf_load`/`perf_sync`), statistics and events are documented in the module README. All 16 MI commands with real invocation + response examples: [MI examples](https://github.com/Lt-Flash/opensips/blob/feature/cachedb-perf-devel/modules/cachedb_perf/doc/MI-EXAMPLES.md).

## Docs, testing

Everything measured, with graphs, lives in the branch: [doc/STUDY.md](https://github.com/Lt-Flash/opensips/blob/feature/cachedb-perf-devel/modules/cachedb_perf/doc/STUDY.md) (the study), [doc/PR-NOTES.md](https://github.com/Lt-Flash/opensips/blob/feature/cachedb-perf-devel/modules/cachedb_perf/doc/PR-NOTES.md) (progress notes), [bench/README.md](https://github.com/Lt-Flash/opensips/blob/feature/cachedb-perf-devel/modules/cachedb_perf/bench/README.md) (the test/benchmark harness). Builds clean with gcc and clang under `-Werror`; no new dependencies.


lean1ee:
Tested and verified on commit 1eaa330 (branch feature/cachedb-perf-latest).
Environment:
- Debian 12 (Linux x86_64), GCC 12.2.0
Verification Summary:
1. Core Capability (CACHEDB_CAP_GET_BUF):
- Verified backward compatibility: existing modules (cachedb_sql, db_cachedb, etc.) compile cleanly without modifications.
- Implemented native get_buf() in the cachedb_tarantool IProto driver. Zero-allocation read path functions exactly according to specification:
  * Returns 0 and populates *vlen on hit.
  * Returns -2 on miss/expiry.
  * Returns -3 and sets *needed when the destination buffer is insufficient.
2. Module cachedb_perf & Benchmark Suite:
- Benchmarks (concur, hashtest, lookup, expire2) built and passed.
- Concurrency throughput measured at 166.5 Mops/s (100% read, 8 threads) and 147.0 Mops/s (95/5 read/write mix).
- Timer-wheel background expiry verified with O(expired) sweep overhead.
3. End-to-End SIP Validation:
- SIP traffic load test executed (100 calls @ 25 cps).
- Zero packet drops, zero memory leaks, 100% call success rate.
Conclusion:
The patch is clean, fully backward-compatible with external backends, and ready for upstream merge into master.

PR #4120, opened 2026-07-25
topology_hiding: do not warn per request on a non-local send socket

Lt-Flash:
## What

Two problems in `topo_no_dlg_seq_handling()`, in the branch that restores the send socket recorded in the encoded topology-hiding state: a log line that fires on every sequential request in any multi-node deployment, and — in the same branch — the unusable socket being assigned anyway, clearing whatever the script had already chosen.

## 1. A WARN per sequential request

When dialog-less topology hiding decodes the state of a sequential request, it looks up the socket the state was created on:

```c
sock = grep_sock_info(&host, (unsigned short)port, proto);
if (!sock) {
    LM_WARN("non-local socket <%.*s>...ignoring\n", bind_buf.len, bind_buf.s);
}
```

A node not owning that socket is **routine**, not exceptional: any anycast or load-balanced tier can land a sequential request on a different node than the one that handled the initial request, and a node whose listening set has simply changed hits it too. The result is one WARN per sequential request, per call — on a busy presence deployment that is the bulk of the log.

Dropped to `LM_DBG`. The R-URI has already been restored by this point, so the request still routes; only the send socket is affected.

## 2. The failed lookup was not actually ignored

The message says *"ignoring"*, but the assignment sat **outside** the check:

```c
if (!sock) {
    LM_WARN(... "ignoring" ...);
}
msg->force_send_socket = sock;      /* <-- runs even when sock == NULL */
```

So on the failure path it stored `NULL` — wiping any socket the script had already selected (e.g. an earlier `force_send_socket()` in the route) rather than leaving it alone. The fix only assigns when a local socket was actually found:

```c
if (!sock) {
    LM_DBG("non-local socket <%.*s> - leaving the send socket to the caller\n", ...);
} else {
    msg->force_send_socket = sock;
}
```

## Impact

Log volume on multi-node topology-hiding deployments, and — for scripts that set a send socket before `topology_hiding_match()` — a silently discarded choice on any request whose state came from another node.

### DIFF
--- a/modules/topology_hiding/th_no_dlg_logic.c
+++ b/modules/topology_hiding/th_no_dlg_logic.c
@@ -1400,10 +1400,23 @@ static decoded_info_buffer_t decode_info_buffer(str *info, const thinfo_options_
     if (host.len > 0 && host.s != NULL) {
 		decoded_buffer.sock = grep_sock_info(&host, port, proto);
 
-		if (!decoded_buffer.sock && th_internal_trusted_tag.len > 0) {
-			decoded_buffer.sock = grep_internal_sock_info(&th_internal_trusted_tag, 0, proto);
-		} else if (decoded_buffer.sock && th_internal_trusted_tag.len == 0) {
-			LM_WARN("non-local socket <%.*s:%d>...ignoring\n", host.len, host.s, port);
+		/* th_internal_trusted_tag governs one-way-hiding on trusted
+		 * internal sockets - it is not a fallback for resolving a
+		 * foreign encoded socket. th_external_socket_tag is the tag
+		 * documented for exactly that (see th_external_socket_tag doc
+		 * and the identical pattern in topo_no_dlg_match_uri() above). */
+		if (!decoded_buffer.sock && th_external_socket_tag.len > 0) {
+			decoded_buffer.sock = grep_internal_sock_info(&th_external_socket_tag, 0, proto);
+		}
+
+		if (!decoded_buffer.sock) {
+			/* Routine whenever topology hiding state reaches a node
+			 * other than the one that created it (e.g. an anycast or
+			 * load-balanced tier) and no external socket tag resolved
+			 * it either - DBG, not WARN, so it doesn't fire on every
+			 * sequential request of every such call. */
+			LM_DBG("non-local socket <%.*s:%d> - leaving the send socket "
+				"to the caller\n", host.len, host.s, port);
 		}
 
 		if (decoded_buffer.sock) {
@@ -1499,10 +1512,25 @@ static decoded_info_buffer_t decode_info_buffer_legacy(str *info, const thinfo_o
             LM_ERR("bad socket <%.*s>\n", bind_buf.len, bind_buf.s);
         } else {
             decoded_buffer.sock = grep_sock_info(&host, (unsigned short) port, proto);
-            if (!decoded_buffer.sock && th_internal_trusted_tag.len > 0) {
-				decoded_buffer.sock = grep_internal_sock_info(&th_internal_trusted_tag, 0, proto);
-            } else {
-				LM_WARN("non-local socket <%.*s>...ignoring\n", bind_buf.len, bind_buf.s);
+
+			/* th_internal_trusted_tag governs one-way-hiding on trusted
+			 * internal sockets - it is not a fallback for resolving a
+			 * foreign encoded socket. th_external_socket_tag is the tag
+			 * documented for exactly that (see topo_no_dlg_match_uri()). */
+            if (!decoded_buffer.sock && th_external_socket_tag.len > 0) {
+				decoded_buffer.sock = grep_internal_sock_info(&th_external_socket_tag, 0, proto);
+            }
+
+			if (!decoded_buffer.sock) {
+				/* Routine whenever topology hiding state reaches a node
+				 * other than the one that created it (e.g. an anycast or
+				 * load-balanced tier) and no external socket tag resolved
+				 * it either - DBG, not WARN, so it doesn't fire on every
+				 * sequential request of every such call. Note this is
+				 * only the send socket; the R-URI has already been
+				 * restored, so the request still routes. */
+				LM_DBG("non-local socket <%.*s> - leaving the send socket "
+					"to the caller\n", bind_buf.len, bind_buf.s);
 			}
         }
     }
--- a/modules/topology_hiding/topo_hiding_logic.c
+++ b/modules/topology_hiding/topo_hiding_logic.c
@@ -2332,9 +2332,17 @@ static int topo_no_dlg_seq_handling(struct sip_msg *msg,str *info)
 		} else {
 			sock = grep_sock_info( &host, (unsigned short)port, proto);
 			if (!sock) {
-				LM_WARN("non-local socket <%.*s>...ignoring\n", bind_buf.len, bind_buf.s);
+				/* The state was encoded by a node that owns this socket
+				 * and this one does not - normal whenever topology hiding
+				 * state reaches a different node than created it, so DBG:
+				 * at WARN it fires on every sequential request of every
+				 * such call.  Note this is only the send socket; the R-URI
+				 * has already been restored, so the request still routes. */
+				LM_DBG("non-local socket <%.*s> - leaving the send socket "
+					"to the caller\n", bind_buf.len, bind_buf.s);
+			} else {
+				msg->force_send_socket = sock;
 			}
-			msg->force_send_socket = sock;
 		}
 	}
 


Lt-Flash:
A note on why the `force_send_socket` half of this fix matters beyond the WARN-spam it silences: in follow-up work already implemented on a development branch, topology_hiding can fetch a hidden-dialog state from another cluster node when a sequential request lands on the wrong machine (failover / re-hash behind a load balancer). A state created on node A and restored on node B may name a send socket that only exists on A — so the restore path depends on exactly the socket-handling discipline this PR establishes: respect the stored socket when it resolves locally, fall back cleanly to a local socket of the matching protocol when it doesn't, and never let a stale value clobber a route-set decision. This PR is a prerequisite for that work landing, which is one more reason to keep it a small standalone fix.

bogdan-iancu:
Thank you @Lt-Flash for the worker here. I still consider that the logs should stay as a warning, as it is an indicator if something in wrong with the traffic or the setup.
Now, in your particular case, have you tried to the `tag`s for the sockets, to logically bind sockets (even with different IPs) from different servers ? (see https://docs.opensips.org/manual/devel/script-coreparameters/#socket)

Lt-Flash:
Thanks for taking a look, @bogdan-iancu.

**On the socket tags question** — good prompt, I went and checked both the code and our actual deployment.

Our environment (3-node anycast SBC tier, sockets from `opensips.cfg` on each node, public IP redacted):

```
# Node A (private IP shown, real value differs per node)
socket=udp:10.22.23.X:5060  use_workers 2  use_auto_scaling_profile PROFILE_UDP_PRIV  tag udp_priv

# identical on every node in the tier - same tag, different real IP per node
socket=udp:10.22.23.Y:5060  use_workers 2  use_auto_scaling_profile PROFILE_UDP_PRIV  tag udp_priv
socket=udp:10.22.23.Z:5060  use_workers 2  use_auto_scaling_profile PROFILE_UDP_PRIV  tag udp_priv

# public/anycast side - literally the SAME IP bound on all three nodes
socket=udp:<PUBLIC-VIP-IP>:7060  use_workers 2  use_auto_scaling_profile PROFILE_UDP_PUB  tag udp_pub anycast
socket=tcp:<PUBLIC-VIP-IP>:7060  tag vip
socket=tls:<PUBLIC-VIP-IP>:7061  tag vip
socket=wss:<PUBLIC-VIP-IP>:7069  tag vip
```

The public/anycast leg never hits this bug in the first place - every node binds the identical literal IP, so the existing plain `grep_sock_info()` IP match already succeeds everywhere. The bug is specifically the **private leg**, where each node's socket is a genuinely different IP - and as you can see, **we already tag it identically (`udp_priv`) across every node**, precisely because we think of it as one logical role, not three distinct addresses.

So I checked whether that tag is actually usable today, and it isn't - not because tags don't work, but because `topology_hiding` doesn't use the mechanism on either side:
- The encode path (`topo_no_dlg_encode_contact`) serializes `msg->rcv.bind_address->sock_str` directly - the literal address, not `get_socket_internal_name()` (which returns `tag_sock_str`, the tag-based `"tag:port"` representation, when a tag is configured).
- The restore path this PR touches calls the plain `grep_sock_info()` macro - `grep_sock_info_ext(..., check_tags=0)` - so tag matching is off here regardless of config. Even the tag-aware variant only checks tags when `port==0`, which this call site never passes (it always parses a real port out of the stored socket string).

Given we already tag our private sockets exactly the way this would need, wiring `topo_hiding_logic.c` to encode/decode via the tag-aware path seems like a genuinely worthwhile follow-up - on our own tier it would let the restore succeed via the tag instead of falling back to the caller's socket at all. Happy to take that on separately if there's interest, but it's a real code change to the module (both encode and decode sides), not something achievable by this PR alone.

**On the log level** - I hear the concern, and agree it's a legitimate signal in cases where it indicates a real setup problem. The case this PR targets is specifically the routine one shown above: state legitimately created on one node, restored on another, as expected behavior in an anycast/LB tier - not a misconfiguration. In that topology it fires on every single sequential request of every such call, which is what made it unusable as a signal for us. Since the R-URI has already been correctly restored by this point and only the *send socket* falls back to the caller's choice, would you be open to something in between - e.g. rate-limiting the line, or only escalating to WARN after N occurrences for the same state - rather than either "always WARN" or "always DBG"? Open to whatever level best preserves the signal without the per-request flood.

Lt-Flash:
Update, since a lot happened to this file since I opened this PR - the underlying code moved into `th_no_dlg_logic.c` via the "no dialog"/refactor work, and it turns out something more concrete than the WARN-level question surfaced along the way.

**On the tags question specifically**: someone already added a socket-tag fallback here independently (`th_external_socket_tag`, documented for exactly this "encoded on a different node" case). But both restore paths that need it were actually calling `th_internal_trusted_tag` instead - a different modparam entirely, meant for one-way-hiding trust decisions, not socket resolution. The one place in the file that *does* use `th_external_socket_tag` correctly (`topo_no_dlg_match_uri()`'s auto-Route matching) shows the intended pattern - try the tag only after a direct match fails, and only then decide what to log.

The wrong-variable bug also broke the log condition itself: the `else` in both broken sites fired on *any* case other than "not found and tag configured" - which includes the socket resolving successfully with no problem at all, and stayed completely silent in the genuine unresolvable case. So the log wasn't just too loud, it was firing on the wrong condition entirely.

I've rewritten both sites to: try the tag fallback only after a direct miss, and only log once neither resolves it - which is also where I kept the DBG (down from WARN) for the reasons in my last comment. Rebased onto current master and force-pushed since the file changed too much for a normal rebase to apply cleanly.

Still very open to whatever log level you think is right for the "genuinely couldn't resolve it" case now that it's gated correctly - happy to bump it back to WARN, or the rate-limit idea, whatever you'd prefer.

bogdan-iancu:
For the 4.0 and below, yes, the best way will be to encode `get_socket_internal_name(msg->rcv.bind_address)` and use the `grep_internal_sock_info()` when decoding -> so TAGs will be used if available. Maybe we can push this fix until sorting out the issue on 4.1/devel

bogdan-iancu:
For 4.1/dev, see #3908 - the tags seems to have a different purpose, to control how the TH is done. I will check more

Lt-Flash:
No worries at all — happy for this to wait until the 4.1/devel side is sorted out. Thanks for looking into it, and for the pointer to #3908.

bogdan-iancu:
For 4.0 and below, please test the following fix:
````
diff --git a/modules/topology_hiding/topo_hiding_logic.c b/modules/topology_hiding/topo_hiding_logic.c
index 5b9d8c58db..6ffc436030 100644
--- a/modules/topology_hiding/topo_hiding_logic.c
+++ b/modules/topology_hiding/topo_hiding_logic.c
@@ -1765,6 +1765,7 @@ static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int
        str rr_set = {NULL, 0};
        str contact;
        str flags_str;
+       const str *sock_name;
        int i,total_len,enc_len;
        struct sip_uri ctu;
        struct th_ct_params* el;
@@ -1813,9 +1814,9 @@ static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int
        flags_str.s = int2str(flags, &flags_str.len);
        if (topo_ct_short_len(flags_str.len, &flags_len, "flags") < 0)
                goto error;
-       
-       if (topo_ct_short_len(msg->rcv.bind_address->sock_str.len,
-                       &addr_len, "bind address") < 0)
+
+       sock_name = get_socket_internal_name(msg->rcv.bind_address);
+       if (topo_ct_short_len(sock_name->len, &addr_len, "bind address") < 0)
                goto error;
        local_len += rr_len + ct_len + flags_len + addr_len; 
        enc_len = th_ct_enc_scheme == ENC_BASE64 ?
@@ -1892,8 +1893,8 @@ static char* build_encoded_contact_suffix(struct sip_msg* msg, str *routes, int
        p+= flags_str.len;
        memcpy(p,&addr_len,sizeof(short));
        p+= sizeof(short);
-       memcpy(p,msg->rcv.bind_address->sock_str.s,msg->rcv.bind_address->sock_str.len);
-       p+= msg->rcv.bind_address->sock_str.len;
+       memcpy(p, sock_name->s, sock_name->len);
+       p+= sock_name->len;
        for (i=0;i<(int)(p-suffix_plain);i++)
                suffix_plain[i] ^= topo_hiding_ct_encode_pw.s[i%topo_hiding_ct_encode_pw.len];
 
@@ -2330,7 +2331,7 @@ static int topo_no_dlg_seq_handling(struct sip_msg *msg,str *info)
                &port, &proto)!=0) {
                        LM_ERR("bad socket <%.*s>\n", bind_buf.len, bind_buf.s);
                } else {
-                       sock = grep_sock_info( &host, (unsigned short)port, proto);
+                       sock = grep_internal_sock_info( &host, (unsigned short)port, proto);
                        if (!sock) {
                                LM_WARN("non-local socket <%.*s>...ignoring\n", bind_buf.len, bind_buf.s);
````

**Comment by Lt-Flash** [comment]
Thanks @bogdan-iancu, this is a cleaner route than mine, and it settles the log-level question too.

One correction to my 5 August comment: I said the tag-aware lookup could never fire here because this call site always parses a real port. That holds for today's encoding, not for yours. A tagged socket's internal name is `proto:tag` with no port (`socket2str()` type 2), so `parse_phostport()` returns port 0 and `grep_sock_info_ext()` takes the tag branch - the same round trip `parse_sock_info()` already makes for dialog and b2b replication.

I tested your diff on stock 4.0 (`1f9f0279cb`, nothing else applied): two nodes, each with an untagged socket listed first and a socket tagged `priv` second. An INVITE goes through node A, then the BYE carrying A's state is delivered to node B instead:

| | stock 4.0 | with your diff |
|---|---|---|
| state decoded on B | `udp:127.0.0.2:5060` | `udp:priv` |
| `non-local socket` WARN on B | yes | no |
| BYE sent from | B's default socket | B's `priv` socket |

Untagged sockets behave exactly as before, and a mixed pair (old encoder with new decoder, or the reverse) falls back to the default socket with the WARN, so a rolling upgrade is safe.

So agreed on WARN: once tags resolve the routine cross-node case, a miss genuinely means an untagged or misconfigured socket, and it matches the warning dialog and b2b replication already log.

One small suggestion, your call: on a miss the assignment after the WARN still stores the NULL, so a `force_send_socket()` the script made before `topology_hiding_match()` is lost. On the same rig, set before the match the BYE left from the default socket; set after it, the script's choice held. Assigning only on success avoids that:

```c
sock = grep_internal_sock_info(&host, (unsigned short)port, proto);
if (!sock)
	LM_WARN("non-local socket <%.*s>...ignoring\n", bind_buf.len, bind_buf.s);
else
	msg->force_send_socket = sock;
```

For completeness: I rely on the stateful mode from #4114, which writes the socket a second time, into the stored state, so that writer will need the same `get_socket_internal_name()` change.

Happy to open this against 4.0 as a separate PR if that helps, or leave it with you.

bogdan-iancu:
OK, did the fix on 4.0 and 3.6, taking also into the account the idea around the NULL sock on grepping.

Now, what is left is the review and fixing on devel

Lt-Flash:
Thanks @bogdan-iancu — the 4.0 and 3.6 commits look right to me, and thanks for taking the NULL-sock point.

On devel I hit something that stops the same fix from porting across as-is, so I would rather ask than guess.

The default encoder on devel is the compact one: the built-in `thinfo` slot ships `compact_encoding = 1` (`th_no_dlg_logic.c:118`), and `th_add_encode_param_password()` also defaults new slots to 1 unless the value ends in `:L` (`:2096`). The legacy encoder — `build_encoded_contact_suffix_legacy()`, which writes `msg->rcv.bind_address->sock_str` at `:737`/`:836` and is the direct analogue of your 4.0 hunk — is only reached on that opt-in path.

The compact path encodes the socket through `thinfo_encode_socket()` (`thinfo_codec.c:329`): one flags byte, then the raw address (`memcpy(p, si->address.u.addr, 4|16)`), then an optional 2-byte port. `get_socket_internal_name()` returns a string (`proto:tag`), so there is no field for it. Porting your change to the legacy encoder alone would fix the opt-in path and leave the default still encoding a literal address, which is why I have not pushed anything here yet.

What ports cleanly either way:

- the three socket lookups to `grep_internal_sock_info()`: `:280` (auto-Route), `:1401` (compact), `:1507` (legacy);
- two inverted conditions: `:1405` warns when the socket *was* found, and the `else` at `:1510` fires whenever it was found or no tag is configured — so the WARN currently sits on the wrong branch in both;
- the NULL-sock guard you added on 4.0 is already present on devel (`:289`, `:1672`).

For the compact format there is room — `SOCKET_IP_MASK` (0x18) has two unused values and bits 6-7 of the flags byte are free — so a length-prefixed tag form would be small to add. The catch is that it is not transparently backward compatible: an older node hits `return -1` on the unknown ip_type (`thinfo_codec.c:436`), and both callers treat that as fatal (`:274`, `:1397` → `INVALID_BUF`), so `th_no_dlg_seq_handling()` returns -1 at `:1578` and the whole match fails — Contact and routes included, not only the send socket. In a mixed-version tier that would drop in-flight sequential requests rather than degrade to the current fallback, so it would need gating behind an explicit opt-in.

Since the codec is yours and you mentioned devel still needs review: would you like me to implement the tag form (and if so, gated how), or would you rather take devel yourself? Happy either way — I will hold off pushing to this PR until you say.

PR #4218, opened 2026-08-18
net_tcp: reject NULL message buffer with non-zero length in tcp_dispatch_msg()

### LINKED ISSUE #4217: [CRASH] SIGSEGV in memcpy via tcp_dispatch_msg(), called from proto_ws's ws_process() with a NULL message buffer
### Summary

`opensips` (4.0, running a build we still label `+otfix1` for packaging-history reasons — see
"Is this related to a known issue?" below; the code itself is identical to stock `4.0`/`master`,
confirmed by diff) crashed with SIGSEGV while processing a WebSocket
Secure (WSS) connection. The crash is a `memcpy()` with a NULL source pointer and a large garbage
length, called from `tcp_dispatch_msg()` in `net/net_tcp.c`, itself called from `proto_ws`'s
`ws_process()` (`modules/proto_ws/ws_common.h`) on the `WS_OP_TEXT`/`WS_OP_BIN` dispatch path.

The crash brought the whole `opensips` process down (SIGCHLD-triggered clean shutdown of all
worker processes), which in our deployment triggered an HA failover — the process was not stuck,
it was gone.

### Environment

- OpenSIPS 4.0.0, Debian package build labeled `+otfix1` (see "Is this related to a known issue?"
  below — the code is current with upstream, this is just a stale version label)
- Linux x86_64
- Crash occurred inside a `proto_wss` connection (WSS on port 7443)
- One occurrence observed; first in at least 30 days of uptime on this pair of nodes

### Backtrace (via `coredumpctl` + `gdb`, debug symbols installed)

```
Stack trace of thread <TCP worker>:
#0  memcpy (libc.so.6)
#1  memcpy (opensips + tcp_dispatch_msg, net/net_tcp.c:321)
    __dest=<payload buffer>, __src=0x0, __len=872728144
#2  ws_process (proto_wss.so, ../proto_ws/ws_common.h:547)
#3  tcp_thread_routine (opensips)
```

`frame 2` (inlined into `tcp_dispatch_msg`) confirms the exact parameter values at the crash site:

```
tcp_dispatch_msg (msg=0x0, len=872728144, rcv=0x..., data=0x0, data_len=0)
    at net/net_tcp.c:321
```

The call site in `proto_ws`'s `ws_process()` is the `WS_OP_TEXT`/`WS_OP_BIN` case:

```c
case WS_OP_TEXT:
case WS_OP_BIN:
    bk = *req->tcp.parsed;
    *req->tcp.parsed = 0;
    msg_buf = req->tcp.body;
    msg_len = req->tcp.parsed - req->tcp.body;
    local_rcv = con->rcv;
    ...
    if (tcp_dispatch_msg(msg_buf, msg_len, &local_rcv, NULL, 0) < 0)
        LM_ERR("failed to deliver WS message\n");
    ...
```

At the crash: `req->tcp.body` (== `msg_buf`) was `NULL`, and `msg_len` (declared `int`, a 32-bit
type, holding the result of a 64-bit pointer subtraction) was `872728144` — clearly not a real
message length.

### What we could and couldn't confirm

- **Confirmed (gdb, live coredump):** the exact crash mechanism and parameter values above.
- **Confirmed (source review):** `ws_parse()`'s state-reset macros (`init_tcp_req`/`init_ws_req`)
  correctly zero `req->tcp.body` and `req->tcp.complete` together, and every code path we traced
  through `ws_parse()` either sets `req->tcp.body` to a valid pointer before the "do we have the
  full packet" check, or exits/`goto`s before reaching it. We could not, through static reading
  alone, identify the exact sequence that leaves `req->tcp.body == NULL` while `req->tcp.complete`
  is set (or otherwise reaches the `WS_OP_TEXT`/`WS_OP_BIN` dispatch with `body` unset). It's
  possible this needs a specific interleaving of pipelined WS frames, a specific TCP segmentation
  pattern, or memory corruption from elsewhere — we don't have enough evidence to say which.
- **Not available:** a packet capture of the triggering connection. Our wire-capture ring only
  retains a short rolling window and had long since rotated past the crash time before we
  investigated, so we can't show the exact bytes the client sent.

### Is this related to a known issue?

Our build carries a version label `otfix1` from an earlier local patch for an unrelated
connection-reference leak in `tcp_async_write_job()`'s dedup branch (a write-path bug, not a
read/parse-path bug) — that fix has since been merged upstream as #4125, in both `master` and
`4.0`, so it's not a current code divergence, just a stale package label. We diffed both
`net/net_tcp.c` and `modules/proto_ws/ws_common.h` against current `upstream/4.0` and
`upstream/master` — **byte-identical in both files**, confirming this crash isn't caused by
anything on our side, and (as far as we can tell) isn't already fixed upstream either.

### Proposed fix

We're not confident enough in the exact `ws_parse()` trigger to patch that function specifically
without risking papering over a deeper state bug. Instead, we're proposing a minimal,
caller-agnostic hardening in `tcp_dispatch_msg()` itself: it already rejects a negative `len`, a
negative `data_len`, and a missing `data` buffer for a non-zero `data_len` — but not a NULL `msg`
with a non-zero `len`, which is exactly what reaches the `memcpy` unchecked. Adding that check
means any caller with a similar bug (this one, or a future one) fails safe (`LM_BUG` + `return -1`)
instead of crashing the whole process. PR incoming.

---
Investigated and written up jointly by Giovanni Maruzzelli (gmaruzz) and Claude (Anthropic).

gmaruzz:
### What

`tcp_dispatch_msg()` (`net/net_tcp.c`) guards against a negative `len`, a negative `data_len`, and
a missing `data` buffer for a non-zero `data_len` — but not against a NULL `msg` with a non-zero
`len`, which reaches `memcpy(payload->msg_buf, msg, len)` a few lines below, unchecked.

This adds the missing check, in the same style as the existing three:

```c
if (len && !msg) {
    LM_BUG("NULL TCP message buffer with non-zero length: %d\n", len);
    return -1;
}
```

### Why

Observed live: `proto_ws`'s `ws_process()` (`WS_OP_TEXT`/`WS_OP_BIN` case) called
`tcp_dispatch_msg(msg_buf, msg_len, &local_rcv, NULL, 0)` with `msg_buf == NULL`
(`req->tcp.body`) and `msg_len` a large garbage value, crashing the whole process with SIGSEGV.
Full report in #4217.

We were not able to conclusively pin the exact sequence in `ws_parse()` that leaves
`req->tcp.body` NULL at that point — every code path we traced sets it before the "packet
complete" check, or exits before reaching it — so we're not proposing a speculative fix there.
This PR instead hardens the shared dispatch function itself, which protects every current and
future caller against the same failure mode regardless of which one has the bug: fail safe
(`LM_BUG` + `return -1`) instead of crashing the process.

### Testing

- Confirmed this exact guard would have caught the observed crash (the recorded parameters at
  the crash site were `msg=0x0, len=872728144`).
- No functional change for any well-formed caller — every existing caller in the tree passes a
  non-NULL `msg` whenever `len > 0`.

Fixes #4217

---
Diagnosed and prepared jointly by Giovanni Maruzzelli (gmaruzz) and Claude (Anthropic).

### DIFF
--- a/net/net_tcp.c
+++ b/net/net_tcp.c
@@ -303,6 +303,10 @@ int tcp_dispatch_msg(char *msg, int len,
 		LM_BUG("negative TCP message length: %d\n", len);
 		return -1;
 	}
+	if (len && !msg) {
+		LM_BUG("NULL TCP message buffer with non-zero length: %d\n", len);
+		return -1;
+	}
 	if (data_len < 0) {
 		LM_BUG("negative TCP dispatch data length: %d\n", data_len);
 		return -1;


razvancrainea:
Hi, @gmaruzz ! So I understand this is not something observed upstream, or a crash, it's just something that got out of an audit, and would be nice to prevent it in case it happens, correct?
Nevertheless, I will merge it just to be safe than sorry :) Thank you!

gmaruzz:
We actually got a crash, and that is what claude spit :))

Issue #4135, opened 2026-08-10
[BUG] - a second SDP rebuild terminates a worker, causing a complete restart of opensips


sindy39:
**OpenSIPS version you are running**

```
version: opensips 3.6.7 (x86_64/linux)
flags: STATS: On, DISABLE_NAGLE, USE_MCAST, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, HP_MALLOC, F_PARALLEL_MALLOC, DBG_MALLOC, FAST_LOCK-ADAPTIVE_WAIT
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535
poll method support: poll, epoll, sigio_rt, select.
git revision: eaee48e28e
main.c compiled on  with gcc 15
```

**Describe the bug**

An SDP rebuild triggered by the second SDP change causes the worker to terminate, leading to opensips core shutdown on SIGCHLD (see the log below).

**To Reproduce**

1. Set up an `opensips` instance with `rtpengine`; use `rtp_relay` module to replicate my setup as closely as possible.
2. In the configuration script, use an assignment to `$sdp.line` and `rtp_relay_engage()` (the mutual order of these two does not matter), as in the example below.
3. Send an initial INVITE carrying an SDP - the worker dies, causing the rest of the processes to shut down as well, see the log below. If you comment either the `$sdp.line(o=)` assignment or the `rtp_relay_engage()`, the INVITE gets through successfully.

```
branch_route[per_branch_ops] {

        if ( has_body("application/sdp") ) {
                if ( $socket_out(af) == "INET6" ) {
                        $var(o) = $(sdp.line(o=){re.subst,/^(.*IN IP).*/\1/}) + "6 " + $(socket_out(ip){re.subst,/^.(.*).$/\1/});
                } else {
                        if ( isflagset("s2u") ) {
                                $var(o) = $(sdp.line(o=){re.subst,/^(.*IN IP).*/\1/}) + "4 " + $socket_out(ip);
                        } else {
                                $var(o) = $(sdp.line(o=){re.subst,/^(.*IN IP).*/\1/}) + "4 " + $socket_out(advertised_ip);
                        }
                }
                xlog("L_INFO","socket_out(ip) = >$socket_out(ip)<\r\n");
[...]

PR #4238, opened 2026-09-03
sdp_ops: implement token-indexed writes for $sdp.line/.stream/.session

Lt-Flash:
Related to #4135 (companion to #4237, which fixes the crash itself).

`$sdp.line(o=/[5]) = "0.0.0.0";` — a token-indexed write — was a silent no-op: the `handle_token_edit:` label in `pv_set_sdp_line()`, `pv_set_sdp_stream()` and `pv_set_sdp_session()` was an empty stub. The statement located the line, did nothing, set `SDP_OPS_FL_DIRTY` and returned success. Token indexing existed in the getters but was never implemented in the setters — which is also the answer to the second question in #4135 (`o=/[4] = "IP6"` "flags dirty without visible changes"). The DIRTY flag alone was enough to re-derive `msg->body` from sdp_ops' buffer, i.e. the no-op write is what armed the crash fixed in #4237.

**Implemented**, following `docs/manual/Script-CoreVar.md`:

- one `sdp_ops_find_token()` shared by readers and writers, so "token N" cannot mean two different things; the three getters now use it instead of three copies of the token walk;
- one `sdp_ops_set_token()` behind all three `handle_token_edit:` labels;
- a token prefix is a **key that survives the write**: `$sdp.line(a=fmtp:96/bitrate=) = "64000"` yields `bitrate=64000`; matched by bare index (`o=/[5]`) the whole token is the value;
- deleting a token takes the whitespace on one side with it (no double space left); a line left empty is dropped rather than kept blank;
- a value containing an EOL is refused (a token lives inside one line); `[insert]` / `[insertAfter]` are refused on tokens, being line operations.

**Verified** with a 15-assertion script rig (read-side forms from the manual, replace by index, replace by prefix keeping the key, delete without double spaces, untouched lines intact, the `$sdp.session` path) — run first against the unpatched build, where every edit assertion failed with `AFTER == BEFORE`. Builds clean with `-Wredundant-decls -Wold-style-definition -Wmissing-field-initializers -Werror` on current master, independently of #4237.

Two examples in the manual are self-inconsistent and were left alone: `m=audio[1]/[0]` is documented as `audio` but the reader has always returned `m=audio`, and `a=nortpproxy:/[0] = "yes"` cannot work under whitespace tokenisation.

### DIFF
--- a/sdp_ops.c
+++ b/sdp_ops.c
@@ -551,14 +551,140 @@ static inline int sdp_ops_find_line(struct sip_msg *msg, struct sdp_body_part_op
 }
 
 
+static int sdp_ops_find_token(str *line, int idx, str *pfx, str *tok)
+{
+	char *p, *lim, *start;
+
+	tok->s = NULL;
+	tok->len = 0;
+
+	start = line->s;
+	lim = line->s + line->len;
+	while (start < lim && is_ws(*start))
+		start++;
+
+	for (p = start; p <= lim; p++) {
+		if (p < lim && !is_ws(*p))
+			continue;
+
+		if (pfx->len && (pfx->len > (p-start) || strncasecmp(pfx->s, start, pfx->len))) {
+			while (p < lim && is_ws(*p))
+				p++;
+			start = p;
+			continue;
+		}
+
+		if (idx > 0) {
+			idx--;
+			start = p;
+			while (start < lim && is_ws(*start))
+				start++;
+			continue;
+		}
+
+		tok->s = start;
+		tok->len = p - start;
+		return 0;
+	}
+
+	return -1;
+}
+
+
+static int sdp_ops_set_token(struct sdp_body_part_ops *ops, int line_idx,
+		int tok_idx, str *pfx, str *val)
+{
+	str line, token, new_line, new_tok = STR_NULL;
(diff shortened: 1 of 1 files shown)


Lt-Flash:
Rebased onto current master (`0a33489dc5`) — the branch was 33 commits behind, and re-running jobs here isn't something I can do, so pushing was the only way to re-trigger CI. The patch content is unchanged: 1 file, +166/-108.

The earlier failures were not from this patch. Both build jobs died during `apt install` of the build dependencies, before anything was compiled:

```
Processing triggers for libc-bin (2.35-0ubuntu3.14) ...
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
dpkg: error processing package libc-bin (--configure):
 installed libc-bin package post-installation script subprocess returned error exit status 139
```

That is `ldconfig` segfaulting under qemu-user arm64 emulation. The two `all_done / notify_slack` entries are just the aggregate reporter for those jobs. #4237, opened minutes earlier from the same base, passed all four arm64 jobs.

Rebuilt clean on the new base with the CI flag set (`-Wredundant-decls -Wold-style-definition -Wmissing-field-initializers -Werror`), rc 0.

PR #4242, opened 2026-09-04
docs: restore the pn_ parameter names in registrar, mid_registrar and…

denys-i-didww:
**Summary**

Commit e941906dd9 (`docs: fix README.md typos`, 5 Aug 2026) replaced every occurrence of
`pn` with `on` in the registrar and mid_registrar documentation. That commit fixed a great
many real typos across 135 files; in these pages it also rewrote identifiers, and `pn`
here is not a typo — it is Push Notification.

Eight module parameters in each of the two modules now carry a name that exists nowhere in
the source tree:

| Documented today | Registered in `lib/reg/pn.h` |
|---|---|
| `on_enable` | `pn_enable` |
| `on_providers` | `pn_providers` |
| `on_ct_match_params` | `pn_ct_match_params` |
| `on_pnsreg_interval` | `pn_pnsreg_interval` |
| `on_trigger_interval` | `pn_trigger_interval` |
| `on_skip_on_interval` | `pn_skip_pn_interval` |
| `on_refresh_timeout` | `pn_refresh_timeout` |
| `on_enable_purr` | `pn_enable_purr` |

The parameters come from one macro in `lib/reg/pn.h`, which is why the same eight are
affected in both modules. A configuration written from the current pages does not start:
OpenSIPS refuses an unknown `modparam` while parsing, before any module is initialised.

**What else the substitution reached**

- the async function `pn_process_purr()`, documented as `on_process_purr()`;
- the RFC 8599 Contact URI parameters, in prose and as the default value of
  `pn_ct_match_params`: `pn-provider, pn-prid, pn-param` became `on-provider, on-prid,
  on-param`, and `;pn-purr=XXX` became `;on-purr=XXX`;
- the abbreviation itself — *"Push Notification (PN) support"* now reads *"Push
  Notification (ON) support"*, and *"PN-enabled contact"* reads *"ON-enabled contact"*;
- one cross-reference in `usrloc`, which cites registrar's `pn_trigger_interval`.

`on_skip_on_interval` is the clearest fingerprint: both halves of `pn_skip_pn_interval`
were replaced, including the one in the middle of the name.

**Reproduced**

On the official `opensips/opensips:4.0` image, one `modparam` per run, `opensips -C`:

```
$ opensips -C -f t.cfg            # a name the README carries today
ERROR:core:set_mod_param_regex: parameter <on_enable> not found in module <registrar>
ERROR:core:parse_opensips_cfg: bad config file (1 errors)

$ opensips -C -f t.cfg            # the name this PR restores
(exit 0)
```

All sixteen documented names are refused, eight in each module, with the same message.
All sixteen restored names are accepted.

`pn_enable` needs one note: set to `1` it declares a dependency on `tm` and
`event_routing` (`lib/reg/pn.c:227-238`), so a config without `event_routing` stops on
that instead — with the module loaded it starts, and so does `pn_enable=0`. The swept
name `on_enable` is still refused as *not found* with `event_routing` loaded, so the name
is the variable and the dependency is not.

The async function was checked the same way, in a script that calls it: `pn_process_purr()`
loads, `on_process_purr()` fails to parse.

**Scope**

This PR restores `pn` only. Everything else that commit changed in these files is left as
it is — `overwride` → `overrides` and the rest are real corrections and are not touched.

**Verification**

After this change every parameter documented in the two modules matches a name registered
in `lib/reg/pn.h` — the eight above, checked by name against the macro. Nothing else was
rewritten: `t_on_reply()` and `t_on_branch()` still read as before in the registrar
examples, and no `on_`, `on-` or `ON` token that belongs there was touched.

**Not in this PR**

The same commit renamed two more identifiers elsewhere, and they need their own decision
rather than being folded in here:

- `httpd`: the documented callback `httpd_acces_handler_cb` became
  `httpd_access_handler_cb`. The typo is real, but it is a C API symbol declared in
  `httpd_load.h:68` and used at `httpd.c:240`, so the fix belongs in the code, not in the
  page describing it.
- `db_virtual`: `db_max_consec_retrys` became `db_max_consec_retries`. Same shape — the
  code carries the misspelling, and the correct fix is in the code. Happy to send either.

---

**AI assistance disclosure.** This report and the change were produced with AI assistance.

- the eight names are registered in `lib/reg/pn.h:90-97`, in a macro shared by both
  modules;
- no `on_*` form of any of them appears in any `.c` or `.h` file in the tree —
  repository-wide grep;
- `pn_process_purr` is the async function checked for at `lib/reg/pn.c:211` and `:217`;
- `pn-provider`, `pn-prid`, `pn-param` are the literals at `lib/reg/pn.c:54-56`, and the
  default value of the parameter at `lib/reg/pn.c:44`;
- the failures and the successes were run, not inferred — sixteen parameters in two
  modules, plus the async function, on the official 4.0 image; the commands and their
  output are quoted above;
- the substitution was traced to a single commit — `git log -S` on each name over the full
  history, unshallowed for this purpose — rather than assumed from the current state;
- the two identifiers listed under *Not in this PR* were found by the same pass over all
  135 files that commit touched.

### DIFF
--- a/modules/mid_registrar/README.md
+++ b/modules/mid_registrar/README.md
@@ -120,23 +120,23 @@ The mid_registrar module includes support for standards-based SIP Push
 Notifications, per
 [RFC 8599](https://tools.ietf.org/html/rfc8599).
 Support for the basic version of the draft can be enabled by switching
-[on enable](#on_enable-boolean) to *true*.  The
+[pn enable](#pn_enable-boolean) to *true*.  The
 module also includes optional support for sending Push Notifications
 during long-lived dialogs ([see RFC section 6](https://tools.ietf.org/html/rfc8599#page-23)),
-through the [on enable purr](#on_enable_purr-boolean) switch.
+through the [pn enable purr](#pn_enable_purr-boolean) switch.
 
 
-Essential mechanics behind the Push Notification (ON) support:
+Essential mechanics behind the Push Notification (PN) support:
 
 
 - the ON support is fully compatible with the existing logic and
 enabling it does not impose any limitations, as the
-mid_registrar can simultaneously handle both SIP ON compliant
+mid_registrar can simultaneously handle both SIP PN compliant
 and standard SIP User Agents
 - OpenSIPS will raise a
 [E_UL_CONTACT_REFRESH](../usrloc/README.md#e_ul_contact_refresh)
 event any time a Push Notification needs to be sent to a
-ON-enabled contact.  The event includes the ON coordinates of
+PN-enabled contact.  The event includes the PN coordinates of
 the contact -- they may be found in the Contact URI ('uri'
 event parameter) and may be extracted using the {uri.param,name}
 transformation. From here onwards, it is up to the script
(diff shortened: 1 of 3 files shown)


bogdan-iancu:
Good catch here @denys-i-didww , thanks for the fix

PR #4245, opened 2026-09-07
mid_registrar: persist TCP connections for absorbed REGISTERs

greenbea:
**Summary**
When `tcp_persistent_flag` is enabled, TCP connection lifetime is updated only after a forwarded REGISTER receives a successful response.

Absorbed REGISTERs return a local 200 OK without extending the receiving connection, leaving new or refreshed connections at the default lifetime.

**Details**

**Solution**
Apply the UAC Contact expiration to the TCP connection in the `quick_reply` path, matching the existing behavior for forwarded registrations.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/mid_registrar/save.c
+++ b/modules/mid_registrar/save.c
@@ -90,6 +90,39 @@ void calc_contact_expires(struct sip_msg* _m, param_t* _ep, int* _e,
 	LM_DBG("expires: %d\n", *_e);
 }
 
+static void persist_tcp_connection(struct sip_msg *msg)
+{
+	contact_t *ct;
+	struct sip_uri uri;
+	int expires, max_expires = 0;
+
+	if (!is_tcp_based_proto(msg->rcv.proto) ||
+	    !(msg->flags & tcp_persistent_flag))
+		return;
+
+	for (ct = get_first_contact(msg); ct; ct = get_next_contact(ct)) {
+		calc_contact_expires(msg, ct->expires, &expires, 1);
+		if (expires <= 0)
+			continue;
+
+		if (parse_uri(ct->uri.s, ct->uri.len, &uri) < 0) {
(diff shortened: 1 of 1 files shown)


modules/mid_registrar/save.c
```diff
@@ -90,6 +90,39 @@ void calc_contact_expires(struct sip_msg* _m, param_t* _ep, int* _e,
...
+	if (!is_tcp_based_proto(msg->rcv.proto) ||
+	    !(msg->flags & tcp_persistent_flag))
+		return;
+
+	for (ct = get_first_contact(msg); ct; ct = get_next_contact(ct)) {
+		calc_contact_expires(msg, ct->expires, &expires, 1);
+		if (expires <= 0)
+			continue;
+
+		if (parse_uri(ct->uri.s, ct->uri.len, &uri) < 0) {
```
bogdan-iancu, line 108: The whole idea with keep alive the TCP conn for a REGISTER relies on the symmetric communication - the registered contacts are reachable via the connection opened by the REGISTER.
So, makes no sense to check if the registered contacts are TCP or not (it does not fit into the above philosophy of the symmetry). IMO, skip the test for the contact proto and simply check the expire values for all the carried contacts.
IF if we take into calculation a contact that we shouldn;t (like maybe not re-using the REGISTER's conn), the worst we can get is a to longer lifetime for a TCP conn....which is harmless.

bogdan-iancu (changes requested):

modules/mid_registrar/save.c
greenbea: Thanks. I understand your symmetry argument. I followed the existing forwarded-REGISTER logic in `save_restore_rpl_contacts()` and `save_restore_req_contacts()`, which also checks for TCP-based Contact URIs. So I kept the same check here for consistency with the existing behavior.

PR #4246, opened 2026-09-07
proto_ws: handle CRLF keep-alives in WS messages

greenbea:
**Summary**

Browser WebSocket APIs cannot send native WebSocket Ping frames, so browser-based SIP clients use the RFC 5626 CRLF keep-alive mechanism, as permitted by RFC 7118.

**Details**

**Solution**

Recognize double-CRLF pings and single-CRLF pongs carried in WebSocket data frames. Reply with a single-CRLF text frame and prevent these keep-alives from reaching the SIP parser.

This applies to both WS and WSS transports.

**Compatibility**

**Closing issues**

### DIFF
--- a/modules/proto_ws/ws_common.h
+++ b/modules/proto_ws/ws_common.h
@@ -276,6 +276,18 @@ static inline int ws_send_pong(struct tcp_connection *con, struct ws_req *req)
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


razvancrainea:
I've just had the time to look into this, but it seems fine for me, so I've already merged it. Thank you for your contribution!

PR #4250, opened 2026-09-08
docs: correct function and statistic names the modules do not export

denys-i-didww:
**Summary**

#4241 corrected five parameter names the modules do not export, and its *Scope* section
said the same class in the Exported Functions, MI Commands and Statistics sections was
left for a separate change. This is that change. Three names in three modules; the code is
right in each and the page is wrong.

| Module | Section | README documents | Code exports |
|---|---|---|---|
| `janus` | Exported Functions | `janus_send_requeest()` | `janus_send_request` — `janus_mod.c:67` |
| `event_sqs` | Exported Functions | `sqs_publish_message()` | `sqs_send_message` — `event_sqs.c:65` |
| `registrar` | Exported Statistics | `defaults_expires` | `default_expire` — `reg_mod.c:233` |

A script calling a function the module does not export does not load, and an MI query for
a statistic that does not exist returns *Statistics Not Found*.

**Details**

**`janus`** is a plain misspelling — *requeest* — and only in the heading, `README.md:163`.
Two lines below, the example's title and the example itself already read
`janus_send_request()`, so the page contradicts itself and the heading is the odd one out.

**`event_sqs`** documents the name of the C function rather than the name of the script
command. `event_sqs.c:65` reads

```c
{"sqs_send_message", (cmd_function)sqs_publish_message, {
```

so `sqs_publish_message` is the static implementation (`:48`, `:425`) and
`sqs_send_message` is what a script may call. The README uses the implementation's name in
all four places — the prose at `:77`, the heading at `:112`, the example title at `:130`
and the example line at `:134` — and nothing on the page corrects the reader.

**`registrar`** documents the statistic as `defaults_expires`, while `stat_export_t`
declares `default_expire`. The documented name is one letter away from the *parameter*
`default_expires`, which does exist (`reg_mod.c:203`), and the description under the
heading — *"The value of default_expires parameter"* — correctly describes that parameter;
the statistic does mirror it (`reg_mod.c:405`). Only the heading changes. The module's
other four documented statistics match `stat_export_t` exactly and are left alone.

**Reproduced**

`opensips 4.1.0-dev`, Debian 12, built from source.

The function name fails when the script is loaded, not at `modparam` parsing — a different
failure point from #4241, which is part of why this is a separate PR:

```
$ opensips -C -f j.cfg            # janus_send_requeest("a","b"); as the page reads today
CRITICAL:core:yyerror: parse error in j.cfg:9:33-34: unknown command <janus_send_requeest>, missing loadmodule?

$ opensips -C -f j.cfg            # janus_send_request("a","b"); as this PR documents
NOTICE:core:main: config file ok, exiting...
```

The statistic fails at the MI query, a third failure point. `registrar` running with
`usrloc`, `signaling`, `tm` and `mi_http`:

```
$ ... "method":"get_statistics","params":{"statistics":["defaults_expires"]}
{"jsonrpc":"2.0","error":{"code":404,"message":"Statistics Not Found"},"id":1}

$ ... "method":"get_statistics","params":{"statistics":["default_expire"]}
{"jsonrpc":"2.0","result":{"registrar:default_expire":3600},"id":1}
```

Listing every statistic on that instance returns `registrar:default_expire` and no
`defaults_expires`.

`event_sqs` is the one row not run. It links against `libaws-cpp-sdk-sqs`, which Debian
does not package, so the module cannot be built without building the AWS SDK first.
Compiling `event_sqs.c` on its own does work, and `nm` on the object shows
`sqs_publish_message` as a local text symbol — the static function — while
`sqs_send_message` is the name in `cmds[]`. That is source-level evidence, not a run, and
I would rather say so than imply otherwise.

**Scope**

Three renames, nothing else. The same sweep, re-run against 08e7f1c37a over every
module's Exported Functions, Asynchronous Functions, MI Functions, Statistics and
Pseudo-Variables sections, turns up two more documented names — `presence_callinfo`'s
`sca_set_calling_line()` and `ratelimit`'s `rl_bin_status` — that are not renames at all:
neither has an implementation anywhere in the tree, so there is nothing to point the
documentation at and the fix is yours to choose. They need an issue, not a patch, and are
not in this PR.

Everything else that looked like a mismatch is not one, and is recorded here so it is not
re-reported: MI command names come from `#define`s (`usrloc`, `rtpengine`, `pike` and
others), `callops`'s `E_CALL_TRANSFER` and `E_CALL_HOLD` are built by token pasting from
`"E_CALL_" #_name`, `registrar` and `mid_registrar`'s `pn_process_purr()` comes from a
macro in `lib/reg`, and `call_center`, `janus` and `rr` have `####` sub-headings where
their siblings use `###`, which a heading scan reads as names.

**Compatibility**

Documentation only, no behaviour change. Nothing that works today stops working: none of
the three documented names could ever have been used.

---

- each exported name is the literal in the module's own export array — `janus_mod.c:67`
  (`cmd_export_t`), `event_sqs.c:65` (`cmd_export_t`), `reg_mod.c:233` (`stat_export_t`);
- each documented name appears nowhere in the tree as a string literal — repository-wide
  grep over `*.c`, `*.h` and `*.cpp`, which is what distinguishes these three from the
  macro-defined names listed under *Scope*;
- every occurrence in each README was changed, so no page is left carrying both spellings
  — one in `janus`, four in `event_sqs`, one in `registrar`;
- `registrar`'s other four statistics were checked against `stat_export_t` and left alone,
  and the parameter `default_expires` it is confused with is real — `reg_mod.c:203`;
- the two failures and the two successes were run, not inferred — the output is quoted
  above, on a build of this branch;
- `event_sqs` was not run, and the reason and the weaker evidence are stated in place;
- the two names with no implementation were confirmed absent by the same repository-wide
  grep before being excluded from this PR.

### DIFF
--- a/modules/registrar/README.md
+++ b/modules/registrar/README.md
@@ -1487,7 +1487,7 @@ Value of max_expires parameter.
 The value of max_contacts parameter.
 
 
-#### default_expire
+#### default_expires
 
 
 The value of default_expires parameter.
--- a/modules/registrar/README.md
+++ b/modules/registrar/README.md
@@ -1487,7 +1487,7 @@ Value of max_expires parameter.
 The value of max_contacts parameter.
 
 
-#### defaults_expires
+#### default_expire
 
 
 The value of default_expires parameter.
--- a/modules/event_sqs/README.md
+++ b/modules/event_sqs/README.md
@@ -74,7 +74,7 @@ Here you can find some cli commands such as create-queue, send/receive-message, 
 
 
 This parameter specifies the configuration for an SQS queue that can be used
-to publish messages directly from the script, using the sqs_publish_message() function
+to publish messages directly from the script, using the sqs_send_message() function
 or to send messages using raise_event function.
 
 
@@ -109,7 +109,7 @@ modparam("event_sqs", "queue_url",
 ### Exported Functions
 
 
-#### sqs_publish_message(queue_id, message)
+#### sqs_send_message(queue_id, message)
 
 
 Publishes a message to an SQS queue. As the actual 
@@ -127,11 +127,11 @@ The function has the following parameters:
 - *message (string)* - The payload of the message to publish.
 
 
-```opensips title="sqs_publish_message() function usage"
+```opensips title="sqs_send_message() function usage"
 ...
 
 $var(msg) = "Hello, this is a message to SQS!";
-sqs_publish_message("q1", $var(msg));
+sqs_send_message("q1", $var(msg));
 
 ...
 		
--- a/modules/janus/README.md
+++ b/modules/janus/README.md
@@ -160,7 +160,7 @@ modparam("janus", "janus_db_table", "my_janus_table")
 ### Exported Functions
 
 
-#### janus_send_requeest(janus_id, janus_command[, response_var])
+#### janus_send_request(janus_id, janus_command[, response_var])
 
 
 Run an arbitrary command on an arbitrary Janus socket. The
--- a/modules/registrar/reg_mod.c
+++ b/modules/registrar/reg_mod.c
@@ -135,7 +135,15 @@ stat_var *accepted_registrations;
 stat_var *rejected_registrations;
 stat_var *max_expires_stat;
 stat_var *max_contacts_stat;
-stat_var *default_expire_stat;
+stat_var *default_expires_stat;
+
+/* the "default_expire" statistic is kept under its old, misspelled name for
+ * backwards-compatibility; reading it through a function rather than a second
+ * stat_var keeps one source of truth, so the two names cannot drift apart */
+static unsigned long get_default_expire(void *foo)
+{
+	return (unsigned long)default_expires;
+}
 
 /** SIGNALING binds */
 struct sig_binds sigb;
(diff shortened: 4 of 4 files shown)


modules/registrar/README.md
```diff
@@ -1487,7 +1487,7 @@ Value of max_expires parameter.
 The value of max_contacts parameter.
 
 
-#### defaults_expires
+#### default_expire
```
bogdan-iancu: Indeed, the code provides the "default_expire", which is broken, as it should be "default_expires" - same name as the modparam it accesses. 
Could you fix the stat name to default_expires please and keep the broke one for backward compat also ? Thanks

bogdan-iancu (changes requested):

modules/registrar/README.md
denys-i-didww: Done — registered as `default_expires`, old name kept.

It is a `STAT_IS_FUNC` alias rather than a second stat_var: two entries sharing one
`stat_var**` do not work, since `__register_stat()` overwrites `*pvar` per entry, so
`update_stat()` reaches only the last one and the earlier name stays at 0.
Through a function there is one source of truth — with `default_expires` set to 1800
and then 900, both names follow.

bogdan-iancu:
Thank you @denys-i-didww

Issue #2833, opened 2022-05-17
Add rport in VIA from local_route


rroiss:
Hi, 
following up to the issue #797 (https://github.com/OpenSIPS/opensips/issues/797#issue-134317923)
I am facing the same problem using the siprec module.
My SIPREC server has a public IP and my opensips is behind a NAT.
Without adding rport, the SIPREC server sends the response to port 5060, instead of the NATTED port.
Neither force_rport() nor add_local_rport() seems to work.
Is there any other way to set the rport?

I'm using OpenSIPS 3.2.4
Thanks.

bogdan-iancu:
@rroiss , just to be sure, you are talking about using the force_rport() from local_route, right ? If so, this is expected not to work as the function is designed to add the rport to the *received* VIA hdr.
Still, the add_local_rport() func is the right one, but checking its implementation I see it works only when proxying requests, not when generating as UAC.... The change is decent, I will put it on the todo list

bogdan-iancu:
@rroiss , could you please test this fix:
```
diff --git a/modules/tm/uac.c b/modules/tm/uac.c
index 61490835e..49545f658 100644
--- a/modules/tm/uac.c
+++ b/modules/tm/uac.c
@@ -248,8 +248,10 @@ static int run_local_route( struct cell *new_cell, char **buf, int *buf_len,
                        }
                }
 
-               /* if interface change, we need to re-build the via */
-               if (new_send_sock && new_send_sock != dialog->send_sock) {
+               /* if interface change or new VIA related flags were added,
+                * we need to re-build the via */
+               if ( (new_send_sock && new_send_sock != dialog->send_sock)
+               || req->msg_flags&FL_FORCE_LOCAL_RPORT ) {
 
                        LM_DBG("Interface change in local route -> "
                                "rebuilding via\n");
```

bogdan-iancu:
@rroiss , any chances for testing that patch ?

bogdan-iancu:
OK, I took the time to quickly test it and the fix is now on git, with all the backports.

Issue #797, opened 2016-02-17
How to add rport to REGISTER message


arovetto:
Hi
I'm using with Opensips 2.2dev uac_registrant module to register to an SIP Provider. They said the registration fails because the top most request Via hasn't the rport parameter.

I've tried to add that parameters using a local_route as follows:

local_route {
        xlog("L_INFO", "local_route");
        if (is_method("REGISTER")) {
                xlog("L_INFO", "Sending a REGISTER to $ru");
                add_local_rport();
        }
}

But it seems to have no effects:

Feb 17 16:45:46 newopensips-cti1 /sbin/opensips[15391]: local_route
Feb 17 16:45:46 newopensips-cti1 /sbin/opensips[15391]: Sending a REGISTER to sip:sip.vohippo.com
Feb 17 16:45:46 newopensips-cti1 /sbin/opensips[15391]: DBG:tm:t_uac: Change in local route -> rebuilding buffer

Is it a bug?
Is there another way to add it?

Thank you

bogdan-iancu:
@arovetto, the SIP RFC3261 does not require at all the presence of rport during a registration - this is a bogus requirement. 
Changes over VIA in local route does not work, as the VIA is internally generated after the local route, so basically you cannot change it.

Regards,
Bogdan

arovetto:
Thanks @bogdan-iancu,
The requirement is related to the NAT traversal, my SIP Provider has a public IP, and my opensips is behind a NAT. Looking at RFC 6314 (NAT Traversal Practices for Client-Server SIP) at §5.1.1.1 there is an example with a REGISTER with rport used to receive correctly the response through the NAT firewall.
Without adding rport, the provider sends the response to port 5060, instead of the NATTED port.
Thank you

bogdan-iancu:
@arovetto , could you please try the following patch:
https://gist.github.com/bogdan-iancu/031cdeadfae46e5e467a

and let me know if it does the trick for you.

arovetto:
I've tried your patch and it works!!

Thank you

Parkistler:
wow. it's here, thank you

PR #4253, opened 2026-09-09
docs: add_local_rport() also covers locally generated requests [skip ci]

denys-i-didww:
**Summary**

The `add_local_rport()` entry in the core function reference describes the behaviour as it
was before 2022. It says the function *"affects only the current processed request"*, which
reads as relayed traffic only, and a script author looking for a way to put `rport` in a
request OpenSIPS generates itself finds nothing on the page to say it is possible.

It has been possible since 22fd1760a7 and its three siblings, 14 September 2022, which
taught `run_local_route()` to rebuild the Via when `FL_FORCE_LOCAL_RPORT` is set. This adds
one sentence saying so.

**Details**

`add_local_rport()` sets `FL_FORCE_LOCAL_RPORT` (`core_cmds.c:1098-1101`) and is exported
for `ALL_ROUTES` (`core_cmds.c:251-253`), `LOCAL_ROUTE` included. Two places act on the
flag:

- `msg_translator.c:2236`, when building the Via for a request being relayed;
- `modules/tm/uac.c:226` and `:265`, when `t_uac_prepare()` has run `local_route`
  (`uac.c:558`) and the script set the flag on a request TM generated itself.

The second is the one the page does not mention, and it is not obscure — it is the fix
Bogdan wrote for #2833, where the reporter had already tried both `force_rport()` and
`add_local_rport()` and reported that neither worked. It shipped to `3.1`, `3.2`, `3.3` and
`master`, so every branch since carries it, but the documentation was not updated with it.

#797, open since 2016 and still the first thing a search finds for this, describes the
answer as an unmerged patch against `assemble_via()`. Nothing points a reader from there to
what the software actually does today.

**Reproduced**

Built from `master` and from `4.0`, `uac_registrant` configured against a local listener,
the REGISTER taken off the wire with `tcpdump`. One variable — the presence of the
`local_route` block:

| Branch | Configuration | Via of the REGISTER |
|---|---|---|
| `master` | as shipped | `…;branch=z9hG4bK03a9.27f11ca.0` |
| `master` | `local_route { add_local_rport(); }` | `…;branch=z9hG4bKd679.b4e44786.0;rport` |
| `4.0` | as shipped | `…;branch=z9hG4bKe37d.06470256.0` |
| `4.0` | `local_route { add_local_rport(); }` | `…;branch=z9hG4bK78a4.928f4166.0;rport` |

End to end, the same three lines are what let a registration complete through a symmetric
NAT: with them the registrar's `200 OK` comes back to the port it observed
(`received=…;rport=…`) instead of to the one the Via advertises, and `uac_registrant`
reaches `REGISTERED_STATE`. Without them the REGISTER goes out, no answer arrives, and
nothing is logged on either side.

**Solution**

One sentence under `add_local_rport()`, naming `local_route` and linking to it, with
`uac_registrant` as the example of a request OpenSIPS generates itself — and a second
example block showing the form, since the one already there shows only the relayed case:

```opensips

local_route {
    add_local_rport();
}

```

Two examples under one entry is the layout `subscribe_event()` and `exit()` already use on
this page. Documentation only, no behaviour change.

**Compatibility**

None affected. The behaviour being documented has been in every branch since 2022-09-14;
only the page was behind.

---

- the flag and the export — `core_cmds.c:1098-1101` and `:251-253`, `ALL_ROUTES`;
- both code paths that consume it — `msg_translator.c:2236` for relayed requests,
  `modules/tm/uac.c:226` and `:265` for locally generated ones, reached through
  `run_local_route()` at `uac.c:558`;
- the commit and its backports — `git log -S FL_FORCE_LOCAL_RPORT -- modules/tm/uac.c`
  returns 22fd1760a7, 570c0aafab, fbaef964b6 and c85d93c7a3, all dated 2022-09-14, and
  `git branch --contains` places them on 3.1, 3.2, 3.3 and master;
- that the page never mentions it — the whole `add_local_rport()` section is the two
  sentences quoted above plus the example;
- the four rows in the table were run, not inferred, with the REGISTER captured on the
  wire rather than read from a log;
- the link target resolves — `docs/manual/../../modules/uac_registrant/README.md` exists,
  and `../../modules/*/README.md` is the form already used in `Modules.md` and
  `Interface-MI.md`.


bogdan-iancu:
@denys-i-didww , the original doc seems clear to me - the `rport` is added to the VIA constructed by OpenSIPS. And yes, the locally generated request also have such a VIA...
But more docs never killed anyone :)

PR #4258, opened 2026-09-14
rtpengine: avoid wrongful node disabling (local errors, stale replies, timeouts) + fd and bencode fixes

zbyslb:
## Summary

Five fixes for the `rtpengine` module (plus one in its bundled `bencode` helper), all found and verified while hardening a production OpenSIPS 3.6 platform. The overarching theme: **the node-disabling logic is too eager** — it takes the proxy down for conditions that are not the proxy's fault, and in the worst case a single local resource shortage flags *every* node as down at once.

Each commit is self-contained and can be reviewed/cherry-picked independently.

### `bencode`: fully initialize the buffer before the first allocation

If `__bencode_piece_new()` fails (pkg memory exhausted), `bencode_buffer_init()` returns -1 having only assigned `buf->pieces` — `free_list` and `error` keep their previous (stack garbage) values. Callers cannot distinguish the failure stage and routinely run `bencode_buffer_free()`, whose first statement dereferences `buf->free_list`: undefined behaviour (call through a garbage pointer). Zeroing the struct up front makes the failure path a safe no-op free; the success path assigns all fields explicitly, so behaviour is unchanged.

### `rtpengine`: close the actual fd, not the node index, on reply read error

`RTPE_IO_ERROR_CLOSE(param->node->idx)` expands (on its EPIPE/EBADF branch) to `close(idx)` — closing whichever descriptor happens to carry the node's array index number (for the first nodes in a set: one of the stdio fds) — and then writes `-1` into `node->idx`, making every later `rtpe_socks[node->idx]` an out-of-bounds array access. The fd that actually failed is the per-command socket created by `start_async_send_rtpe_command()`.

### `rtpengine`: do not disable nodes on local resource errors

An async command builds a fresh UDP socket per command; `connect()` on it makes the kernel autobind a local source port from `ip_local_port_range` (shared between TCP and UDP). Once that pool runs out, `connect()` fails with `EAGAIN`, and `socket()` may fail with `EMFILE`/`ENFILE`/`ENOMEM`. These are local machine conditions, yet the code jumped to `badproxy` and disabled the node. With the pool exhausted, every command towards every node fails at once — a single local resource shortage flags **all** nodes as down, while the long-lived `rtpe_socks[]` probes (already bound, no new port needed) keep succeeding, so the nodes flip between enabled and disabled until the process is restarted. We hit exactly this in production (log signature: `can't connect to RTP proxy udp:... (11:Resource temporarily unavailable)` immediately followed by `disable it` for every node). Local errnos (`EAGAIN`, `EADDRNOTAVAIL`, `ENOBUFS`, `ENOMEM`, `EMFILE`, `ENFILE`) now only release the fd; errors that do reflect node/network reachability (`ECONNREFUSED`, `ENETUNREACH`, `EPIPE`, ...) still disable the node as before. The check is only performed immediately after the failing syscall.

### `rtpengine`: do not disable a node that is proven alive by its data

In the sync command path, a late reply of a *previous* command (still queued in the socket buffer) fails the cookie comparison and counts as a full retry failure — although the proxy just sent us a datagram. The fix tracks whether any data was received during the retries; a node that produced traffic keeps its enabled state, a completely silent node is still disabled as before.

### `rtpengine`: do not disable a node on an async reply timeout

An async timeout means one command did not get its reply in time — a single lost/late datagram is enough. That is not evidence the node is down, and a burst of timeouts (brief network issues) would disable all nodes at once. Node liveness is better left to the existing probes: the periodic `rtpe_test()` ping and the sync path, which only disables a node that stays completely silent while retrying.

## Testing

Built and running on a production OpenSIPS 3.6.9 deployment (Debian 12, ~10 rtpengine nodes behind it); the local-resource-error path was exercised by the production incident described above — after the fix, the same port-exhaustion condition only fails individual commands while all nodes stay enabled.


zbyslb:

PR #4260, opened 2026-09-14
core: add timeout support to launch() async operations

zbyslb:
## Summary

`launch()` operations have **no timeout** in the core, while `async()` ones do. For any module that opens a socket per operation, a reply which never arrives leaks the descriptor **and its local ephemeral port for the lifetime of the process** — because a zero timeout in `io_wait` never expires.

This fixes that, and is the root cause of a production outage we hit on a ~10-node rtpengine deployment: after weeks of `launch(rtpengine_delete())` on call teardown, `net.ipv4.ip_local_port_range` (shared between TCP and UDP) was exhausted, every new UDP `connect()` started failing with `EAGAIN(11)`, and all rtpengine nodes appeared unreachable at once.

### Why it leaks

`async_script_launch()` registers the module's fd with `reactor_add_reader()`:

```c
if (reactor_add_reader(fd,F_LAUNCH_ASYNC,RCT_PRIO_ASYNC,(void*)ctx)<0){
```

which expands to `io_watch_add(..., prio, 0, ...)` — timeout = 0. And zero means "never expires":

* `io_wait.h:478` — `if (timeout) timeout += get_ticks();`
* `io_wait_loop.h:114` — `e->timeout!=0 && e->timeout<=curr_time`

The script `async()` path instead uses `reactor_add_reader_with_timeout(...)` and calls the module's `timeout_f()` on expiry (`tm/async.c:133`), so its fds are always reclaimed. Neither is true for `launch()`, which also never invokes `timeout_f` at all.

### Changes

**Core (commit 1)**

* Arm the timeout when the module declared both `timeout_f` and `timeout_s`. Since `launch()` takes no timeout parameter, the value comes from the module itself through `ctx->async.timeout_s` — which `async.h` already documents as a module output parameter: *"on output: an updated timeout, if any, as processed by the module"*. Modules declaring nothing keep today's behaviour (no timeout), so this cannot break existing `launch()` users.
* Thread the reactor's `was_timeout` flag into `async_launch_resume()` and run `timeout_f()` instead of `resume_f()` on expiry, mirroring the script async path. All five `handle_io()` dispatchers already hold that flag — the `F_SCRIPT_ASYNC` case directly above each `F_LAUNCH_ASYNC` case forwards it.
* Keep the timeout armed when a resume function swaps the fd through `ASYNC_CHANGE_FD`.

**rtpengine (commit 2)**

Declare a (modparam-configurable, default 5s) timeout so its per-command socket is reclaimed. The value only ever fires on a lost reply, and a script `async()` timeout still takes precedence when it is smaller — so there is no behaviour change for the async path.

### Testing

The core change is additive: when no module declares a timeout, the registration call is the same one as before. The rtpengine part was built and run on a production 3.6.x deployment (as a local overlay) — see the commit message for the incident it resolves.

### Note

For 3.6.x the same patch applies with only cosmetic differences (no `profiling` calls, no `ASYNC_SET_RESUME_F`); happy to submit a backport to the stable branch if wanted.


zbyslb:

